PR #9, opened 2016-10-25
Remove routing-related code

mmcgahan:
Some of the components in the library had an implicit dependency on a particular routing setup (both in terms of what routes existed, e.g. `/:urlname`, and the routing package used, i.e. `react-router`). This doesn't work for apps choosing to use a different routing configuration.

For a general-use UI component lib, there shouldn't be _any_ routing dependency, so this PR removes all `<Link>` elements, which become the responsibility of the parent component.

``` jsx
// before:
<Avatar to='/path' />
<GroupCard group={group} />
<NavItem to='/path'>My Path</NavItem>

//after
<Link to='/path'><Avatar /></Link>
<Link to={`/${group.urlname}`}><GroupCard group={group} /></Link>
<NavItem><Link to='/path'>My Path</Link></NavItem>
```

The parent application could also provide its own wrappers for these base components if its maintainers decided to avoid the manual tooling requirement


sadafie:
Yeah, I was wondering about this, but my branches were getting to large and I left it. Good call. We prob need a corresponding PR is `mup-web` to reflect these changes once this is merged.

sadafie:
@mmcgahan this looks good to merge. I had a corresponding question in a different PR, that I can't find, but basically take `GroupCard` as an example. What we want to do in the consumer code is:

```
<Link ...>
   <GroupCard ...>
</Link>
```

But how do you handle the case where the `GroupCard` wants to render a link inside of it, say an  RSVP button (though technically that would dispatch an action, but let's say it was just a link). This seems wrong and wouldn't work:

```
<Link ...>
   <GroupCard ...>
       <Link>
        </Link>
    </GroupCard>
</Link>
``` 

Anyways, let's merge this because it's the right thing to do. Chances are I'm just missing something obvious :)

mmcgahan:
@sadafie my responses is [here](https://github.com/meetup/mup-web/pull/345#discussion_r85807716), but basically you would pass any embedded routing elements as `props` - they would either be click handlers or React components

```jsx
<GroupCard onRsvpClick={myHandlerFn} />

// or

const myLink = <Link to='/'>Foo</Link>;
<GroupCard theLink={myLink} />
```

PR #12, opened 2016-10-31
WC-2 Create npm package using travis and semver

sadafie:
I copied over the `.travis.yml` file from `meetup-web-platform`, and removed the `npm run coverage` command since that is not set up in this repo yet. I copied over the `Makefile` from `meetup-web-platform` as well.

Still need to complete the following:

- [x] Create a .travis.yml
- [x] Set travis env vars
- [x] Create Makefile for versions
- [x] Create `beta` versions for every PR build
- [x] Compile using babel to create es5 modules under `lib/`
- [x] Update README
- [x] `npm run lint`
- [x] Tests pass


.travis.yml
```diff
@@ -0,0 +1,58 @@
+sudo: required
```
sadafie: This is identical to `meetup-web-platform .travis.yml` except doesn't run the coveralls step yet

Makefile
```diff
@@ -0,0 +1,6 @@
+CI_BUILD_NUMBER ?= $(USER)-snapshot
```
sadafie: identical to `meetup-web-platform Makefile`

src/AvatarMember.jsx
```diff
@@ -30,7 +30,7 @@ class AvatarMember extends React.Component {
 				src={(member.photo || {}).photo_link}
 				className={classNames}
 				{...other} />
-			);
+		);
```
sadafie: lint autofix

src/avatar.story.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import Avatar from './Avatar.jsx';
 import { storiesOf, action } from '@kadira/storybook';
-import { MOCK_MEMBER } from 'meetup-web-platform/util/mocks/api';
+import { MOCK_MEMBER } from 'meetup-web-platform/lib/util/mocks/api';
```
sadafie: @mmcgahan is it no longer possible to import the es6 modules? I was getting an error importing `meetup-web-platform/util/mocks/api`. Or is it just that now it would have to be `meetup-web-platform/src/util/mocks/api`?

sadafie:
@mmcgahan also wondering if I should move `utils/`, `assets/`, and `icons/` under `src/`?

mmcgahan:
Anything that gets built should be in `src/`, but I think the icons don't get built? I'm also not 100% sure what happens with the SCSS, but if it's part of the build then yes it should go in `src/`

sadafie:
> Anything that gets built should be in src/, but I think the icons don't get built? I'm also not 100% sure what happens with the SCSS, but if it's part of the build then yes it should go in src/

SCSS is just for storybook, so no, I only moved utils

PR #13, opened 2016-11-03
Read from env var for all npm publishing

mmcgahan:
no need to 'dynamically' update `.npmrc`. You can't publish from your local machine without an `NPM_TOKEN` env variable, but we never publish from local machines anyway

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,6 +14,7 @@ node_modules
 npm-debug.log
 .swo
 .swp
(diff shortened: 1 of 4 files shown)


sadafie:
thanks for fixing this @mmcgahan... sorry i didn't get to it in time.

mmcgahan:
turns out it's still not an ideal solution - #16

PR #17, opened 2016-11-15
Storybook SVG icons

sadafie:
See https://meetup.atlassian.net/browse/WC-5.

The svgstore webpack plugin doesn't work with well with storybook. Instead I'm attempting to use the svg sprite from the `swarm-icons` pkg.

From what I can tell this should be working, but when you run `npm run storybook` and select any of the `Icon` components, nothing renders.

The sprite is emitted via the webpack `file-loader` at `/<hash>.inc`, and the path is being injected into the `use xlink:href` correctly. 

Here's sample generated markup:

```
<span data-reactroot="" class="svg svg--chevron-right">
	<svg
		preserveAspectRatio="xMinYMin meet"
		width="16" height="16" viewBox="0 0 16 16"
		class="svg-icon valign--middle"
		role="img"
		aria-labelledby="chevron-right_title">
			<title id="chevron-right_title"></title>
			<use role="presentation" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/d7dfcb588c5221f44c7adc243bca425a.inc#icon-chevron-right"></use>
	</svg>
</span>
```

cc @akdetrick @mmcgahan I could use some help with this one, if anyone has ideas. I tried a few other ways of loading the sprite as well, but since this is being loaded and emitted correctly, I'm assuming the issue is somewhere the JSX.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,7 +1,5 @@
 const path = require('path');
-const SvgStore = require('webpack-svgstore-plugin');
 
-const ICON_PATH = path.resolve(__dirname, '../icons', '**/*.svg');
 const SCSS_PATH = path.resolve(__dirname, '../assets', 'scss');
 const CSS_PATH = path.resolve(__dirname, '../assets', 'css');
 const SRC_PATH = path.resolve(__dirname, '../src');
@@ -40,23 +38,6 @@ module.exports = {
 
 	resolve: {
 		extensions: ['', '.js', '.jsx']
-	},
-
-	plugins: [
-		new SvgStore(
-			[ICON_PATH], // input path
-			'svg',       // output path
-			{
-				name: '[hash].sprite.svg',
-				chunk: 'preview',
-				prefix: 'icon-',
-				svgoOptions: {
-					plugins: [
-						{ removeTitle: true }
-					]
-				}
-			}
-		)
-	]
+	}
 };
 
--- a/package.json
+++ b/package.json
@@ -36,12 +36,9 @@
     "babel-preset-react": "6.16.0",
     "babel-preset-stage-2": "6.18.0",
     "babel-register": "6.18.0",
-    "bourbon": "4.2.7",
     "classnames": "2.2.5",
     "eslint": "3.9.1",
     "eslint-plugin-react": "6.4.1",
-    "julep": "1.4.5",
-    "meetup-swatches": "3.1.1",
     "meetup-web-platform": "0.5.189",
     "node-sass": "3.10.1",
     "react": "15.3.2",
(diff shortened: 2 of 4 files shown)


sadafie:
> loading it into the DOM directly and then referencing it from the individual <svg> elements rather than treating it as an external resource

I tried that as well, but couldn't get it working.

mmcgahan:
dang. Have you tried manually writing the HTML to get it to work? I think I would try to get the SVGs loading that way and then work on tweaking the JSX to produce the same output - if it can be done in HTML, it can (almost) definitely be done in JSX

sadafie:
I had tried this, in `icon.story.jsx`:

```
 const svgSprite = require('raw!swarm-icons/dist/sprite/sprite.inc');
 
 storiesOf('Icon', module)
 → .addDecorator((story) => (
 → → <div class='svgContainer'>
 → → → <div dangerouslySetInnerHTML={{ __html: svgSprite }} />
 → → → {story()}
 → → </div>
 → ))
 → .add('Default', () => (
 → → <Icon shape='chevron-right' />
 → ))
```

But dangerouslySetInnerHTML didn't seem to work. `svgSprite` is the string content of the sprite file, so I'd assume this would render into the dom, and then I could wrap each Icon tag with this. It just prints out the string, so I'm not sure if there's a bug wit create React svg, rect, etc elements, or if I'm missing something.

akdetrick:
If we want to make this work, the best thing to do is to include the icon sprite right after the opening `body` tag in a document (this helps avoid some cross-browser quirks). I'm not finding an easy way to change the storybook template to allow this...

I'm also wondering if we even really need to use `swarm-icons` to test our `Icon` component. It's the component we're testing and documenting, not the individual SVG icons. I'll look into adding an arbitrary SVG sprite to each icon story.

akdetrick:
@sadafie @mikem-mup 

I've proposed an alternate way of handling icons in `meetup-web-components` in a new PR:
https://github.com/meetup/meetup-web-components/pull/18

mmcgahan:
> I'm also wondering if we even really need to use swarm-icons to test our Icon component

Good call - as long as the markup is what we're expecting for the `Icon`, I agree that the challenge of actually loading the sprite could be independent

sadafie:
>If we want to make this work, the best thing to do is to include the icon sprite right after the opening body tag in a document (this helps avoid some cross-browser quirks). I'm not finding an easy way to change the storybook template to allow this... 

Nope, that's exactly what I was trying to do, and there doesn't seem to be a way with Storybook, which is unfortunate.

>I'm also wondering if we even really need to use swarm-icons to test our Icon component. It's the component we're testing and documenting, not the individual SVG icons. I'll look into adding an arbitrary SVG sprite to each icon story.

Good point about not needing `swarm-icons`, but I did try a version of loading the swarm sprite into `Icon.story.jsx` but that didn't work either. Perhaps I'm doing it wrong, but it'd be great if there's a workaround.

> I've proposed an alternate way of handling icons in meetup-web-components in a new PR:
#18

I'll take a look. happy to close this PR if there's a better way.

sadafie:
Closing. Fixed by #18

PR #20, opened 2016-11-28
Tabs "dumb" component

akdetrick:
**tl;dr: only opened this up for discussion. I'm currently investigating other directions we can take this.**

Due to coupling between `<Link` and `react-redux-router`, I set out to make a composite "dumb" component with the expectation we'd write a wrapper in `mup-web` to simplify the interface.

Maintaining this level of simplicity is a concern because:
1. we'll need to implement a wrapper component in each new application to generate `id` and `aria` attributes
2. `Tabs` currently does little more than abstract basic HTML
3. we _might_ be able to use an existing react tabs component with a thin wrapper that simply adds the correct SQ2 classes
4. I'm worried that we'll run into this situation with other composite components, like [menu](http://meetup.github.io/sassquatch2/ui_components.html#menu)

Any thoughts?


mmcgahan:
Is this closed because you don't any feedback? I agree these are tricky concerns, and we might want to re-think the expectation that `meetup-web-components` is `<Link>`-free.

The reasons _against_ including `<Link>` in MWC are:

1. `<Link>` has a specific router library dependency, most of which needs to be implemented by the consumer app - we would need to set a `peerDependency` on a particular version of React Router. This is probably manageable for a small number of consumer apps, but would become a large maintenance burden if we have lots of consumer apps, and isn't very OSS-oriented
2. Components with router dependencies suggests that the component is relatively complex, and may not be a good fit for a 'foundation'-level composable library - there are definitely exceptions (Tabs and other components specifically targeted at navigation might be valid exceptions), but it's an objective criteria for evaluating complexity.

One way to address the concern is by passing in the Link class directly to the component that needs it, like regular dependency injection:

```jsx
// in the consumer application:
import Link from 'react-router/lib/Link';
// ...
<Tab Link={Link} to='#lodi-dodi' />

// in MWC Tabs code
class Tab extends React.Component {
  render() {
    const { Link, to, id } = this.props;
    const aria = '...';
    return (
      <div>
        <Link to={to} aria={aria} id={id}
          <h3>My label</h3>
        </Link>
      </div>
    );
  }
}
Tab.propTypes {
  Link: React.PropTypes.func.isRequired,
  to: React.PropTypes.string.isRequired,
  // ...
}
```

akdetrick:
closed this because I had an Idea to make this better before I brought up the question again; this feedback is helpful though, thanks!

akdetrick:
Reopening; I think the answer is this:

1) Pass a `tabsRef` prop from the base `Tabs` component to its children in order to generate `id` and `aria` attributes on rendered elements
2) Allow users to pass any component (such as `Link`) as a child of `TabsListTab` (CSS will support any inline element as a child of `.tabs-tab`)
3) ... write a wrapper in `mup-web` to make the linking and selection toggling easier?

src/Tabs.jsx
```diff
@@ -0,0 +1,180 @@
...
+import cx from 'classnames';
+
+/**
+ * @module TabsListTab
+ *
+ * Direct child of `TabsList`.
+ * Renders individual tab control.
+ * Accepts a `selected` prop.
+ */
+export const TabsListTab = React.createClass({
```
mmcgahan: by convention, we're using `class TabsListTab extends React.Component {`

src/Tabs.jsx
```diff
@@ -0,0 +1,180 @@
...
+			className
+		);
+
+		return (
+			<li
+				role='tab'
+				id={`${tabsRef}_tab_${tabsIndex}`}
+				aria-controls={`${tabsRef}_panel_${tabsIndex}`}
+				onClick={onClick}
+				className={classNames} {...other}>
```
mmcgahan: for readability, `{ ...other}` and `>` can each go on a new line

```jsx
    className={classNames}
    {...other}
>
```

src/Tabs.jsx
```diff
@@ -0,0 +1,180 @@
...
+		full: React.PropTypes.bool
+	},
+	defaultProps: {
+		full: false
+	},
+	renderChildren() {
+		const { tabsRef, children } = this.props;
+
+		return React.Children.map(children, (kid, index) => {
+			if (kid.type === TabsListTab) {
```
mmcgahan: It's better to separate your type checking from the render logic - use a custom function in `propTypes` to check for invalid children so that the React component lifecycle stays consistent:

```jsx
TabsList.propTypes = {
  full: React.PropTypes.bool.isRequired,
  children(props, propName, componentName) {
    const prop = props[propName];
    let error = null;
    React.Children.forEach(prop, child => {
      if (child.type !== TabsListTab) {
        error = new Error(`'${componentName}' children should be of type 'TabsListTab'.`);
      }
    })
    return error;
  }
}
```

src/Tabs.jsx
```diff
@@ -0,0 +1,180 @@
...
+				'display--none': !selected
+			},
+			className
+		);
+
+		return (
+			<div
+				role='tabpanel'
+				id={`${tabsRef}_panel_${tabsIndex}`}
+				aria-lablledby={`${tabsRef}_tab_${tabsIndex}`}
```
mmcgahan: typo `labelledby`

src/tabs.story.jsx
```diff
@@ -0,0 +1,49 @@
+
+import React from 'react';
+import { Tabs, TabsList, TabsListTab, TabsPanel } from './Tabs';
+import { storiesOf } from '@kadira/storybook';
+import { Annotate } from './utils/storyComponents';
+
+storiesOf('Tabs', module)
+	.add('first selected', () => {
+		return (
+		<Annotate notes='The `tabsRef` prop of `Tabs` will generate aria attributes'>
```
mmcgahan: no biggie, but this could use another indent level

src/Tabs.jsx
```diff
@@ -0,0 +1,180 @@
...
+	propTypes: {
+		tabsRef: React.PropTypes.string.isRequired,
+	},
+	renderChildren() {
+		const { tabsRef, children } = this.props;
+
+		return React.Children.map(children, (kid, index) => {
+			// pass `tabsRef` prop to list and panel children
+			if (kid.type === TabsList || kid.type === TabsPanel) {
+				return React.cloneElement(kid, { tabsRef: tabsRef, tabsIndex: index });
```
mmcgahan: trivial ES6 nicety: if you rename the `index` arg to `tabsIndex`, the object arg on this line could just be `{ tabsRef, tabsIndex }`

src/tabs.story.jsx
```diff
@@ -0,0 +1,49 @@
...
+				<TabsPanel>
+					<p className='text--bold'>Snake content</p>
+				</TabsPanel>
+			</Tabs>
+		</Annotate>
+		);
+	})
+	.add('full width tab control', () => {
+		return (
+		<Annotate notes='The boolean prop `full` on `TabsList` will make `Tabs` 100% width'>
```
mmcgahan: same

src/tabs.test.jsx
```diff
@@ -0,0 +1,15 @@
...
+describe('TabsContainer', function() {
+
+	it('exists', function() {
+		const tabs = TestUtils.renderIntoDocument(<TabsContainer />);
+		const tabsNode = ReactDOM.findDOMNode(tabs);
+
+		expect(tabsNode).not.toBeNull();
+	});
+
+});
```
mmcgahan: probably good to add more tests to make sure that `selected` works

akdetrick: Definitely pending tests.

akdetrick:
We can made a SQ2 adjustment if needed; the direct child of `.tabs-tab` can be set to `block` with appropriate padding

src/Tabs.jsx
akdetrick: This was a style choice for tabs because this is a composite component - it was easier for me to look at `propTypes` inside of the object literals. Now that I'm looking at this again, I don't think that's a compelling enough argument to break from standard convention. I'll change this back.

mmcgahan: That's strange - we definitely have custom methods on [components in mup-web](https://github.com/meetup/mup-web/blob/master/src/features/find/FindContainer.jsx#L28) so I'll try to dig in to this.

Not sure if you tried it exactly as your example code indicates, but `render` does have to return a React element (usually with JSX), so there's an off chance that's the problem

mmcgahan: Okay I dug in and at first I was getting the error you're describing but then I realized I wasn't exporting all the classes that I needed to, so once I fixed that everything worked fine. [Here's the commit](https://github.com/meetup/meetup-web-components/commit/c627058) that has everything working using `class` - let me know if you're still having issues after checking that out

akdetrick: It's too bad we can't use the more compact definitions in the body of the class... maybe there's a babel setting we can flip later on. Thanks for taking a look at this!

src/Tabs.jsx
```diff
@@ -0,0 +1,209 @@
...
+export class TabsList extends React.Component {
+	renderChildren() {
+		const { tabsRef, children } = this.props;
+
+		return React.Children.map(children, (kid, tabsIndex) => {
+			return React.cloneElement(kid, { tabsRef, tabsIndex });
+		});
+	}
+	render() {
+		// remove non-standard DOM attributes before render returns
```
akdetrick: would it be better to just add these prop names to this?

```
const { className, full, tabsRef, tabsIndex, ...other } = this.props;
```

That would avoid dumping those two invalid DOM attributes into the `...other` spread

akdetrick: ^ this _would_ work, except we'd get a "defined, but never used" error.

mmcgahan: yeah you can always tell eslint to ignore that rule for those lines, though: http://eslint.org/docs/user-guide/configuring.html#configuring-rules

akdetrick:
Adding prelim tag; going to merge some of this work with the responsify tabs @mperrotti and @rxb wrote

akdetrick:
Removed Prelim tag; this is good to re-review and merge now.

src/tabs.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+		const firstTabClass = selectedTabNodes[0].getAttribute('class');
+
+		expect(selectedTabNodes.length).toBe(1);
+		expect(firstTabClass.indexOf('tabs-tab--selected') > -1).toBe(true);
+	});
+
+	it('applies variant classes correctly', function() {
+		const tabsClass = tabsNode.getAttribute('class');
+		const hasFullWidthClass = tabsClass.indexOf('tabs--full') > -1;
+		const hasBorderedClass = tabsClass.indexOf('tabs--bordered') > -1;
```
mmcgahan: a component that _always_ applied the variant regardless of input props would also pass this test - be sure to test the counterfactual case

src/tabs.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+		expect(hasFullWidthClass).toBe(true);
+		expect(hasBorderedClass).toBe(true);
```
mmcgahan: trivial, but you can use `toBeGreaterThan(-1)` along with `indexOf`:

```js
expect(tabsClass.indexOf('tabs--full')).toBeGreaterThan(-1);
```

not necessarily better than what you have, just wanted to throw it out as an alternative

src/utils/decorators.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+
+export const locale = story => {
+	const locale = 'en-US';
```
mmcgahan: might be good to use a var name that is different than the function name. Function could be `applyLocale`?

src/Tabs.jsx
```diff
@@ -0,0 +1,84 @@
...
+				{...other}>
+				<Link
+					role='menuitem'
+					to={url}>
+					{label}
+				</Link>
+			</li>
+		);
+	}
+}
```
mmcgahan: I recommend one-class-per-module.

src/Tabs.jsx
```diff
@@ -0,0 +1,84 @@
...
+
```
mmcgahan: propTypes for TabsTab?

PR #23, opened 2016-12-14
Addressed feedback about Flex and FlexItem components

mperrotti:

PR #24, opened 2016-12-15
Grid List

mperrotti:


sadafie:
Same question here as in #23 --

Is this intentionally merging back into the `responsify` branch? Any reason why it can't just go into `master`?

akdetrick:
lgtm, but I still want to check in with @rxb on the "all" breakpoint before merging to `master`

PR #26, opened 2016-12-20
Break out Modal component from responsify

nlaz:
#### Description
The modal component needs to be added to MUG Comm. This PR breaks out that component from responsify land so that we can use it in mup-web. The goal was to keep this component as "dumb" as possible so we are passing in a dismiss handler rather than handling dismissal within the component.

#### Screenshot

screen shot 2016-12-22 at 7 22 35 am

### DIFF
--- a/src/Modal.jsx
+++ b/src/Modal.jsx
@@ -1,0 +1,98 @@
+import React from 'react';
+import cx from 'classnames';
+import Icon from './Icon';
+import Button from './Button';
+
+export const MODAL_CLOSE_BUTTON = 'modal-closeButton';
+
+/**
+ * SQ2 Modal component
+ * @see {@link http://meetup.github.io/sassquatch2/views.html#modals}
+ * @module Modal
+ */
+class Modal extends React.Component {
+	constructor(props){
+		super(props);
+		this.onDismiss = this.onDismiss.bind(this);
+		this.onKeyDown = this.onKeyDown.bind(this);
+	}
+
+	onDismiss(e) {
+		e.stopPropagation();
+
+		if (this.props.onDismiss) {
+			this.props.onDismiss(e);
+		}
+	}
+
+	onKeyDown(e) {
+		if (e.key === 'Escape') {
+			this.onDismiss(e);
+		}
+	}
+
+	render() {
+		const {
+			className,
+			children,
+			fullscreen,
+			...other
+		} = this.props;
+
+		delete other.onDismiss; // onDismiss is consumed in this.onDismiss - do not pass it along to children
+
+		const classNames = cx(
+			className,
+			'modal'
+		);
+
+		const modalClasses = cx(
+			'view',
+			'padding--all',
+			{
+				'view--modalFull': fullscreen,
+				'view--modalSnap': !fullscreen
+			}
+		);
+
+		const dismissButtonClasses = cx(
+			MODAL_CLOSE_BUTTON,
+			'border--none'
+		);
+
+		return (
+			<div
+				role='dialog'
+				tabIndex='0'
+				onKeyDown={this.onKeyDown}
+				className={classNames}
+				{...other}>
+
+				<div className='overlayShim' onClick={this.onDismiss}>
+					<div className='inverted'></div>
+				</div>
+
+				<div className={modalClasses} >
+					<div className='align--right'>
+						<Button onClick={this.onDismiss} className={dismissButtonClasses}>
+							<Icon shape='cross' size='s' />
+						</Button>
+					</div>
+
+					{children}
+				</div>
+			</div>
+		);
+	}
+}
+
+Modal.propTypes = {
+	onDismiss: React.PropTypes.func.isRequired,
+	fullscreen: React.PropTypes.bool
+};
+
+Modal.defaultProps = {
+	fullscreen: false
+};
+
+export default Modal;
--- a/src/modal.test.js
+++ b/src/modal.test.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
+import { findComponentsWithType } from 'meetup-web-mocks/lib/testUtils';
 import Modal, { MODAL_CLOSE_BUTTON } from './Modal';
 
 describe('Modal', () => {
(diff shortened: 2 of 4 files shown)


nlaz:
Still need to add tests and comments and will also try to test it within mup-web land before removing the Prelim tag.

mmcgahan:
There are a few features that would be nice to nail early on like keyboard control (`esc` to close). It might be worth adding an optional `onKeydown` listener and figuring out the best way to capture those events in the modal.

Have you checked out some other React modal component implementations, e.g. https://reactcommunity.org/react-modal/? There are a lot of good ideas out there.

I came across a couple good articles with some great examples and ideas about implementing them:
- https://medium.com/@david.gilbertson/modals-in-react-f6c3ff9f4701#.4uuxpxbfd
- https://reactjsnews.com/modals-in-react

It looks like this is just the 'open' state - would it be possible to include a functioning open-and-close modal story so that we have a complete reference implementation?

nlaz:
@mmcgahan Those are great points. I think the modal component should definitely include accessibility features like keyboard controls. There's probably a lot more we can/should do with this component before it's complete. 

To clarify, this PR isn't meant to define what the modal component is, but rather create a minimal component for use in the `mup-web` module. Additional feature conversations should probably involve @akdetrick and the design systems team. However, I'm hoping this component is good enough currently to use in mup-web.

mmcgahan:
@nlaz would a11y be built before launch or would that be a post-launch followup? I'm hesitant to merge anything in this repo that isn't production-ready because once it's in `master` here, it's available for any platform consumer to use. For something interactive like this modal, it's probably helpful to outline a 'spec' of some sort so that we can evaluate how complete it is.

Since there's more to be done to make it production-ready (e.g. a11y and implementation instructions/examples), you can definitely build it into `mup-web` until it's complete, and then move it over here with some more comprehensive example stories.

akdetrick:
I would like to make keyboard navigation part of the "MVP" for any component we merge... 

```
	handleKeyDown(e) {
		switch(e.key) {
		case 'Escape':
			this.onDismiss();
			break;
		}
	}
```

```
onKeyDown={this.handleKeyDown}
```

src/Modal.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+	onDismiss(e) {
+		e.stopPropagation();
+
+		if (this.props.onDismiss) {
+			this.props.onDismiss(e);
+		}
+	}
+
+	handleKeyDown(e) {
```
mmcgahan: excellent - now you just need to make sure the modal elements gets 'focus' as soon as the modal opens. I think you'll want to do that in `componentDidMount`, depending on how it's used

src/Modal.jsx
```diff
@@ -0,0 +1,86 @@
...
+			{
+				'view--modalFull': fullscreen,
+				'view--modalSnap': !fullscreen
+			}
+		);
+
+		return (
+			<div
+				tabIndex='0'
+				onKeyDown={this.handleKeyDown}
```
mmcgahan: minor, but the convention for naming event handlers that are assigned to particular `onX` attributes in React is just to name them `onX` - `this.handleKeyDown` would be `this.onKeyDown`.

nlaz: Solid nitpick. I'll get a hang of these naming conventions soon.

akdetrick:
Looks like we're starting with a "dumb" component in that the interactivity must be added in the application where it's being used. There is a "fullscreen" prop, but I see there's already a story covering that. There may not be much else to add to the modal story...

mmcgahan:
LGTM - I'm hoping we can add the behavioral wrappers to this repo over time, and probably add some docs.

src/Modal.jsx
```diff
@@ -0,0 +1,87 @@
...
+					{children}
+				</div>
+			</div>
+		);
+	}
+}
+
+Modal.propTypes = {
+	fullscreen: React.PropTypes.bool
+};
```
mmcgahan: `propTypes` for `onDismiss` and `fullscreen`

src/modal.test.js
```diff
@@ -0,0 +1,65 @@
...
+		expect(modalEl.classList).toContain('modal');
+	});
+
+	it('displays modal content', () => {
+		expect(modalEl.innerHTML).toContain(content);
+	});
+
+	it('creates a Button component for dismissal', () => {
+		const len = TestUtils.scryRenderedComponentsWithType(modal, Button).length;
+		expect(len).toBeGreaterThan(0);
```
mmcgahan: this is another case where you could give the `dismiss` button a classname and search for that instead of looking for an anonymous `Button` somewhere in the modal - this test is finding more than just the `dismiss` button, so it would pass even if the dismiss button was gone.

src/modal.test.js
```diff
@@ -0,0 +1,65 @@
...
+	});
+
+	it('creates an Icon component for dismissal', () => {
+		const len = TestUtils.scryRenderedComponentsWithType(modal, Icon).length;
+		expect(len).toBeGreaterThan(0);
```
mmcgahan: be more specific - look for a classname, because the component could add other `Icon`s over time.

src/modal.test.js
```diff
@@ -0,0 +1,65 @@
...
+	});
+
+	it('executes onDismiss when dismiss button is clicked', () => {
+		const closeButton = modalEl.getElementsByTagName('button')[0];
```
mmcgahan: be more specific

eilinora:
This modal is LGTM, @mmcgahan or @akdetrick any final comments?

src/modal.test.js
```diff
@@ -0,0 +1,65 @@
...
+	it('has SQ2 modal styles', () => {
+		expect(modalEl.classList).toContain('modal');
+	});
+
+	it('displays modal content', () => {
+		expect(modalEl.innerHTML).toContain(content);
+	});
+
+	it('creates an Icon component for dismissal', () => {
+		const icon = modalEl.getElementsByClassName('svg-icon')[0];
```
mmcgahan: be more specific - this will match any icon that is ever part of the modal component, but you want to be able to select _only_ the icon that is used for dismissal.

mmcgahan:
LGTM, you can use [`array.find`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find) to find particular items in an array rather than `filter` in the future. Not supported in some browsers, but fine in Node

PR #27, opened 2016-12-20
Basic layout components

mperrotti:
Wait until #23 is merged to merge this.

### Basic layout components include:
- Bounds
- Chunk
- Section
- SectionTitle
- Stripe


assets/scss/section.scss
```diff
@@ -1,5 +1,36 @@
...
+
+$sectionSpaceMultiple: 2;
+
 @include responsiveBase() {
 	.section {
-		padding: $base*2 $base 0;
+		padding: $base*$sectionSpaceMultiple $base 0;
+	}
+
+	.section.bordered{
```
mmcgahan: by convention, we'd probably join this with the `.section {` block above with

```scss
.section {
  &.bordered { ... }
}
```

assets/scss/section.scss
```diff
@@ -1,5 +1,36 @@
...
+		padding-bottom: $base*$sectionSpaceMultiple - $base;
+		&:before{
+			left: $base;
+			right: $base;
+		}
+	}
+}
+
+.section.bordered {
```
mmcgahan: might also be more future-forward to nest the two class selectors here as well, your call

src/FlexItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+				[`flex-item--${growFactor}`] : typeof growFactor === 'number',
+			},
+			className
+		);
+
+		return (
+			<div
+				className={classNames}
+				{...other}>
+					{children}
```
mmcgahan: there's an extra indent here, but it's a little masked because the `prop`s are also indented - I recommend putting the `>` on its own line at the same indent level as `<div` to make is clear where the tag ends and the children begin

src/Section.jsx
```diff
@@ -9,11 +9,15 @@ class Section extends React.Component {
...
 			children,
 			className,
+			bordered,
 			...other
 		} = this.props;
 
 		const classNames = cx(
 			'section',
+			{
+				bordered: typeof bordered === 'boolean'
```
mmcgahan: ?

so `bordered` will get added as a class even when `bordered === false`?

src/SectionTitle.jsx
```diff
@@ -34,6 +33,12 @@ class SectionTitle extends React.Component {
...
 				</FlexItem>
 				{children &&
+					<Chunk>
+						<FlexItem shrink
+							>
+							{children}
+						</FlexItem>
+					</Chunk>
 					<FlexItem shrink>
 						{children}
```
mmcgahan: `children` get output twice? It's not clear how this would work

src/SectionTitle.jsx
```diff
@@ -34,6 +33,12 @@ class SectionTitle extends React.Component {
 					</Chunk>
 				</FlexItem>
 				{children &&
+					<Chunk>
+						<FlexItem shrink
+							>
```
mmcgahan: this doesn't really line up with anything. I think it should just go on the line above

src/Stripe.jsx
```diff
@@ -9,19 +10,35 @@ class Stripe extends React.Component {
 		const {
 			children,
 			className,
+			backgroundImage,
+			collection,
+			inverted,
+			isHero,
```
mmcgahan: all of these need `propTypes`

src/Stripe.jsx
```diff
@@ -9,19 +10,35 @@ class Stripe extends React.Component {
...
 			<div
 				className={classNames}
+				style={backgroundImage ? {backgroundImage: `url(${backgroundImage})`} : null}
 				{...other}>
-					{children}
+				{ isHero ?
+					<Bounds className='stripe-heroContent'>
+						{children}
+					</Bounds>
+					:
```
mmcgahan: by convention, this can go after `</Bounds>` or before `children`

src/bounds.story.jsx
```diff
@@ -1,8 +1,15 @@
...
 import { storiesOf } from '@kadira/storybook';
 
 storiesOf('Bounds', module)
-	.add('default', () => <Bounds />);
+	.add('default', () => (
+		<div style={{width: '100%'}}>
+			<Stripe><Bounds>
+				<h3 className='text--display2'>Bounds</h3>
+				<p>Bounds is used as a non-visual content container that manages content measure and centers children.</p>
+			</Bounds></Stripe>
```
mmcgahan: I'd recommend against over-specifying each Story - they should ideally contain only a single component so that it's clear what style comes from the component rather than parent components. You can make a standalone reference example story that has lots of nested components, but the basic story should be pure.

src/flexItem.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+import { IntlProvider } from 'react-intl';
+import FlexItem from './FlexItem';
+
+function intlRender(component) {
+	return TestUtils.renderIntoDocument(
+		<IntlProvider locale='en-US'>
+			{component}
+		</IntlProvider>
+	);
+}
```
mmcgahan: no need for `intlRender` here - there aren't any translated messages in the components being tested.

src/section.test.jsx
```diff
@@ -1,21 +1,11 @@
...
-		</IntlProvider>
-	);
-}
-
-describe('SectionContainer', function() {
+describe('Section', function() {
 
 	it('exists', function() {
-		const section = intlRender(<SectionContainer />);
+		const section = intlRender(<Section />);
```
mmcgahan: `intlRender` is gone, this story will be the brokeded

assets/scss/section.scss
mperrotti: I don't want to nest that class in the `responsiveBase` mixin because it will get output at each breakpoint

mmcgahan: ah, I was just saying that this could be

```scss
.section {
  .bordered {
    // ...
  }
}
```

instead of

```scss
.section.bordered {
  // ...
}
```

The former is more flexible for future maintenance and parallels the structure inside the mixin, although it's a pretty trivial difference.

src/Section.jsx
mperrotti: I don't remember why this is like this. Simplifying...

src/SectionTitle.jsx
mperrotti: This must have been result of a bad merge or something. I'm only seeing one `<FlexItem>` containing `chlidren`

src/SectionTitle.jsx
mperrotti: This must have been result of a bad merge or something. I'm only seeing one `<FlexItem>` containing `chlidren`

mmcgahan: yeah, i think adam cleaned it up in a different branch, no worries

src/Stripe.jsx
mperrotti: What are you referring to when you say "this"?

mmcgahan: sorry, the colon

```jsx
</Bounds> :
```

instead of

```jsx
</Bounds>
:
```

Stylistically, it's a little unconventional to have a colon alone on a line

mperrotti: Simple fix, just changed. Thanks :)

src/Stripe.jsx
mperrotti: Added

mperrotti:
Addressed feedback from @mmcgahan

assets/scss/stripe.scss
```diff
@@ -1,5 +1,123 @@
...
+---
+A `.stripe` class is rendered by the `<Stripe>` component to
+visually divide the page with borders or stripes of color.
+
+To align content within a stripe, nest children in a
+`<Bounds>` component.
+
+#### Stripe variants
+A class of `.stripe` is default; the following classes are
+optional variants:
```
mmcgahan: out of curiosity, how does this relate to what is already being imported in SQ2? Do we have some redundancy?

mperrotti: This was a temporary way to bring the necessary styles over to get the components into use. I showed Adam what we were doing with component styles when we started the `responsify` branch, and he was interested in keeping styles close to components.
This was a while ago, so I don't recall how he wants to handle this for the future of SQ2.

assets/scss/stripe.scss
```diff
@@ -1,5 +1,123 @@
...
+`.stripe--purple`    |  applies the standard `$C_purple` background color
+`.stripe--plum`      |  applies the standard `$C_plum` background color
+`.stripe--black`     |  applies the standard `$C_black` background color
+`.stripe--photoHero` |  rendered by the `backgroundImage` prop, for hero sections with a photo (see below)
+*/
+
+%stripe,
+.stripe {
+	@include standardBorder();
+	background-color: $C_contentBG;
```
mperrotti: It shouldn't be too much effort to go back and fix this. For the future I'll make sure to only include overrides

akdetrick:
# reminder
This is a PR into `responsify`, not into `master`. If we'd like to get this into `master`, let's merge this and open a new PR into `master`.

Any objections @mperrotti ?

mperrotti:
@akdetrick - I'd be fine with that, except Rick and I are using `responsify` for [the prototype](https://github.com/meetup/meetup-web-prototype/). Is there a way to make sure all the changes make it into there as well?

akdetrick:
For this PR, merging will update the `responsify` branch of this repo.

For future PRs, we could just merge to `master` and maintain the `responsify` branch by merging from `master` back into `responsify`...

eilinora:
There are changes needed. However, we are going to be migrating these over to master in other PR's. So, I'm clearing these for merging. As this work is needed for a Design review and being moved out into a secondary repo

PR #29, opened 2016-12-22
PageHead and related components

mperrotti:


mmcgahan:
We should review things from the top of the dependency tree on down - it's difficult to evaluate the structural changes in this PR without reviewing `PageActions`, `PageAction`, `Flex`, `FlexItem`, `Chunk`, `Section`, etc...

If we can start from the top and then follow on with PRs that combine already-reviewed components, that will make it easier to see the big picture. It will also avoid the awkwardness of having to refactor already-reviewed components when their dependencies change due to CR.

I'm particularly interested in `PageActions` and `PageAction`, which ideally would not have any `context`, and could possibly be combined into a single component since it looks like they wouldn't ever be used independently.

mmcgahan:
just noticed the PR for Flex + Chunk + Section, my bad. PageAction(s) would be really good to separate out though

mperrotti:
The layout of the individual page actions (PageActionButton.jsx) is different if you have 2 or less page actions. Should I just write a test to check if that's working, or does that deserve it's own story under PageActions?

Screenshots included for clarity:
pageactionssample

mmcgahan:
@mperrotti definitely a separate story for each layout, but probably don't have to worry about a unit test for that kind of thing IMO. Looking at the screenshots, it looks like the buttons are to the right of the text in the "< 3" case, and to the left of the text in the "> 3" case, is that right? That looks like a pretty significant usability problem.

eilinora:
There are changes needed. However, we are going to be migrating these over to master in other PR's. So, I'm clearing these for merging. As this work is needed for a Design review and being moved out into a secondary repo

PR #30, opened 2016-12-28
Wc 7 popover component

akdetrick:
(marking prelim until tests are complete)

Popover menu component (wraps links into a drop down menu activated by a button)

screen shot 2016-12-28 at 3 53 57 pm

### DIFF
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom';
 import cx from 'classnames';
 import bindAll from './utils/bindAll';
 
+const POPOVER_MENU_CLASS = 'popover-menu-option';
 /**
  * @module Popover
  */
@@ -44,22 +45,23 @@ class Popover extends React.Component {
 		this.setState({ isActive: false });
 	}
 
+	focusCheck() {
+		const focusedOptionClass = document.activeElement.parentNode.classList;
+		// don't close the popover if we're moving focus to an menu item
+		if (focusedOptionClass && focusedOptionClass.contains(POPOVER_MENU_CLASS)) {
+			return;
+		}
+
+		this.closeMenu();
+	}
+
 	onBlur() {
 		// On blur, browsers always focus `<body>` before moving focus
 		// to the next actual focused element.
 		//
 		// This zero-length timeout ensures the browser will return the
 		// actual focused element instead of `<body>`
-		window.setTimeout(() => {
-			const focusedOptionClass = document.activeElement.parentNode.classList;
-
-			// don't close the popover if we're moving focus to an menu item
-			if (focusedOptionClass && focusedOptionClass.contains('popover-menu-option')) {
-				return;
-			}
-
-			this.closeMenu();
-		}, 0);
+		window.setTimeout(() => this.focusCheck(), 0);
 	}
 
 	onClick(e) {
@@ -88,7 +90,7 @@ class Popover extends React.Component {
 			this.updateFocusBy(-1);
 			break;
 		case 'Enter':
-			if (this._menuItems.get(this.state.selectedIndex)) {
+			if (this._menuItems.get(this.state.selectedIndex) && this._menuItems.get(this.state.selectedIndex).props.onClick) {
 				this._menuItems.get(this.state.selectedIndex).props.onClick(e);
 			}
 			break;
@@ -107,7 +109,7 @@ class Popover extends React.Component {
 			return (
 				<li
 					key={i}
-					className='popover-menu-option arg'
+					className={POPOVER_MENU_CLASS}
 					role='menuitem'
 					onKeyDown={this.onKeyDownMenuItem}
 				>
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -1,0 +1,200 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import cx from 'classnames';
+import bindAll from './utils/bindAll';
+
+/**
+ * @module Popover
+ */
+class Popover extends React.Component {
+	constructor(props) {
+		super(props);
+
+		bindAll(this,
+			'updateFocusBy',
+			'openMenu',
+			'closeMenu',
+			'onClick',
+			'onKeyUp',
+			'onKeyDown',
+			'onClick',
+			'onBlur'
+		);
+
+		this.state = {
+			isActive: false,
+			selectedIndex: 0
+		};
+	}
+
+	updateFocusBy(delta) {
+		const targetIndex = this.state.selectedIndex + delta;
+		const optionsLength = this.props.options.length;
+
+		if (targetIndex >= 0 && targetIndex < optionsLength) {
+			this.setState({ selectedIndex: targetIndex });
+		}
+	}
+
+	openMenu() {
+		this.setState({ isActive: true });
+	}
+
+	closeMenu() {
+		this.setState({ isActive: false });
+	}
+
+	onBlur() {
+		// On blur, browsers always focus `<body>` before moving focus
+		// to the next actual focused element.
+		//
+		// This zero-length timeout ensures the browser will return the
+		// actual focused element instead of `<body>`
+		window.setTimeout(() => {
+			const focusedOptionClass = document.activeElement.parentNode.classList;
+
+			// don't close the popover if we're moving focus to an option
+			if (focusedOptionClass && focusedOptionClass.contains('popover-menu-option')) {
+				return;
+			}
+
+			this.closeMenu();
+		}, 0);
+	}
+
+	onClick(e) {
+		this.openMenu();
+	}
+
+	onKeyDown(e) {
+		switch(e.key) {
+		case 'Enter':
+			if (!this.state.isActive) {
+				this.openMenu();
+				break;
+			}
+			if (this.selectedItemEl && this.selectedItemEl.props.onClick) {
+				this.selectedItemEl.props.onClick(e);
+				break;
+			}
+		case 'Escape':
+			this.closeMenu();
+			break;
+		}
+	}
+
+	onKeyUp(e) {
+		switch(e.key) {
+		case 'ArrowDown':
+			this.updateFocusBy(1);
+			break;
+		case 'ArrowUp':
+			this.updateFocusBy(-1);
+			break;
+		}
+	}
+
+	componentDidUpdate() {
+		if (this.selectedItemEl) {
+			ReactDOM.findDOMNode(this.selectedItemEl).focus();
+		}
+	}
+
+	renderOptionItems() {
+		return this.props.options.map((option,i) => {
+			const isSelected = this.state.isActive && this.state.selectedIndex === i;
+
+			return (
+				<li
+					key={i}
+					className='popover-menu-option'
+				>
+					{/*
+					* treat each user-provided option element as the
+					* keyboard-navigable, focusable 'menuitem' role
+					*/}
+					{
+						React.cloneElement(option,
+							{
+								ref: (el) => {
+									if (isSelected) {
+										this.selectedItemEl = el;
+									}
+								},
+								role: 'menuitem',
+								tabIndex: '-1',
+								onKeyUp: this.onKeyUp,
+								className: 'popover-menu-option-target'
+							}
+						)
+					}
+				</li>
+			);
+		});
+	}
+
+	render() {
+		const isActive = this.state.isActive;
+		const {
+				trigger,
+				options, // eslint-disable-line no-unused-vars
+				className,
+				...other
+			} = this.props;
+
+		const classNames = {
+			popover: cx(
+				'popover',
+				className
+			),
+			trigger: cx(
+				'popover-trigger',
+				{
+					'popover-trigger--active': isActive
+				}
+			),
+			menu: cx(
+				'popover-container',
+				'popover-container--menu',
+				{
+					'display--none': !isActive
+				}
+			)
+		};
+
+		return (
+			<div
+				className={classNames.popover}
+				aria-haspopup='true'
+				onKeyDown={this.onKeyDown}
+				onBlur={this.onBlur}
+				{...other}
+			>
+
+				<div
+					className={classNames.trigger}
+					tabIndex='0'
+					onClick={this.onClick}
+				>
+					{trigger}
+				</div>
+
+				<nav>
+					<ul
+						className={classNames.menu}
+						role='menu'
+						aria-hidden={!isActive}
+					>
+						{this.renderOptionItems()}
+					</ul>
+				</nav>
+			</div>
+		);
+	}
+}
+Popover.propTypes = {
+	trigger: React.PropTypes.element.isRequired,
+	options: React.PropTypes.arrayOf(React.PropTypes.element).isRequired,
+	className: React.PropTypes.string,
+};
+export default Popover;
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -15,7 +15,6 @@ class Popover extends React.Component {
 			'openMenu',
 			'closeMenu',
 			'onClick',
-			'onKeyUp',
 			'onKeyDown',
 			'onClick',
 			'onBlur'
@@ -80,7 +79,7 @@ class Popover extends React.Component {
 		}
 	}
 
-	onKeyUp(e) {
+	onKeyDownMenuItem(e) {
 		switch(e.key) {
 		case 'ArrowDown':
 			this.updateFocusBy(1);
@@ -97,7 +96,7 @@ class Popover extends React.Component {
 		}
 	}
 
-	renderOptionItems() {
+	renderMenuItems() {
 		return this.props.menuItems.map((menuItem, i) => {
 			const isSelected = this.state.isActive && this.state.selectedIndex === i;
 
@@ -120,7 +119,7 @@ class Popover extends React.Component {
 								},
 								role: 'menuitem',
 								tabIndex: '-1',
-								onKeyUp: this.onKeyUp,
+								onKeyDown: this.onKeyDownMenuItem,
 								className: 'popover-menu-option-target'
 							}
 						)
@@ -182,7 +181,7 @@ class Popover extends React.Component {
 						role='menu'
 						aria-hidden={!isActive}
 					>
-						{this.renderOptionItems()}
+						{this.renderMenuItems()}
 					</ul>
 				</nav>
 			</div>
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -29,9 +29,8 @@ class Popover extends React.Component {
 
 	updateFocusBy(delta) {
 		const targetIndex = this.state.selectedIndex + delta;
-		const optionsLength = this.props.options.length;
 
-		if (targetIndex >= 0 && targetIndex < optionsLength) {
+		if (targetIndex >= 0 && targetIndex < this.props.menuItems.length) {
 			this.setState({ selectedIndex: targetIndex });
 		}
 	}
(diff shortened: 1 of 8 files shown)


src/popover.test.jsx
```diff
@@ -7,35 +7,83 @@ import PopoverMenu from './PopoverMenu';
...
+		triggerEl = null;
+		menuEl = null;
+		optionEls = null;
+	});
 
-		it('exists', () => {
-			expect(el).not.toBeNull();
-		});
+	it('exists; menu hidden by default', () => {
+		console.warn(popoverEl);
```
mmcgahan: you shouldn't need to log anything in tests - they should be self-documenting

src/PopoverMenu.jsx
```diff
@@ -0,0 +1,83 @@
...
+	updateFocusBy(delta) {
+		const menuItems = findDOMNode(this).querySelectorAll('.popover-menu-item');
+		const targetIndex = this.selectedIndex + delta;
+
+		if (menuItems[targetIndex] == undefined) {
+			return;
+		}
+
+		this.selectedIndex = targetIndex;
+		menuItems[targetIndex].focus();
```
mmcgahan: this is a DOM-oriented approach to updating UI state, which sort of steps around the React component lifecycle and app structure conventions.

The main indication that something might not be quite right is that you're using `findDOMNode` in order to get an ordered list of particular `children`. It would probably be a better API to have `menuItems` be a `prop` on the `<PopoverMenu>` (propType: `React.PropTypes.arrayOf(React.PropTypes.instanceOf(PopoverMenuItem))` so that you don't have to select the corresponding nodes with `querySelectorAll`, and instead you can use `props.menuItems` directly.

Then, each menu item can have a `selected` prop that it can use to call `focus()` on its own DOM node - you should use the [`ref` prop](https://facebook.github.io/react/docs/refs-and-the-dom.html) on its rendered HTML elements to get the target element for `focus()`.

Advantages:

1. `PopoverMenu` doesn't need to know anything about the DOM
2. `PopoverMenuItem` is responsible for its own 'selected' UI state, which is prop-controllable
3. React component lifecycle stays intact because you'll need to use `setState({ selectedIndex: ... })` to get the menu items to re-render.
4. `PopoverMenuItem` won't be mixed in with other children (if that's even allowed? if it's not allowed to have other types of children, you can add a `propType` for `children` - it's currently ambiguous)

src/PopoverMenuItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+				to={url}
+				{...other}>
+				{children}
+			</Link>
+		);
+	}
+}
+PopoverMenuItem.propTypes = {
+	onClick: React.PropTypes.func
+};
```
mmcgahan: propTypes for `url` and `handleKeyUp`

src/PopoverMenuItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @module PopoverMenuItem
+ */
+class PopoverMenuItem extends React.Component {
+
+	render() {
+		const {
+			children,
+			url,
```
mmcgahan: might be better to call this `to`, since it shouldn't be a fully-qualified URL, but rather a route/location. `to` also makes a better connection to the underlying `<Link>` interface

src/Popover.jsx
```diff
@@ -0,0 +1,111 @@
...
+				onBlur={this.handleBlur}
+				{...other}>
+				{this.renderTrigger()}
+				{this.renderMenu()}
+			</div>
+		);
+	}
+}
+
+export default Popover;
```
mmcgahan: needs propTypes

src/PopoverMenu.jsx
```diff
@@ -0,0 +1,83 @@
...
+					React.Children.map(children, (child) => {
+						return React.cloneElement(child, { handleKeyUp });
+					})
+				}
+			</div>
+		);
+	}
+}
+
+export default PopoverMenu;
```
mmcgahan: needs propTypes

src/PopoverMenuItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+				{children}
+			</Link>
+		);
+	}
+}
+PopoverMenuItem.propTypes = {
+	onClick: React.PropTypes.func
+};
+
+export default PopoverMenuItem;
```
mmcgahan: needs propTypes

src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,42 @@
...
+				onBlur={handleBlur}
+				tabIndex='0'
+				{...other}>
+				{children}
+			</div>
+		);
+	}
+}
+
+export default PopoverTrigger;
```
mmcgahan: needs propTypes

src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+class PopoverTrigger extends React.Component {
+
+	render() {
+		const {
+			children,
+			className,
+			handleClick,
+			handleKeyDown,
+			handleBlur,
```
mmcgahan: The React convention is to name event handlers with an `on..` prefix to match up with the `prop` they get assigned to, e.g. `onClick={onClick}` - just a tiny bit of extra consistency.

src/popover.test.jsx
```diff
@@ -0,0 +1,89 @@
...
+	});
+
+	it('exists; menu hidden by default', () => {
+		console.warn(popoverEl);
+		expect(popoverEl).not.toBeNull();
+		expect(menuEl.classList.contains(class_hidden)).toBe(true);
+	});
+
+	it('menu appears on trigger click', () => {
+		TestUtils.Simulate.click(triggerEl);
```
mmcgahan: for user interaction tests, it's good practice to test the state both before and after the simulated interaction so that you can be sure that something changed. It's not quite as big a deal here because the previous unit test tests the 'before' state, but it's also best practice to keep unit tests completely independent

src/popover.test.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+		expect(menuEl.classList.contains(class_hidden)).toBe(false);
+	});
+
+	it('menu dismissed on popover blur', () => {
+		TestUtils.Simulate.click(triggerEl);
+		TestUtils.Simulate.blur(popoverEl);
```
mmcgahan: this would be a good place to put an expect, between the two simulated interactions, so that you can be sure the state didn't change to the final state before the blur.

src/popover.test.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+		expect(menuEl.classList.contains(class_hidden)).toBe(true);
+	});
+
+	it('menu is keyboard navigatable with arrows', () => {
+		const defaultSelected = optionEls[0];
+		const targetSelected = optionEls[1];
+
+		TestUtils.Simulate.click(triggerEl);
+		TestUtils.Simulate.keyUp(defaultSelected, {key: 'ArrowDown'});
```
mmcgahan: It's worth checking for whether the 'up' arrow works since you have explicit logic for that event as well - should help your code coverage as well

mmcgahan:
sorry, i just saw that this was `Prelim` - apologies if some of the changes i recommended were already on the todo list!

src/Popover.jsx
```diff
@@ -0,0 +1,111 @@
...
+				menu = React.cloneElement(child, { handleKeyDown, isActive });
+			}
+		});
+		return menu;
+	}
+
+	render() {
+		const {
+			className,
+			...other
```
akdetrick: conveniently, React appears to clean this up when the attribute is applied to an element/component

src/Popover.jsx
```diff
@@ -0,0 +1,111 @@
...
+			'popover',
+			className
+		);
+
+		return (
+			<div
+				className={classNames}
+				aria-haspopup='true'
+				onBlur={this.handleBlur}
+				{...other}>
```
akdetrick: i dig it

src/popover.test.jsx
```diff
@@ -0,0 +1,81 @@
...
+		expect(menuEl.classList.contains(class_hidden)).toBe(true);
+	});
+
+	it('menu is keyboard navigatable with arrows', () => {
+		const defaultSelected = ReactDOM.findDOMNode(options[0]);
+		const targetSelected = ReactDOM.findDOMNode(options[1]);
+		const triggerEl = ReactDOM.findDOMNode(trigger);
+
+		TestUtils.Simulate.click(triggerEl);
+		TestUtils.Simulate.keyUp(defaultSelected, {key: 'ArrowDown'});
```
mmcgahan: I think my previous comment got lost, but just to keep it visible: you should call `expect` before and after every simulated user interaction to ensure that the UI is in the intermediate state you expect. Without calling it _before_ the interaction, you can't be certain that something actually changed as a consequence of the simulated interaction.

akdetrick:
Seems likely we'll need to leave this in `Prelim` until I can have another look; after making a few changes, I'm running into some scary stuff:
screen shot 2016-12-30 at 1 40 50 pm

src/PopoverMenu.jsx
akdetrick: The idea was to allow for a generic content container inside a popover, which means you would be able to do either of the following later on:

```
<Popover>
     <PopoverMenu> ... items... </PopoverMenu>
</Popover>
```

```
<Popover>
     <PopoverContent> ... arbitrary content ... </PopoverContent>
</Popover>
```

Now that I'm looking at this though, These should probably just be separate components.

akdetrick: I think `ref` is what I was conceptually overlooking... The above changes make sense to me.

akdetrick:
In regard to moving the trigger and menu from children to props (mentioned in [this comment above](https://github.com/meetup/meetup-web-components/pull/30#discussion_r94084870))...

Consider this usage:
```
const triggerJSX = (
	<PopoverTrigger>Trigger</PopoverTrigger>
);

const menuJSX = (
	<PopoverMenu>
		<PopoverMenuItem to='/fakeurl'>One</PopoverMenuItem>
		<PopoverMenuItem to='/fakeurl'>Two</PopoverMenuItem>
		<PopoverMenuItem to='/fakeurl'>Three</PopoverMenuItem>
	</PopoverMenu>
);

<Popover trigger={triggerJSX} menu={menuJSX} />
```

I suppose this could be simplified like so:
```
<Popover
	triggerLabel='Trigger label'
	menuItems={[
		{ to: '/someurl', label: 'First option' },
		{ to: '/someurl', label: 'Second option' },
		{ to: '/someurl', label: 'Third option' },
	]}
/>
```

The props in the simplified example could be used as a configuration to render child components, but it feels like a custom API for this component where actual child components would work. I could be taking the wrong approach here...

mmcgahan:
@akdetrick the top example in your comment is exactly what I would recommend in order to leverage `propTypes` and make it clear what the expected inputs are for a `<Popover>`. Nesting the trigger and menu as `children` means you have to do extra work to parse them inside the component, and `<Popover>` becomes more opaque and difficult to work with because the `children` don't behave the way they do for regular components

from the comment I linked in my original comment:

> Inventing non-standard behaviors on a per-component basis seems like it should be a last resort, to be used only after you've concluded that established practices isn't a good fit. It shouldn't be all-or-nothing, just because children can be used and looks pretty, doesn't mean they should.
>
> The solution could imaginably be `<Tabs tabs={[...]} />`, it's no longer an opaque structure and the special behavior likely apparent to the user as well (just saying). If you ask me, each tool should be used only for its intended purpose, children has a purpose too and I think it's a mistake to expand that definition unless we're sure it makes sense.

eilinora:
Sorry i'm late to the convo i see the "to" prop being used but we could have menu items that are linking to actions and or full url's

src/PopoverMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+				aria-hidden={!isActive}
+				{...other}
+			>
+				{
+					React.Children.map(children, (child, i) => {
+						const isSelected = isActive && this.state.selectedIndex === i;
+						const newProps = Object.assign({},
+							{ onKeyUp },
+							{ isSelected }
+						);
```
mmcgahan: Looks like you might have over-thought this assignment.

```js
const newProps = { onKeyUp, isSelected };
```

src/PopoverMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+		this.state = {
+			selectedIndex: 0
+		};
+	}
+
+	updateFocusBy(delta) {
+		const targetIndex = this.state.selectedIndex + delta;
+		const itemLength = React.Children.toArray(this.props.children).length;
+
+		if (targetIndex >= 0 && targetIndex <= itemLength) {
```
mmcgahan: `targetIndex` should be `< itemLength` rather than `<= itemLength`,

src/PopoverMenuItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+import ReactDOM from 'react-dom';
+import cx from 'classnames';
+
+/**
+ * @module PopoverMenuItem
+ */
+class PopoverMenuItem extends React.Component {
+
+	componentDidUpdate() {
+		if (this.selectedItem) {
```
eilinora: Might be a little nitpicky but; I would maybe change this up slightly, instead of having logic in the jsx to check if the menuItem is selected instead i would perhaps do that here.

so something like

```
if (this.prop.isSelected) {
  ReactDOM.findDOMNode(this.myLink).focus();
}
```

and may a reference to `myLink`

src/PopoverMenuItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+			ReactDOM.findDOMNode(this.selectedItem).focus();
+		}
+	}
```
mmcgahan: This setup can be simplified by checking for 'selected' status in one place rather than two, and since the `focus` call happens here, I recommend keeping a consistent reference to the `<Link>` instance with `ref={link => this.link = link}` in your `render` function. Then, in `componentDidUpdate`, you can check the props to determine whether or not to focus.

```js
componentDidUpdate() {
  if (this.props.isSelected) {
    ReactDOM.findDOMNode(this.link).focus();
  }
}
```

src/Popover.jsx
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	onBlur() {
+		// On blur, browsers always focus `<body>` before moving focus
+		// to the next actual focused element.
+		//
+		// This zero-length timeout ensures the browser will return the
+		// actual focused element instead of `<body>`
+		window.setTimeout(() => {
+			const focusedElementClass = document.activeElement.getAttribute('class');
```
mmcgahan: we can use `el.classList` freely now

src/PopoverMenuItem.jsx
mmcgahan: oops, jinx.

akdetrick:
Marked as `Prelim` again due to the scope of refactor and missing tests (test are on the way)...

src/Popover.jsx
```diff
@@ -0,0 +1,206 @@
...
+					{trigger}
+				</div>
+
+				<nav>
+					<ul
+						className={classNames.menu}
+						role='menu'
+						aria-hidden={!isActive}
+					>
+						{
```
akdetrick: any preference for breaking out `option` rendering into a separate function for readability?

mmcgahan: i do

src/Popover.jsx
```diff
@@ -0,0 +1,206 @@
...
+				options,
+				className,
+				...other
+			} = this.props,
+			{
+				onClick,
+				onKeyUp,
+				onKeyDown,
+				onBlur
+			} = this;
```
mmcgahan: Nitpick, but destructuring instance properties into new `const`s feels like an over-abstraction that makes usage more ambiguous - leaving them as `this.x` makes it clear that each method is an instance method rather than a standalone function defined somewhere, with the associated expectation of context binding.

akdetrick: Makes sense; I can see where it could be an annoyance for someone reading this later on

src/Popover.jsx
```diff
@@ -0,0 +1,206 @@
...
+
+	componentDidUpdate() {
+		if (this.selectedItemEl) {
+			ReactDOM.findDOMNode(this.selectedItemEl).focus();
+		}
+	}
+
+	render() {
+		const isActive = this.state.isActive,
+			{
```
akdetrick: I was wondering about this; sounds good to me. Chaining seems more reasonable for things like

```
let just,
   initializing,
   some,
   vars,
   lol;
```

src/Popover.jsx
```diff
@@ -0,0 +1,206 @@
...
+	onKeyDown(e) {
+		switch(e.key) {
+		case 'Enter':
+			if (this.state.isActive
+				&& this.selectedItemEl
+				&& this.selectedItemEl.props.onClick
+			) {
+				this.selectedItemEl.props.onClick(e);
+				break;
+			} else {
```
akdetrick: nice; this felt a little messy. The line breaks made me feel better about keeping the expression in the `if` more readable, but is there a better way?

src/Popover.jsx
akdetrick: I'll go with `renderOptionList` as an instance method. Keeping `Popover` a single, self-contained component feels right.

src/Popover.jsx
mmcgahan: I think this is how I'd write it, to completely separate the 'open' logic from the 'fire onClick' logic:

```js
if (!this.state.isActive) {
  this.openMenu();
  return;
}
if (this.selectedItemEl && this.selectedItemEl.props.onClick) {
  this.selectedItemEl.props.onClick(e);
  return;
}
```

However, it does feel a little odd for this parent component to be responsible for firing the `selectedItem`'s `onClick`. I wonder if you could inject an `onKeyDown` into each `option` that takes care of that functionality? It's still a little 'controlling', but at least the `option` fires its own action handlers. That might be over-engineering, though.

nlaz:
Made sort of an opinionated change. Refactored `options` to `menuItems` because `menuItems` is more explicit.

src/popover.story.jsx
```diff
@@ -0,0 +1,39 @@
...
+		);
+	})
+	.add('DIV trigger with SPAN menu items', () => {
+		return (
+			<Popover
+				trigger={<div>Open</div>}
+				menuItems={[
+					<span onClick={logSelection}>First option</span>,
+					<span to='somepath/' onClick={logSelection}>Second option</span>,
+					<span to='somepath/' onClick={logSelection}>Third option</span>,
```
mmcgahan: this `to` would probably throw an error on a JSX `span`? 'unsupported property' or something?

src/Popover.jsx
```diff
@@ -0,0 +1,200 @@
...
+				>
+					{
+						/*
+						* treat each user-provided menu item element as the
+						* keyboard-navigable, focusable 'menuitem' role
+						*/
+						React.cloneElement(menuItem,
+							{
+								tabIndex: '-1',
+								className: 'popover-menu-option-target',
```
nlaz: Setting the className like this overrides the elements className. You can keep the className passed into the prop with:
```
className: `popover-menu-option-target ${menuItem.props.className}`
```

However, I'm curious if we want to prevent that by design. Should we allow users of the Popover component set the className of elements within menu? I'm leaning towards allowing it since it will be useful with SQ2 classes.

mmcgahan: @nlaz and I talked about this on Slack, and decided to gracefully incorporate `menuItem` classes since MWC should try to stay as adaptable as possible.

PR #31, opened 2017-01-04
className should be applied to ul instead of nav element in Tabs

akdetrick:
this ensures tab class variants actually work

### DIFF
--- a/src/Tabs.jsx
+++ b/src/Tabs.jsx
@@ -15,8 +15,7 @@ class Tabs extends React.Component {
 			...other
 		} = this.props;
 
-		const classNames = cx(
-			className,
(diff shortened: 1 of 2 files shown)


sadafie:
also, looks like build failed because tabs tests failed.

mmcgahan:
@akdetrick I'm not sure what editor you use, but I _highly_ recommend integrating ESLint into it to avoid linting failures in the PR builder: https://www.npmjs.com/package/eslint_d#editor-integration

akdetrick:
Yeah, running into this kind of thing a lot - I need to reconfigure my editor for new platform stuff. Thanks for the link!

PR #34, opened 2017-01-12
WC-19: Section

eilinora:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-19

#### Description
Helps, divides page area's into sections, enhancing stories and tests

screen shot 2017-01-12 at 10 14 56 am
screen shot 2017-01-12 at 10 15 05 am


assets/scss/_section.scss
```diff
@@ -0,0 +1,37 @@
...
+---
+Used to group sets of related components
+into discrete sections by adding padding
+around it's children.
+*/
+
+$sectionSpaceMultiple: 2;
+
+@include responsiveBase() {
+	.section {
```
akdetrick: we should be using an additive class pattern with the [sq2 "PCV" convention](https://github.com/meetup/sassquatch2/wiki/Code-standards-&-guidelines#pcv-class-name-convention)...

```
// normal
let className = 'section';

// padded
let className = 'section section--padded`
```

This not only helps us reason about base vs. modifier classes, it also keeps specificity down. You no longer have to nest `padded` inside of `section` in this CSS, instead declaring a basic class, `section--padded`.

src/Section.jsx
```diff
@@ -0,0 +1,41 @@
...
+			children,
+			className,
+			padded,
+			...other
+		} = this.props;
+
+		const classNames = cx(
+			SECTION_CLASS,
+			{
+				[SECTION_PADDED_CLASS]: padded
```
akdetrick: see above comment on PCV class naming convention

assets/scss/_section.scss
```diff
@@ -0,0 +1,37 @@
...
+		padding: $base*$sectionSpaceMultiple $base 0;
+		&.section--padded {
```
akdetrick: it would be good to also unnest this so `section--padded` has the same specificity as `section`. Keeping modifiers at the same specificity level as the base class means that if we declare them lower in the cascade than the base class, the behavior of conflicting properties is always predictable.

```css
.section.section--padded {
    padding: $something;
}
.section--borderedFancy {
    border-top: $size solid cornsilk;
    padding-top: $space; /* this will be overridden by the overqualified modifier class above */
}
```

PR #36, opened 2017-01-12
WC-25: TabsTab - adding other proptypes for labels

eilinora:
We are using other proptypes for labels like formatMessage, making list of acceptable labels more flexible.

### DIFF
--- a/src/TabsTab.jsx
+++ b/src/TabsTab.jsx
@@ -34,7 +34,10 @@ class TabsTab extends React.Component {
 }
 TabsTab.propTypes = {
 	url: React.PropTypes.string.isRequired,
-	label: React.PropTypes.string.isRequired,
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element,
+	]).isRequired,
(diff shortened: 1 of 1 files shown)


src/TabsTab.jsx
```diff
@@ -34,7 +34,10 @@ class TabsTab extends React.Component {
 }
 TabsTab.propTypes = {
 	url: React.PropTypes.string.isRequired,
-	label: React.PropTypes.string.isRequired,
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element,
```
nathanstilwell: This looks good to me, in reference to the [ticket](https://meetup.atlassian.net/browse/WC-25) we would just add more types as needed later?

eilinora: yeah, we can refine the types. I'm not sure at the moment which ones other that FormatMessage and Link we would have here, but at the sametime it would be nice it was flexible to accept any time

nathanstilwell: Cool. LGTM. :shipit:

PR #37, opened 2017-01-12
Popover menu alignment options

nlaz:
#### Related Issues
Implements https://meetup.atlassian.net/browse/WC-26

#### Details
Added props that will set the appropriate classes within the `Popover` component.


src/Popover.jsx
```diff
@@ -200,4 +204,9 @@ Popover.propTypes = {
 	className: React.PropTypes.string,
 };
 
+Popover.defaultProps = {
+	alignLeft: false,
+	alignRight: false,
```
mmcgahan: add `propTypes` for these as well

eilinora: also you don't really need to set defaults as they are bool's it will evaluate to false on the cx() checks

src/Popover.jsx
```diff
@@ -139,6 +139,8 @@ class Popover extends React.Component {
 		const {
 				trigger,
 				menuItems, // eslint-disable-line no-unused-vars
+				alignLeft,
```
eilinora: Tests :D

src/Popover.jsx
```diff
@@ -139,6 +139,8 @@ class Popover extends React.Component {
...
+				alignRight,
```
akdetrick: instead of two different alignment props, could we instead do something like this?

```
<Popover ... align='left' />
```

```
React.propTypes = {
     align: React.PropTypes.oneOf(['right', 'left'])
}
```

This would allow us to add more alignment types in the future (eg. 'center', 'bottom') and is more likely to match up with an "align" prop signature in other components.

nlaz: Good suggestion. I like this approach 👍

eilinora:
changes LGTM 🚀

PR #39, opened 2017-01-12
WC-27 seperating sassquatch

eilinora:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-27

#### Description
Separating sassquatch from main.scss file so we can simply import that file in other *-web platforms


nathanstilwell:
Looks fine, let 'er rip. 🚢

PR #40, opened 2017-01-12
Fix Avatar links

nlaz:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-28

#### Description
This PR checks whether `href` or `to` props are passed into the component and conditionally returns a `<a>` or `<span>` element. This should fix the problem of not being able to click a `<span>` element in `mup-web`.

#### Prelim
Adding the Prelim tag until I confirm this works on `mup-web`.

### DIFF
--- a/src/Avatar.jsx
+++ b/src/Avatar.jsx
@@ -55,18 +55,10 @@ class Avatar extends React.Component {
 		};
 
 		if (this.props.to || this.props.href) {
-			return (
-				<a {...allProps}>
-					{alt}
-				</a>
-			);
+			return <a {...allProps}>{alt}</a>;
 		}
 
-		return (
-			<span {...allProps}>
-				{alt}
-			</span>
-		);
+		return <span {...allProps}>{alt}</span>;
 	}
 }
 
--- a/src/Avatar.jsx
+++ b/src/Avatar.jsx
@@ -54,11 +54,19 @@ class Avatar extends React.Component {
 			...other
 		};
 
-		return (
-			<span {...allProps}>
-				{alt}
-			</span>
-		);
+		if (this.props.to || this.props.href) {
+			return (
+				<a {...allProps}>
+					{alt}
+				</a>
+			);
+		} else {
+			return (
+				<span {...allProps}>
+					{alt}
+				</span>
+			);
+		}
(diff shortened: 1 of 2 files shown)


src/Avatar.jsx
```diff
@@ -54,11 +54,19 @@ class Avatar extends React.Component {
...
-				{alt}
-			</span>
-		);
+		if (this.props.to || this.props.href) {
+			return (
+				<a {...allProps}>
+					{alt}
+				</a>
+			);
+		} else {
```
mmcgahan: no need for `else` if the `if` returns

nlaz: good call 👍

nlaz:
Verified this change works in `mup-web`. Removed the `Prelim` tag.

src/avatar.test.js
```diff
@@ -37,4 +37,23 @@ describe('Avatar', function() {
...
+
+		expect(avatarComponent.tagName.toLowerCase()).toBe('a');
+	});
+	it('renders a link element if `to` passed in', () => {
+		const avatar = TestUtils.renderIntoDocument(<Avatar className='test' to='/' />);
+		const avatarComponent = TestUtils.findRenderedDOMComponentWithClass(avatar, 'test');
+
+		expect(avatarComponent.tagName.toLowerCase()).toBe('a');
+	});
+	it('renders a span element if `href` passed in', () => {
```
mmcgahan: I think this description should be `renders a span element if no 'href' or 'to' passed in`

src/Avatar.jsx
```diff
@@ -54,6 +54,14 @@ class Avatar extends React.Component {
 			...other
 		};
 
+		if (this.props.to || this.props.href) {
+			return (
+				<a {...allProps}>
+					{alt}
+				</a>
+			);
```
mmcgahan: trivial, but this is a lot of line breaks for a small element - unless you expect a lot of embellishment of the `<a>`, I recommend keeping it tight to keep the overall component more readable

```jsx
return <a {...allProps}>{alt}</a>
```

nlaz: I also like it concise like this

PR #48, opened 2017-01-19
dependencies into dev and peer dependencies

mattkime:

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,14 @@
     "url": "https://github.com/meetup/meetup-web-components/issues"
   },
   "homepage": "https://github.com/meetup/meetup-web-components#readme",
-  "dependencies": {
+  "peerDependencies": {
+    "classnames": "^2.2.5",
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "react-intl": "^2.2.3"
+  },
+  "devDependencies": {
+    "@kadira/storybook": "2.35.3",
     "babel-cli": "6.18.0",
     "babel-core": "6.21.0",
     "babel-jest": "18.0.0",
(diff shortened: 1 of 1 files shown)


mattkime:
>can you explain why these changes please?

we need to use peerDependencies to make sure we're not sending two different versions of the same module to the browser. there have been points where we send different versions of rxjs or react to the browser due to differences between top level and module requirements.

sadafie:
How did you go about determining what should be a peer dependency? Were you looking solely at mup-web, or the other consumer repos as well?

mattkime:
>How did you go about determining what should be a peer dependency? Were you looking solely at mup-web, or the other consumer repos as well?

I looked at the overlap between `mup-web`, `meetup-web-components`, and `meetup-web-platform`

Thankfully the result is rather simple, based around `react`, `rxjs`, and `classnames`

mattkime:
I could look at other consumer repos as well but they shouldn't vary from whats outlined above

package.json
```diff
@@ -26,7 +26,14 @@
     "url": "https://github.com/meetup/meetup-web-components/issues"
   },
   "homepage": "https://github.com/meetup/meetup-web-components#readme",
-  "dependencies": {
+  "peerDependencies": {
+    "classnames": "^2.2.5",
```
sadafie: See related issue here: https://github.com/meetup/meetup-web-mocks/issues/12

But we shouldn't use `^` notation in peerDependencies since they don't cover major version updates.

sadafie:
@mattkime now that we've figured out how we want to handle dependencies in the mocks repo, i'm circling back to this one.

Why did all the dependencies move into devDependencies? Not sure I understand / follow that bit.

mattkime:
>Why did all the dependencies move into devDependencies? Not sure I understand / follow that bit.

`dependencies` should only consist of modules required to use/consume this module. in this instance, any required modules were already `peerDependencies`. anything else was being used to build / lint / test.

sadafie:
But will this work when being building a docker container, where environment is set to production?

mattkime:
>But will this work when being building a docker container, where environment is set to production?

yes, that is absolutely necessary.

mattkime:
@sadafie I'm closing this since much of this was completed in another PR but I'm still happy to talk about any questions / concerns.

PR #50, opened 2017-01-19
use direct child selectors for popover option children styling

akdetrick:
This fixes a bug where indirect children of `.popover-menu-option` were being unintentionally styled as options.


nlaz:
LGTM 🎉 🚢 ⚡️

PR #51, opened 2017-01-19
refactor tabs to accept tab items as a prop

akdetrick:
**removes `Link` component as a dependency of `Tabs`**

#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-29

#### Description
Instead of using a separate `TabsTab` component clumsily as required `children` of `Tabs`, this PR refactors the `Tabs` component to accept tabs as a prop (array of elements).


mmcgahan:
ignore failed coverage report - there are fewer lines to cover, but `src/Tabs.jsx` is still at 100% coverage so there's no problem.

PR #55, opened 2017-01-20
set peerDependencies / devDependencies

mattkime:
the ongoing saga of dependency cleanup

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,19 @@
     "url": "https://github.com/meetup/meetup-web-components/issues"
   },
   "homepage": "https://github.com/meetup/meetup-web-components#readme",
-  "dependencies": {
+  "peerDependencies": {
+    "classnames": "^2.2.5",
+    "react": "^15.4.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,38 +26,43 @@
...
-    "babel-loader": "6.2.10",
-    "coveralls": "2.11.15",
-    "css-loader": "0.26.1",
-    "eslint-loader": "1.6.1",
-    "inquirer": "2.0.0",
-    "jest-cli": "18.1.0",
-    "npm-run-all": "4.0.1",
+    "react-router": "3.0.2",
     "sass-loader": "4.1.1",
+    "sassquatch2": "2.4.4",
```
mmcgahan: This should also be a peer dependency to avoid competing sassquatches

mattkime: addressed. but it also needs a devDep reference for storybook. it appears to be common that peerDependencies need a devDependency version as well if they're used in testing.

PR #57, opened 2017-01-23
Mup 7632 text field

unjust:
#### Related issues

Fixes https://meetup.atlassian.net/browse/MUP-7632

#### Description

Preliminary pull request to get feedback.

This TextField is actually a Label component, input tag, and `p.class='text--error' element put together.

Questions/issues:

* Is TextField a good name? 
Should the text input inside TextField be broken off into another TextInput component?

* TextField is unable to show keyboard input in storybook. I see the state updating when I type but getting overwritten. Am I doing something wrong with the way state is maintained and updated? (I'm using the controlled component pattern that React suggests https://facebook.github.io/react/docs/forms.html)

* is having a dynamic `formAttrs` property too open ended? Should each be specified?

* I've used some form markup in some of the stories to demo the html5 validation - ie to see what happens when a required field isn't filled in. Is this the best way to demo that?

* The HTML5 / browser messages sure are ugly and aren't too specific sometimes. Should we support pattern validation and use the browser messages?
screen shot 2017-01-23 at 12 40 41 pm


src/TextField.jsx
```diff
@@ -0,0 +1,58 @@
...
+			name,
+			label,
+			placeholder,
+			formAttrs,
+			children,
+			className,
+			...other
+		} = this.props;
+
+		const classNames = cx(
```
eilinora: Are you going to be adding custom classes? If not than you don't need to do this cx call

src/TextField.jsx
```diff
@@ -0,0 +1,58 @@
...
+			className
+		);
+
+		return (
+			<div>
+				<label htmlFor={elId}>{label}</label>
```
eilinora: might be a nice enhancement to allow for label and div custom classnames, i know we do that a lot for http://meetup.github.io/sassquatch2/forms.html#inputStack

src/textField.test.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+import TextFieldContainer from './TextFieldContainer';
+
+describe('TextFieldContainer', function() {
+
+	it('exists', function() {
```
eilinora: maybe some more tests :D

src/TextField.jsx
```diff
@@ -0,0 +1,58 @@
...
+				<input
+					type='text'
+					id={elId}
```
mmcgahan: maybe `inputId` instead of `elId` since there are lots of `el`s

unjust: cool. I can follow whatever convention you guys have already established.

src/textField.test.jsx
unjust: yes

src/Label.jsx
```diff
@@ -0,0 +1,37 @@
...
+				{children}
+			</label>
+		);
+	}
+}
+
+Label.propTypes = {
+};
+
+export default Label;
```
mmcgahan: it looks like `<Label>` doesn't do much more than a regular `<label>` - it just changes the name `htmlFor` attribute to `inputId` and adds a switch for the `.required` classname.

IMO, the attribute/prop renaming obscures the implementation more than necessary, and at that point, writing a component for the sake of a className on an element is probably overkill

src/TextField.jsx
```diff
@@ -0,0 +1,75 @@
...
+		const {
+			inputId,
+			name,
+			label,
+			labelClassName,
+			placeholder,
+			formAttrs,
+			className,
+			children,
+			...other
```
mmcgahan: Since you're passing `...other` to the `<input>`, I think it might be simpler to leave out any mention of `placeholder` and `name` - they will be populated correctly by `other`. Similarly, you can remove `inputId` and just let the parent component specify `id`, which will be applied to the `<input>` by `other`. The fewer props that have to be renamed before they get to the HTML output, the better.

src/Label.jsx
unjust: ok. I'd rather not create another component if its not necessary. @adamd do you agree with Mike? I can just keep `label` and html element then. 

@mikem-mup any thought on breaking out the html text input into its own component while TextField (can rename) would be something more compound with label, error, and possibly char counter?

mmcgahan: I like the compound component `TextField` setup, although at the moment it doesn't look like the `<input>` itself needs much special treatment, so I'd probably avoid making a separate component for that.

src/TextField.jsx
unjust: ok so its fine to be less explicit?
should I add some documentation?

unjust: I'm thinking name should be a required attribute. Is that possible to make it required using `other`?

mmcgahan: > ok so its fine to be less explicit?
inside the `render` function, yes.

> should I add some documentation?
Yeah, you could probably mention that attributes passed to the component will be sent to the `<input>` internally. You can also add some `propTypes` to make certain attributes required.

> Is that possible to make it required using `other`?
Yes - `other` is just an alias used inside `render`. The component `propTypes` apply to any `this.props`.

> And are you suggesting to just use `other` for any attributes and get rid of `formAttrs` as well
Yes - feel free to add `propTypes` for anything that you want to require/validate

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+		);
+
+		const labelClassNames = cx(
+			{ required : other ? other.required : false },
+			labelClassName
+		);
+
+		const errorElClassNames = cx(
+			'text--error',
+			{ 'display--none': !this.state.error }
```
mmcgahan: in React, it's _much_ better to just skip rendering entirely for elements that are hidden under certain conditions. In the JSX, use this pattern instead of setting a `classNames` var:

```jsx
{ !this.state.error && <p className='text--error'>{error}</p> }
```

src/textField.test.jsx
```diff
@@ -0,0 +1,62 @@
...
+		expect(labelEl.textContent).toEqual(LABEL_TEXT);
+	});
+
+	it('should have a required attribute when specified', () => {
+		expect(inputEl.attributes.required).not.toBeNull();
+	});
+
+	it('should specify attributes that are passed in', function() {
+		expect(inputEl.getAttribute('maxLength')).toEqual(MAX_LEN);
+	});
```
mmcgahan: this needs another test for error rendering

mmcgahan: and a test for calling `handleChange` - check out [the docs for `TestUtils.Simulate`](https://facebook.github.io/react/docs/test-utils.html#simulate) to see how to simulate an input change, and then check the `textField.state` before and after the change - you'll need to make sure that reference to the React element (not just the DOM node) is in scope for the test function.

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+		);
+
+		return (
+			<div>
+				<label className={labelClassNames} htmlFor={other.id}>
+					{label}
+				</label>
+
+				<input type='text'
+					value={this.state.value}
```
akdetrick: could use a `name` and `id` attribute of `{other.id}`

mmcgahan: `{...other}` takes care of `id={other.id}`, but I think you could just make `name` a required `propType` and let devs set the name like they would do with other inputs?

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+ * @module TextField
+ */
+class TextField extends React.Component {
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			value: props.value,
+			error: props.error
+		};
```
mmcgahan: you only need something in `this.state` if you are going to be dynamically updating it independently of `props` passed in by the parent. Currently, errors are _always_ provided by the parent, but if you develop this to add something like a char counter that would have to show an error while someone is typing, it would be good to add `error` to `this.state` at that point. Probably better _not_ to have it in `this.state` until you need it for `this.setState()`

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+					{...other} />
+
+				<p className={errorElClassNames}>{error}</p>
+				{children}
+			</div>
+		);
+	}
+}
+
+TextField.propTypes = {};
```
mmcgahan: The convention we're going with is to _not_ include any html-attribute props in `propTypes` unless you want to make one of them required. All 'custom' props like `label` and `labelClassName` should have propTypes

mmcgahan:
Sorry, I should have read the description to respond to your questions:

> Is TextField a good name?

It's a little ambiguous because it's not clear whether it's a drop-in replacement for an `<input>` or a `<textarea>` (or both). I think if it's intended to _always_ be used when a UI needs an `<input>`, I would call it `<TextInput>`. We'll probably have another component for long-input `<textarea>` cases?

> Should the text input inside TextField be broken off into another TextInput component?

no

> TextField is unable to show keyboard input in storybook. I see the state updating when I type but getting overwritten. Am I doing something wrong with the way state is maintained and updated? (I'm using the controlled component pattern that React suggests https://facebook.github.io/react/docs/forms.html)

Probably not - as long as you have some stories for empty, partially filled, and 'too-long' states, that's fine.

> is having a dynamic formAttrs property too open ended? Should each be specified?

I think we've cleaned this up, but yeah `formAttrs` is probably too loose

> I've used some form markup in some of the stories to demo the html5 validation - ie to see what happens when a required field isn't filled in. Is this the best way to demo that?

> The HTML5 / browser messages sure are ugly and aren't too specific sometimes. Should we support pattern validation and use the browser messages?

For the first pass, I think it's okay to let consumers pass in their own `pattern` prop, which will be passed automatically to the `<input>`. One option is to have an optional `validate` prop, which would be a function that is called inside `handleChange` containing the value of the input, which would allow the parent to run custom validation functions in addition to the 'basic' built-in functionality.

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+	constructor(props) {
+		super(props);
+		this.state = {
+			value: props.value,
+			error: props.error
+		};
+		this.handleChange = this.handleChange.bind(this);
+	}
+
+	handleChange(e) {
```
mmcgahan: this should be named `onChange` since it's the `onChange` handler for the input

src/TextField.jsx
unjust: ok. I think @adamd wanted to try to have a charcounter sketched out in this branch. so I'll fix this up depending on that.

src/textField.test.jsx
unjust: done

src/TextField.jsx
```diff
@@ -0,0 +1,68 @@
...
+				<label className={labelClassNames} htmlFor={other.id}>
+					{label}
+				</label>
+
+				<input type='text'
+					name={name}
+					value={this.state.value}
+					className={classNames}
+					onChange={this.onChange}
+					{...other} />
```
unjust: ah ok, so its ok not to use a var? its just a linter thing?

I extracted it and used it in a console log 😂

mmcgahan: ah okay yeah totally - the only reason the linter complains about this usage is that technically the destructured object assignment syntax is not an official standard yet. when it does become standard, ESLint will stop complaining but until then we have to tell it to be quiet manually

src/textField.test.jsx
unjust: SO I think I did this - not sure how to get rid of the request for changes.

src/TextField.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+				<input type='text'
+					name={name}
+					value={this.state.value}
+					required={required}
+					className={classNames}
+					onChange={this.onChange}
+					{...other} />
+
+				{ this.props.maxLength && <p className='text--caption align--right'>{this.state.value.length} / {this.props.maxLength}</p> }
```
mmcgahan: seems like this should trigger an error state when the length is greater than the max? If not, you don't need `this.state`

src/TextField.jsx
```diff
@@ -0,0 +1,72 @@
...
+				{ this.state.error && <p className='text--error'>{error}</p> }
+				{children}
+			</div>
+		);
+	}
+}
+
+TextField.propTypes = {
+	name: React.PropTypes.string.isRequired
+};
```
mmcgahan: there are a few more prop types to define here - `error`, `labelClassName`, `required`

src/textField.test.jsx
mmcgahan: the reviewer has to approve changes or dismiss the review

src/TextField.jsx
unjust: so right now I'm not sure what the character counter should do. the text field has a max length so you cant type beyond it anyways. maybe I will just remove state.error then until it becomes a requirement.

mmcgahan:
Any more thoughts about naming? It would be nice to make it clear whether this is the right component to use when you need a `<textarea>`-like input

src/TextField.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * @module TextField
+ */
+class TextField extends React.Component {
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			value: props.value || '',
```
unjust: I think this is how value should be set so you don't have to specify it as a prop if `undefined` and so you won't get the controlled/uncontrolled component warning.

mmcgahan: yep, looks good

src/TextInput.jsx
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+TextInput.propTypes = {
+	id: React.PropTypes.string,
+	name: React.PropTypes.string.isRequired,
+	error: React.PropTypes.string,
+	label: React.PropTypes.string,
+	labelClassName: React.PropTypes.string,
+	value: React.PropTypes.string,
```
mmcgahan: up to you, but I've generally been using the convention that any optional 'html native' attributes passed as props don't need to be specified in `propTypes`, but that is totally a subjective call on my part. My reasoning is that web devs are familiar enough with things like `id`, `name`, and `value` that they should expect them to be strings, and there are technically a lot more 'legal' attribute-props that could be passed in here that we shouldn't have to think about validating... totally up to you, and there's minimal cost to validating the props you specifically handle

src/TextInput.jsx
```diff
@@ -0,0 +1,75 @@
...
+				{children}
+			</div>
+		);
+	}
+}
+
+TextInput.propTypes = {
+	name: React.PropTypes.string.isRequired,
+	error: React.PropTypes.string,
+	label: React.PropTypes.string,
```
eilinora: Sorry this was missed initial review but label should also be made a bit more flexible to take in an Element along with a string. You can get a formatted string using, but easiest IMHO if this allowed us to pass <FormatMessage />

src/textInput.test.jsx
```diff
@@ -11,34 +10,32 @@ describe('TextInput', function() {
...
 
 	afterEach(() => {
-		textInputEl = null;
+		textInputComponent = null;
 		inputEl = null;
 	});
 
 	it('exists', () => {
-		expect(textInputEl).not.toBeNull();
+		expect(inputEl).not.toBeNull();
```
eilinora: Unfortunately this test will not capture the error property, you will want to do something like this instead (event for tag) as findRenderDOM always expects to find at least 1 element or throws an error. Where as https://facebook.github.io/react/docs/test-utils.html#scryrendereddomcomponentswithtag will return an empty array and you can instead test the array length. Both are good options for this.

src/textInput.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+		expect(inputEl.name).toEqual(NAME_ATTR);
+	});
+
+	it('should have a value when one is specified', () => {
+		expect(inputEl.value).toEqual(VALUE);
+	});
+
+	it('should have a label when label is given', () => {
+		const labelEl = TestUtils.findRenderedDOMComponentWithTag(textInputComponent, 'label');
+		expect(labelEl).not.toBeNull();
```
eilinora: this won't be null. We should update the test to match exists setup https://github.com/meetup/meetup-web-components/pull/57/files#diff-0c6dbf0ccf27a396991373907ec031d3R38

src/textInput.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+		expect(inputEl.attributes.required).not.toBeNull();
+	});
+
+	it('should specify attributes that are passed in', function() {
+		expect(inputEl.getAttribute('maxLength')).toEqual(MAX_LEN);
+	});
+
+	it('should have an error when one is specified', function() {
+		const errorEl = TestUtils.findRenderedDOMComponentWithClass(textInputComponent, 'text--error');
+		expect(errorEl).not.toBeNull();
```
eilinora: same as above

unjust: derp

eilinora:
🚀

unjust:
@mikem, @eilinora - @mperrotti asked about support for setCustomValidity. I added something in the latest commit. I'm still unsure about how we will be doing validation - either showing errors inline under the field or allowing the browser to validate. Is this code is too early to add until we come up against this requirement and get validation sussed out? or is it good just to have the option in now?

mmcgahan:
@unjust I'd leave out custom validity hooks unless @akdetrick and the design team has settled on using browser-native UI for errors instead of custom error message UI - my impression was that we were going to always use custom UI. If there's an immediate need for that functionality, fine to leave in as-is, but it might create confusion if there are two ways of showing error states

unjust:
ok I'll revert the changes. for some reason it was erroring that 
`TypeError: this.input.setCustomValidity is not a function` anyways...

PR #58, opened 2017-01-23
WP-179: add swarm icons and cleanup

eilinora:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WP-179

#### Description
Cleaning out old icon setups, adding icons to decorator, adding appropriate swarm and raw loader dependencies 
-- nice  benefit no more console errors from TestIconSprites

screen shot 2017-01-23 at 1 23 41 pm


src/utils/decorators.jsx
```diff
@@ -3,9 +3,21 @@ import { IntlProvider } from 'react-intl';
...
+	 * raw SVG sprite from `swarm-icons`
+	 */
+	const iconSpriteStyle = { display: 'none' };
+	const iconSprite = require('raw-loader!swarm-icons/dist/sprite/sprite.inc');
+
 	return (
 		<IntlProvider locale={locale}>
-			{story()}
+			<div>
+				<div style={iconSpriteStyle} dangerouslySetInnerHTML={{__html: iconSprite}} />
```
nlaz: Why does this need to set the inner html?

eilinora: `IntlProvider` only allows for 1 child node :(

mmcgahan: the `<div>` is the single child, so that's fine. This uses `dangerouslySetInnerHTML` because the `iconSprite` is an HTML string fragment in a separate file - not a React component - so [this is how to load it](https://github.com/meetup/mup-web/blob/28bb929fa5f92bec244066304b77d64750c4fe77/src/components/PageWrap.jsx#L90)

PR #59, opened 2017-01-23
Remove popover overflow style

nlaz:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MUP-7694

#### Description
The popover menu had a weird padding on _some_ screens. Removing this overflow style attribute seemed to do the trick, but I'm not sure why. After chatting with @akdetrick, this attribute might have been unintentional, so I don't see the harm in removing it.


eilinora:
LGTM!

PR #61, opened 2017-01-24
MW-617: Beefing storybook w/ addon's - adding info panel, notes panel, controls, and i18N controls

eilinora:
#### Related issues

Fixes https://meetup.atlassian.net/browse/MW-617

#### Description
DoD:
Adding easier note taking, controls for PropTypes, and documentation for code examples.

screen shot 2017-01-24 at 4 59 58 pm
screen shot 2017-01-24 at 4 59 51 pm
screen shot 2017-01-24 at 4 41 38 pm
screen shot 2017-01-24 at 4 41 23 pm
screen shot 2017-01-24 at 5 00 36 pm
screen shot 2017-01-24 at 4 41 18 pm


src/footer.story.jsx
```diff
@@ -1,14 +1,20 @@
...
-			style={{ width: '100%' }}>
-			<Footer>Footer content</Footer>
-		</Annotate>
-	));
+	.addWithInfo(
+		'default',
+		'This is the basic usage with the component.',
+		() => (
+			<InfoWrapper>
+				<Annotate
```
mmcgahan: Presumably we can deprecate `Annotate` now that we have `notes`?

eilinora:
@mmcgahan From playing with bluekit (WHY do i always call it blueprint! :) ), i think it may not be the right solution for us. Reasons: it's gulp vs webkit, little to no documentation, and struggles with multi exports, and doesn't offer options for excluding tests & subfolders. Still evaluating it to make sure I'm not missing something. 

But in meantime, i wanted to see if the things we like about Bluekit is available in storybook and looks to be. Like generating auto documentation. :)

willh-meetup:
FWIW, @eilinora , blueprint is totally a better name than bluekit.

PR #64, opened 2017-01-27
WC-22 page head actions

eilinora:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-22

#### Description
- Adding PageHead & Actions Button
- Updating Button component with Flex/FlexItem and it's corresponding tests
- Fixing styles for info container

screen shot 2017-01-26 at 7 11 52 pm
screen shot 2017-01-26 at 7 11 48 pm
screen shot 2017-01-26 at 7 11 44 pm
screen shot 2017-01-26 at 7 11 40 pm


src/PageHead.jsx
```diff
@@ -0,0 +1,88 @@
...
+				>
+					<FlexItem>
+						<Chunk className='align--center atMedium_align--left'>
+							<h1 className='text--display1'>{title}</h1>
+							{subtitle &&
+								<p className='text--secondary'>{subtitle}</p>
+							}
+						</Chunk>
+					</FlexItem>
+					{Boolean(menuItems.length) &&
```
nlaz: `menuItems.length > 0` might be cleaner and easier to understand?

src/PageHead.jsx
```diff
@@ -0,0 +1,88 @@
...
+						<FlexItem shrink>
+							<Flex justify='around' className={PAGE_ACTIONS_CLASS}>
```
nlaz: didn't realize this until now but the justify type `around` is sort of ambiguous. Can we change the Flex prop to accept `spaceAround` since it's closer to the css attribute `justify-content: space-around;`?

assets/scss/_pageHead.scss
```diff
@@ -0,0 +1,10 @@
+@include responsiveBase() {
+	.pageHead {
+		padding-bottom: $base*2 - $base;
+	}
+
+	.pageActionButton {
+		min-height: $base*2.25*1.25; // text--display height * line-height
```
nlaz: Idk if this is possible but if sass multiplying variables: 
```
min-height: $base*$line-height-largeText*2.25;
```

assets/scss/_pageHead.scss
```diff
@@ -0,0 +1,10 @@
+@include responsiveBase() {
+	.pageHead {
+		padding-bottom: $base*2 - $base;
```
nlaz: Hmmm `2x - 1x = x`?  or rather `$base*2 - $base = $base`?

assets/scss/_pageHead.scss
```diff
@@ -0,0 +1,10 @@
...
+	}
+
+	.pageActionButton {
```
nlaz: Is there a reason why this style is include here rather `_pageActionButton.scss`?

nlaz:
@eilinora Are `PageHead` and `PageActionButton` always met to be used together or are they designed to be used separately? Could I have a `PageHead` without a `PageActionButton` and vice versa?

assets/scss/_pageActionButton.scss
```diff
@@ -0,0 +1,5 @@
+.pageActionButton {
+	border:none;
+	border-radius:0;
+	width:100%;
```
nlaz: for style consistency: add a space between these fields and their values.

nlaz:
You might want to tweak the breakpoints:
screen shot 2017-01-27 at 12 17 46 pm

src/PageActionButton.jsx
```diff
@@ -0,0 +1,82 @@
...
+			switchDirection,
+			justify;
+
+		// if column
+		// icons and labels in rows,
+		// adjust to take up space based on position and numbers
+		//
+		//  OTHERWISE:
+		// always a nice column of actions
+		// each action is a horizontal icon and label
```
nlaz: Sorry to be particular here but it's unclear what exactly this comment mean. Can you clarify this a bit?

src/button.test.jsx
```diff
@@ -63,73 +64,53 @@ describe('Button', () => {
...
 		afterEach(() => {
-			buttonEl = null;
+			button = null;
 		});
 
-		it('creates a first-child icon element', () => {
-			const iconEl = buttonEl.firstChild;
-			expect(iconEl.classList.contains(BUTTON_ICON)).toBe(true);
+		it('should render a `Flex` container for icons and label', () => {
+			const flex = findComponentsWithType(button, 'Flex');
```
mmcgahan: testing for Flex/FlexItem is sort of okay, but doesn't seem to me like a very useful unit test since the `<Flex>` usage seems more like an implementation detail as opposed to something that is fundamental to a button being a button. Put another way, if the button was changed to use something other than `<Flex>`, should we require that tests be updated? I'm sort of ambivalent, but in general would just want there to be a test for the existence of an icon and a test for the existence of a label - those are baseline requirements for a button. Beyond that, you just have to make a call how much you want to 'set' the implementation requirements.

At a minimum, probably worth breaking out separate tests for `label` and `icon` independent of the component that wraps them.

src/pageHead.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+			const flexItems = findComponentsWithType(pageHead, 'Flex');
+			const pageTitle = flexItems.filter(flex => flex.props.className.indexOf(PAGE_TITLE_CLASS) !== -1);
+			expect(pageTitle.length).toBe(1);
+		});
+		it(`should NOT have a 'Flex' component with a class of ${PAGE_ACTIONS_CLASS}`, () => {
+			const flexItems = findComponentsWithType(pageHead, 'Flex');
+			const pageActions = flexItems.filter(flex => flex.props.className.indexOf(PAGE_ACTIONS_CLASS) !== -1);
+			expect(pageActions.length).toBe(0);
+		});
+		it('should NOT have a \'<p>\' tag', () => {
```
mmcgahan: why can't this PageHead ever contain a `<p>`? Is this intended to ensure that the `subtitle` is not rendered? If so, give the subtitle a classname and check for the non-existence of that

src/pageHead.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+			);
+			pageHeadEl = ReactDOM.findDOMNode(pageHead);
+		});
+
+		it('should display a sub title area', () => {
+			expect(pageHeadEl.innerHTML.indexOf(subtitle)).not.toBe(-1);
+		});
+
+		it('should have a \'<p>\' tag with a class of \'text--secondary\'', () => {
+			const subtitleEl = pageHeadEl.getElementsByTagName('P');
```
mmcgahan: does this have to be implemented with a `<p>` tag? Seems over-prescriptive and it would be better if you could just check for the existence of an element with a `.pageHead-subtitle` class or something - let the implementation change but ensure that a suitable element is rendered

src/pageHead.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+		beforeEach(() => {
+			pageHead = TestUtils.renderIntoDocument(
+				<PageHead title={pageTitle} menuItems={menu} />
+			);
+			pageHeadEl = ReactDOM.findDOMNode(pageHead);
+		});
+
+		it(`should have a 'Flex' component with a class of ${PAGE_ACTIONS_CLASS}`, () => {
+			const flexItems = findComponentsWithType(pageHead, 'Flex');
+			const pageActions = flexItems.filter(flex => flex.props.className && flex.props.className.indexOf(PAGE_ACTIONS_CLASS) !== -1);
```
mmcgahan: you can probably use [`scryRenderedDOMComponentsWithClass()`](https://facebook.github.io/react/docs/test-utils.html#scryrendereddomcomponentswithclass) to select all the `PAGE_ACTIONS_CLASS` components regardless of 'type' - as long as you have a `PAGE_ACTIONS_CLASS` component, it probably doesn't matter if it's a `<Flex>`

eilinora:
@nlaz from your comment about `PageActionsButton`, yes you can have `PageActionsButton`'s outside of `PageHead`

src/button.test.jsx
eilinora: Updated in latest push :)

nlaz:
So these components have a hard dependency on the `swarm-icons` library because we use the `Icon` component and pass in the icon name. We might want to make that optional. Perhaps we could pass in the Icon component instead of the iconShape string?

eilinora:
@nlaz updated breakpoint
screen shot 2017-02-06 at 1 37 02 pm

src/pageActionButton.test.jsx
```diff
@@ -0,0 +1,138 @@
...
+		expect(pageActionButtonEl).not.toBeNull();
+	});
+
+	it(`should render a 'button' as the root element with '${PAGE_ACTION_BUTTON_CLASS}'`, () => {
+		expect(pageActionButtonEl.nodeName).toBe('BUTTON');
+		expect(pageActionButtonEl.classList).toContain(PAGE_ACTION_BUTTON_CLASS);
+	});
+
+	it('should render a `Button` component with `onClick` prop', () => {
+		const button = findComponentsWithType(pageActionButton, 'Button');
```
mmcgahan: more info - @nlaz brought up the `findRenderedComponentsWithType` method with me last week and wondered why we've been using the `findComponentsWithType` function that I wrote ages ago instead. `findComponentsWithType` does a similar search, although it takes a string rather than a reference to a class, which is helpful when you don't have a convenient direct reference to the class you're searching for, but it otherwise a little bit roundabout, and I'm not sure it even works with pure functional components.

Going forward, `TestUtils.findRenderedComponentsWithType` will almost always be the preferred method to do this kind of check.

src/button.test.jsx
```diff
@@ -1,35 +1,35 @@
...
-	const BUTTON_CLASS = 'button';
-
 	describe('is a HTML button element', () => {
-		let buttonEl;
+		let button;
 
-		beforeEach(() => {
-			const button = TestUtils.renderIntoDocument(<Button />);
-			buttonEl = ReactDOM.findDOMNode(button);
+		beforeAll(() => {
```
mmcgahan: I very strongly recommend you leave this as `beforeEach` - there's no way to guarantee that _no_ future test will mutate (or even reassign) `button`, so `beforeAll` exposes the tests to a class of bugs that could lead to some significant unit test problems. Since the performance cost of `beforeEach` vs `beforeAll` is negligible, just use `beforeEach`

src/PageHead.jsx
```diff
@@ -0,0 +1,88 @@
...
+PageHead.propTypes = {
+	title: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element
+	]).isRequired,
+	subtitle: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element
+	]),
+	menuItems: React.PropTypes.array,
```
eilinora: for the moment i wanted to keep this flexible since we haven't actually implemented a use case yet. Wanna see how things shake out.

PR #66, opened 2017-01-31
delete tab isSelected prop before rendering

akdetrick:
`isSelected` is not a standard DOM property. This fix avoids rendering anchors like this:

```
<a href="..." isSelected>...
```


src/Tabs.jsx
```diff
@@ -12,6 +12,10 @@ class Tabs extends React.Component {
 				{'tabs-tab--selected': tab.props.isSelected}
 			);
 
+			// remove prop that is not standard DOM
+			// before rendering
+			delete tab.props.isSelected;
```
nlaz: I'm trying to think if there's another way to do this but this seems like the simplest way. LGTM after tests pass.

akdetrick:
Just remembered `props` is an immutable object in react. I have to figure out a different way to deal with rendering each `{tab}` without passing along the invalid prop.

mmcgahan:
I think the best option might be to go back to building a `<Tab>` component that wraps an arbitrary element and injects the 'selected' className into it without passing along the `isSelected` prop

```jsx
<Tabs>
  <Tab isSelected>
    <a href='http://example.com'>whatever</a>
  </Tab>
</Tabs>

// renders:

<div class='tabs'>
  <a class='tabs-tab tabs-tab--selected' href='http://example.com'>whatever</a>
</div>
```

akdetrick:
Yeah, unfortunately, I think a `<Tab>` wrapper is the best available option. This might be a sign that there was something fundamentally weird about passing a prop to one component intended for a wrapping component

akdetrick:
bump;

This is finally ready to go. I've added both modules to the same file (as I did with `GridList`) for a cleaner import: `import { Tabs, TabsTab } from 'Tabs';`

PR #73, opened 2017-02-09
Checkbox updates from responsify checkbox

unjust:
#### Related issues

Fixes https://meetup.atlassian.net/browse/MUP-7777

#### Description

Checkbox component for Event create. I pulled the initial files out of responsify and made some changes. Tests to come.

screen shot 2017-02-09 at 12 03 20 pm

----

screen shot 2017-02-09 at 2 54 42 pm


src/Checkbox.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+		const elId = id || `${name}-${value}`;
+
+		return (
+			<Flex align='center'
+				className={classNames}
+				{...other}>
+				<FlexItem shrink>
+					<input	type='checkbox'
+						name={name}
```
akdetrick: My information may be out of date, but a good practice used to be making `name` and `id` attributes have the same value

src/Checkbox.jsx
```diff
@@ -0,0 +1,70 @@
...
+					{children}
+				</FlexItem>
+			</Flex>
+		);
+	}
+}
+
+Checkbox.propTypes = {
+	checked: React.PropTypes.bool,
+	label: React.PropTypes.string,
```
eilinora: We would ideally make this a shape which can take an element along with string so it can handle translations. For reference you can see https://github.com/meetup/meetup-web-components/blob/master/src/PageHead.jsx#L76

mmcgahan: there is technically [a way to get a raw translated string](https://github.com/meetup/mup-web/blob/master/docs/i18n.md#get-the-translated-string-as-a-value-in-js), but I agree that it would be nice to just pass in a `<FormattedMessage>` element.

src/checkbox.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+		const labelEl = checkboxNode.querySelector('label');
+
+		expect(checkboxEl.getAttribute('id')).toEqual(`${name}-${value}`);
+		expect(labelEl.getAttribute('for')).toEqual(`${name}-${value}`);
+	});
+
+
+	it('should be checked when specified', function() {
+		checkbox = TestUtils.renderIntoDocument(<Checkbox name='greeting' id='hello' value='hello' checked />);
+		checkboxNode = ReactDOM.findDOMNode(checkbox);
```
eilinora: We have been going back and forth on a few tests using findDOMNode so not gonna mark this as a change request but... offering this as an alternative approach to writing these tests. You can also read more about this [http://reactkungfu.com/2015/07/approaches-to-testing-react-components-an-overview/](here) 

```
it("renders a list in a box with proper CSS classes", function() {
    let box = this.TestUtils.findRenderedDOMComponentWithTag(this.component, "div");
    expect(box.props.className).toEqual("invitation-list-container");

    let list = this.TestUtils.findRenderedDOMComponentWithTag(box, "ul");
    expect(list.props.className).toEqual("invitation-list");
  });
```

src/Checkbox.jsx
```diff
@@ -0,0 +1,70 @@
...
+					<input	type='checkbox'
+						name={name}
+						value={value}
+						checked={this.state.checked}
+						id={elId}
+						onChange={this.onChange}
+					/>
+				</FlexItem>
+				<FlexItem>
+					<label className='label--minor' htmlFor={elId}>{label}</label>
```
mperrotti: TL;DR: Checkbox labels shouldn't be bold

The `label--minor` class was something we were playing with to "un-bold" `<label>` for radio and checkbox inputs. We wanted to use the bold text for `<legend>` when we have a set of checkboxes that need to be labeled. 
E.g.: ``` <fieldset>
    <legend>Colors</legend>
    <input type="checkbox" id="red">
    <label for="red">Red</label>
    <input type="checkbox" id="blue">
    <label for="blue">Blue</label>
  </fieldset>```

The CSS behind `label--minor` didn't make it into this repo, it should probably go in [swarm-sasstools](https://github.com/meetup/swarm-sasstools)

unjust: ok. what class do we prefer?

akdetrick: Go ahead and use `label--minor` in this component; I'll make sure the class will make it into `swarm-sasstools` with this PR: https://github.com/meetup/meetup-web-components/pull/80

This has been filed in Jira as: https://meetup.atlassian.net/browse/SDS-157

Once `swarm-sasstools` is merged into this repo, the class will be defined and your labels will get prettier.

src/checkbox.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+	});
+
+	afterEach(() => {
+		checkboxComponent = null;
+		checkbox = null;
+	});
+
+	it('exists', function() {
+		expect(checkbox).not.toBeNull();
+	});
```
eilinora: because your using findRenderedDOMComponent this wont return null when it fails, instead it throws an error. i would do 
```
expect(() => TestUtils.findRenderedDOMComponentWithType(checkboxComponent, Checkbox)).not.toThrow() 
```
instead

PR #74, opened 2017-02-10
add test and component for section title

akdetrick:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-20

#### Description
`SectionTitle` component - headings for `Section` components, with optional children for actions on the right.

screen shot 2017-02-10 at 12 41 20 pm


src/SectionTitle.jsx
```diff
@@ -0,0 +1,48 @@
...
+						<FlexItem shrink>
+							{children}
+						</FlexItem>
+					</Chunk>
+				}
+			</Flex>
+		);
+	}
+}
+
```
eilinora: PropTypes 😸

src/sectionTitle.test.jsx
```diff
@@ -0,0 +1,28 @@
...
+		sectionTitle = TestUtils.renderIntoDocument(<SectionTitle />);
+		titleNode = ReactDOM.findDOMNode(sectionTitle);
+	});
+	afterEach(() => {
+		sectionTitle = null;
+		titleNode = null;
+	});
+
+	it('exists', function() {
+		expect(titleNode).not.toBeNull();
```
eilinora: Offering the same comment here as in another PR

We have been having discussions on using findDOMNode in some PR's so not gonna mark this as a change request but... offering this as an alternative approach to writing these tests. You can also read more about this [http://reactkungfu.com/2015/07/approaches-to-testing-react-components-an-overview/](here) 

```
it("exists", () => {
    expect(()=>TestUtils.findRenderedDOMComponentWithTag(sectionTitle, SectionTitle)).toThrow();
});
it("renders a list in a box with proper CSS classes", function() {
    let box = this.TestUtils.findRenderedDOMComponentWithTag(this.component, "div");
    expect(box.props.className).toEqual("invitation-list-container");

    let list = this.TestUtils.findRenderedDOMComponentWithTag(box, "ul");
    expect(list.props.className).toEqual("invitation-list");
  });
```

src/SectionTitle.jsx
akdetrick: I left these out, but realized we do in fact pass a `title` prop. And then I realized an explicit `action` prop would be nice instead of implicitly using `children` for the optional right side link

src/SectionTitle.jsx
```diff
@@ -0,0 +1,52 @@
...
+							{action}
+						</Chunk>
+					</FlexItem>
+				}
+			</Flex>
+		);
+	}
+}
+SectionTitle.propTypes = {
+	title: React.PropTypes.oneOf(React.PropTypes.element, React.PropTypes.string),
```
eilinora: should we make this required, since it's a title display component?

src/sectionTitle.story.jsx
```diff
@@ -0,0 +1,41 @@
+
```
eilinora: obnoxious nitpick, ☠️  space

src/SectionTitle.jsx
```diff
@@ -0,0 +1,52 @@
...
+			className
+		);
+
+		return (
+			<Flex
+				align='center'
+				className={classNames}
+				{...other}
+				>
+				<FlexItem>
```
eilinora: Since chunk is in both would this structure still give the look we need

```
<Chunk>
  <FlexItem>
    ....
  </FlexItem>
  {action && <FlexItem />
</Chunk>

File: src/sectionTitle.story.jsx [hunk 6]
```diff
@@ -0,0 +1,41 @@
...
+import React from 'react';
+import Bounds from './Bounds';
+import Section from './Section';
+import SectionTitle from './SectionTitle';
+import { storiesOf } from '@kadira/storybook';
+
+storiesOf('SectionTitle', module)
+	.add('Default', () => (
```
eilinora: let's make sure to add the auto documentation, https://github.com/meetup/meetup-web-components/blob/master/src/bounds.story.jsx#L7

src/sectionTitle.test.jsx
```diff
@@ -0,0 +1,58 @@
...
+	it('should not render a second flex item if there is no action', function() {
+		const flexItems = TestUtils.scryRenderedDOMComponentsWithClass(component, 'flex-item');
+
+		expect(flexItems.length).toBe(1);
+	});
+
+});
+
+describe('SectionTitle with action', function() {
+	const title = 'Test title';
```
eilinora: Maybe also a check to test that text and actions are in the component's render tree

src/SectionTitle.jsx
akdetrick: sure would!

src/SectionTitle.jsx
akdetrick: yup

src/SectionTitle.jsx
akdetrick: actually, no - that would override the display property of `Chunk` (the chunks would become flex children). Wrapping the whole thing could work, but I don't want to overload a `Chunk` with the section title class, which may also mess with the `display` style property

mmcgahan: I was talking with Mike P about this a couple weeks ago, asking about cases where we want chunk-like functionality on a non-chunk component, and he suggested that in those cases you might just want a `margin--bottom` className on the outer component, in this case the `<Flex>`. I think that expresses the layout intention reasonably clearly without having to worry about awkward/redundant layout composition. Just an option.

akdetrick: For this component, I don't feel strongly about moving away from `Chunk`. It was a typical pattern in Sassquatch2 to put chunks inside of row items (rather than around them). It makes sense to me, but if it gets weird for other people, I'm on board with `margin--bottom` classes. 

The nice thing is that `SectionTitle` abstracts it away either way.

src/sectionTitle.test.jsx
akdetrick: Added test

PR #79, opened 2017-02-13
Accordion component

mperrotti:
#### Related issues
For the upcoming work on event create

#### Description
A component to handle cases where we hide and show panels of related content. Each panel has a label that is always visible and summarizes the content that will be shown by clicking the label.

screen shot 2017-02-13 at 3 06 19 pm

### DIFF
--- a/assets/scss/_accordion.scss
+++ b/assets/scss/_accordion.scss
@@ -1,0 +1,37 @@
+//
+// Animation values taken from https://github.com/google/material-design-lite/blob/mdl-1.x/src/_variables.scss
+// Maybe we can bring this into `swarm-constants` or `swarm-animation`?
+//
+
+$A_curve-fast-out-slow-in: cubic-bezier(0.4, 0, 0.2, 1);
+$A_curve-default: $A_curve-fast-out-slow-in;
+$A_duration: 0.2s;
+
+.accordionPanel-animator {
+	overflow: hidden;
+	transition: height $A_duration $A_curve-default;
+}
+
+.accordionPanel-animator--collapse {
+	height: 0;
+	margin-bottom: 0 !important;
+}
+
+.accordionPanel-label {
+	background: inherit;
+	border: inherit;
+	text-align: inherit;
+	padding: 0;
+}
+
+.accordionPanel-icon .svg--chevron-down .svg-icon {
+	-webkit-transform: rotate(0deg);
+	transform: rotate(0deg);
+	transition: transform $A_duration $A_curve-default;
+
+	.accordionPanel--active & {
+		-webkit-transform: rotate(180deg);
+		transform: rotate(180deg);
+	}
+
+}
--- a/src/AccordionPanel.jsx
+++ b/src/AccordionPanel.jsx
@@ -1,0 +1,169 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import cx from 'classnames';
+
+import Chunk from './Chunk';
+import Flex from './Flex';
+import FlexItem from './FlexItem';
+import Icon from './Icon';
+
+/**
+ * @module AccordionPanel
+ */
+class AccordionPanel extends React.Component {
+	/**
+	 * Document this for PR
+	 */
+	constructor(props){
+		super(props);
+		this.state = {
+			open: this.props.isOpen
+		};
+
+		this._handleToggle = this._handleToggle.bind(this);
+		this._updateHeight = this._updateHeight.bind(this);
+	}
+
+	/**
+	 * Document this for PR
+	 */
+	_handleToggle(event){
+		this.setState({
+			open: !this.state.open
+		});
+	}
+
+	_updateHeight() {
+		const content = ReactDOM.findDOMNode(this.content);
+		const animator = ReactDOM.findDOMNode(this.animator);
+		const height = this.props.isOpen ? content.clientHeight : 0;
+		const cssHeight = `${height}px`;
+
+		if (animator.style.height === cssHeight) {
+			return;
+		}
+
+		animator.style.height = cssHeight;
+	}
+
+	componentDidMount() {
+		this._updateHeight();
+	}
+
+	componentDidUpdate() {
+		this._updateHeight();
+	}
+
+	componentWillReceiveProps(nextProps) {
+		if (nextProps.isOpen !== this.state.open) {
+			this.setState({ open: nextProps.isOpen });
+		}
+	}
+
+	render() {
+		const {
+			isOpen, // eslint-disable-line no-unused-vars
+			name,
+			onTriggerClick,
+			panelContent,
+			triggerIconShape,
+			triggerIconShapeActive,
+			triggerIconSize,
+			triggerIconAlign,
+			triggerLabel,
+			animated,
+			classNamesActive,
+			className,
+			...other
+		} = this.props;
+
+		const classNames = {
+			accordionPanel: cx(
+				'accordionPanel',
+				{
+					'accordionPanel--active': this.state.open,
+					[classNamesActive]: this.state.open && classNamesActive
+				},
+				className
+			),
+			content: cx(
+				{
+					'accordionPanel-animator': animated,
+					'accordionPanel-animator--collapse': !this.state.open,
+					'visibility--a11yHide': !animated && !this.state.open
+				}
+			)
+		};
+
+		console.log(onTriggerClick);
+
+		return(
+			<Flex
+				className={classNames.accordionPanel}
+				rowReverse={triggerIconAlign == 'left' ? 'atAll' : null}
+				{...other}>
+
+				<FlexItem>
+					<Chunk>
+						<button
+							role='tab'
+							id={`label-${name}`}
+							aria-controls={`panel-${name}`}
+							aria-expanded={this.state.open}
+							aria-selected={this.state.open}
+							className='accordionPanel-label display--block span--100'
+							onClick={onTriggerClick}>
+								{triggerLabel}
+						</button>
+					</Chunk>
+
+					<Chunk
+						role='tabpanel'
+						aria-labelledby={`label-${name}`}
+						aria-hidden={!this.state.open}
+						className={classNames.content}
+						ref={(el) => { this.animator = el; }}>
+						<div
+							className='accordionPanel-content'
+							ref={(div) => { this.content = div; }}>
+							{panelContent}
+						</div>
+					</Chunk>
+				</FlexItem>
+
+				{triggerIconShape &&
+					<FlexItem
+						className='accordionPanel-icon'
+						onClick={onTriggerClick}
+						shrink>
+							{this.state.open && triggerIconShapeActive ?
+								<Icon
+									shape={triggerIconShapeActive}
+									size={triggerIconSize} /> :
+								<Icon
+									shape={triggerIconShape}
+									size={triggerIconSize} />
+								}
+					</FlexItem>
+				}
+
+			</Flex>
+		);
+	}
+}
+
+AccordionPanel.defaultProps = {
+	triggerIconAlign: 'right',
+	triggerIconShape: 'chevron-down',
+	triggerIconSize: 'xs',
+	isOpen: false
+};
+
+AccordionPanel.propTypes = {
+	isOpen: React.PropTypes.bool,
+	triggerIconAlign: React.PropTypes.string,
+	triggerIconShape: React.PropTypes.string,
+	triggerIconSize: React.React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
+};
+
+export default AccordionPanel;
--- a/src/AccordionPanel.jsx
+++ b/src/AccordionPanel.jsx
@@ -95,8 +95,6 @@ class AccordionPanel extends React.Component {
 			)
 		};
 
-		console.log(onTriggerClick);
-
 		return(
 			<Flex
 				className={classNames.accordionPanel}
@@ -112,7 +110,7 @@ class AccordionPanel extends React.Component {
 							aria-expanded={this.state.open}
 							aria-selected={this.state.open}
 							className='accordionPanel-label display--block span--100'
-							onClick={onTriggerClick}>
+							onClick={onTriggerClick || this._handleToggle}>
 								{triggerLabel}
 						</button>
 					</Chunk>
@@ -134,7 +132,7 @@ class AccordionPanel extends React.Component {
 				{triggerIconShape &&
 					<FlexItem
 						className='accordionPanel-icon'
-						onClick={onTriggerClick}
+						onClick={onTriggerClick || this._handleToggle}
 						shrink>
 							{this.state.open && triggerIconShapeActive ?
 								<Icon
@@ -163,7 +161,7 @@ AccordionPanel.propTypes = {
 	isOpen: React.PropTypes.bool,
 	triggerIconAlign: React.PropTypes.string,
 	triggerIconShape: React.PropTypes.string,
-	triggerIconSize: React.React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
+	triggerIconSize: React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
 };
 
 export default AccordionPanel;
--- a/src/AccordionPanelGroup.jsx
+++ b/src/AccordionPanelGroup.jsx
@@ -84,6 +84,7 @@ class AccordionPanelGroup extends React.Component {
 			accordionPanels, // eslint-disable-line no-unused-vars
 			triggerIconAlign, // eslint-disable-line no-unused-vars
 			triggerIconShape, // eslint-disable-line no-unused-vars
+			triggerIconShapeActive, // eslint-disable-line no-unused-vars
 			triggerIconSize, // eslint-disable-line no-unused-vars
 			multiSelectable,
 			className,
@@ -120,6 +121,7 @@ AccordionPanelGroup.propTypes = {
 	accordionPanels: React.PropTypes.arrayOf(React.PropTypes.element).isRequired,
 	multiSelectable: React.PropTypes.bool,
 	triggerIconAlign: React.PropTypes.string,
+	triggerIconShapeActive: React.PropTypes.string,
 	triggerIconShape: React.PropTypes.string,
 	triggerIconSize: React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
 };
--- a/src/AccordionPanelGroup.jsx
+++ b/src/AccordionPanelGroup.jsx
@@ -121,7 +121,7 @@ AccordionPanelGroup.propTypes = {
 	multiSelectable: React.PropTypes.bool,
 	triggerIconAlign: React.PropTypes.string,
 	triggerIconShape: React.PropTypes.string,
-	triggerIconSize: React.React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
+	triggerIconSize: React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
 };
 
 export default AccordionPanelGroup;
--- a/src/AccordionPanelGroup.jsx
+++ b/src/AccordionPanelGroup.jsx
@@ -1,0 +1,127 @@
+import _ from 'lodash';
+import React from 'react';
+import cx from 'classnames';
+
+/**
+ * @module AccordionPanelGroup
+ */
+class AccordionPanelGroup extends React.Component {
+	constructor(props) {
+		super(props);
+		const activePanelsArr = [];
+
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
+			if(accordionPanel.props.isOpen) {
+				activePanelsArr.push(i);
+			}
+			return activePanelsArr;
+		});
+
+		this.state = {
+			activePanels: activePanelsArr || []
+		};
+		this._onClickTrigger = this._onClickTrigger.bind(this);
+	}
+
+	/**
+	 * Document this for PR
+	 */
+	_onClickTrigger(e, i){
+		const index = this.state.activePanels.indexOf(i);
+		const activePanelArr = this.state.activePanels;
+		const filtered = activePanelArr.filter(panel => i !== activePanelArr.indexOf(i));
+
+		if (index > -1){
+			activePanelArr.splice(index, 1);
+		} else {
+			activePanelArr.push(i);
+		}
+
+		if (!this.props.multiSelectable) {
+			_.pullAll(activePanelArr, filtered);
+		}
+
+		this.setState({
+			activePanels: activePanelArr
+		});
+
+	}
+
+	renderAccordionPanels() {
+		const toCamelCase = (str) => str.slice(0,20).replace(/(?:^\w|[A-Z]|\b\w)/g, (ltr, idx) => idx === 0 ? ltr.toLowerCase() : ltr.toUpperCase()).replace(/\s+/g, '');
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
+
+			return (
+				<li
+					className='list-item'
+					key={i}
+				>
+					{
+						React.cloneElement(accordionPanel,
+							{
+								triggerIconAlign: this.props.triggerIconAlign,
+								triggerIconShape: this.props.triggerIconShape,
+								triggerIconShapeActive: this.props.triggerIconShapeActive,
+								triggerIconSize: this.props.triggerIconSize,
+								animated: this.props.animated,
+								name: toCamelCase(accordionPanel.props.triggerLabel),
+								className: accordionPanel.props.className,
+								isOpen: this.state.activePanels.includes(i),
+								onTriggerClick: (e) => this._onClickTrigger(e, i),
+							}
+						)
+					}
+				</li>
+			);
+		});
+
+		return this.accordionPanels;
+	}
+
+	render() {
+		const {
+			animated, // eslint-disable-line no-unused-vars
+			accordionPanels, // eslint-disable-line no-unused-vars
+			triggerIconAlign, // eslint-disable-line no-unused-vars
+			triggerIconShape, // eslint-disable-line no-unused-vars
+			triggerIconSize, // eslint-disable-line no-unused-vars
+			multiSelectable,
+			className,
+			...other
+		} = this.props;
+
+		const classNames = cx(
+			'accordionPanelGroup',
+			'list',
+			className
+		);
+
+		return (
+			<ul
+				role='tabList'
+				aria-multiselectable={multiSelectable}
+				className={classNames}
+				{...other}>
+					{this.renderAccordionPanels()}
+			</ul>
+		);
+	}
+}
+
+AccordionPanelGroup.defaultProps = {
+	multiSelectable: true,
+	triggerIconAlign: 'right',
+	triggerIconShape: 'chevron-down',
+	triggerIconSize: 'xs',
+};
+
+
+AccordionPanelGroup.propTypes = {
+	accordionPanels: React.PropTypes.arrayOf(React.PropTypes.element).isRequired,
+	multiSelectable: React.PropTypes.bool,
+	triggerIconAlign: React.PropTypes.string,
+	triggerIconShape: React.PropTypes.string,
+	triggerIconSize: React.React.PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
+};
+
+export default AccordionPanelGroup;
--- a/src/accordionPanel.story.jsx
+++ b/src/accordionPanel.story.jsx
@@ -1,0 +1,26 @@
+
+import React from 'react';
+import AccordionPanel from './AccordionPanel';
+import { storiesOf } from '@kadira/storybook';
+
+/*
+ * -- Inline SVG icon sprite --
+ *
+ * raw SVG sprite from `swarm-icons`
+ */
+const iconSpriteStyle = { display: 'none' };
+const iconSprite = require('raw-loader!swarm-icons/dist/sprite/sprite.inc');
+
+storiesOf('AccordionPanel', module)
+	.add('default', () => (
+		<div className='span--100 padding--all'>
+			<AccordionPanel
+				triggerLabel='First Section'
+				panelContent={
+					<div className='runningText'>
+						<p>Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard McClintock, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the undoubtable source.</p>
+					</div>
+				} />
+			<div style={iconSpriteStyle} dangerouslySetInnerHTML={{__html: iconSprite}} />
+		</div>
+	));
--- a/src/accordionPanel.test.jsx
+++ b/src/accordionPanel.test.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+import AccordionPanel from './AccordionPanel';
+
+describe('AccordionPanel', function() {
+
+	it('exists', function() {
+		const accordion = TestUtils.renderIntoDocument(<AccordionPanel />);
+		const accordionNode = ReactDOM.findDOMNode(accordion);
+
+		expect(accordionNode).not.toBeNull();
+	});
+
+});
--- a/src/accordionPanelGroup.test.jsx
+++ b/src/accordionPanelGroup.test.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+import AccordionPanelGroupContainer from './AccordionPanelGroupContainer';
+
+describe('AccordionPanelGroupContainer', function() {
+
+	it('exists', function() {
+		const accordionPanelGroup = TestUtils.renderIntoDocument(<AccordionPanelGroupContainer />);
+		const accordionPanelGroupNode = ReactDOM.findDOMNode(accordionPanelGroup);
+
+		expect(accordionPanelGroupNode).not.toBeNull();
+	});
+
+});
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -14,4 +14,5 @@
 //
 // COMPONENT SPECIFIC
 //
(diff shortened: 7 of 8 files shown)


src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+	renderAccordionPanels() {
+		const toCamelCase = (str) => str.slice(0,20).replace(/(?:^\w|[A-Z]|\b\w)/g, (ltr, idx) => idx === 0 ? ltr.toLowerCase() : ltr.toUpperCase()).replace(/\s+/g, '');
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
+
+			return (
+				<li
+					className='list-item'
+					key={i}
+				>
```
unjust: nitpick - should this just be on one line? also can line 51 be broken up on a few lines since its so long?

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+								triggerIconAlign: this.props.triggerIconAlign,
+								triggerIconShape: this.props.triggerIconShape,
+								triggerIconShapeActive: this.props.triggerIconShapeActive,
+								triggerIconSize: this.props.triggerIconSize,
+								animated: this.props.animated,
+								name: toCamelCase(accordionPanel.props.triggerLabel),
+								className: accordionPanel.props.className,
+								isOpen: this.state.activePanels.includes(i),
+								onTriggerClick: (e) => this._onClickTrigger(e, i),
+							}
```
unjust: should `i` become a prop of the Accordion element, like a data attribute, in order to identify which panel is being open and closed in `_onClickTrigger`? Maybe using the index from iterating in map is fine, but this way you could just have `onTriggerClick: this._onClickTrigger`

src/accordionPanelGroup.test.jsx
```diff
@@ -0,0 +1,15 @@
...
+describe('AccordionPanelGroupContainer', function() {
+
+	it('exists', function() {
+		const accordionPanelGroup = TestUtils.renderIntoDocument(<AccordionPanelGroupContainer />);
+		const accordionPanelGroupNode = ReactDOM.findDOMNode(accordionPanelGroup);
+
+		expect(accordionPanelGroupNode).not.toBeNull();
+	});
+
+});
```
unjust: a couple tests to add would be to simulate clicks on the accordion - make sure the handler gets called and that its in the expected open or closed state. Also check that the accordion panel group state gets updated.

assets/scss/_accordion.scss
```diff
@@ -0,0 +1,37 @@
...
+
+$A_curve-fast-out-slow-in: cubic-bezier(0.4, 0, 0.2, 1);
+$A_curve-default: $A_curve-fast-out-slow-in;
+$A_duration: 0.2s;
+
+.accordionPanel-animator {
+	overflow: hidden;
+	transition: height $A_duration $A_curve-default;
+}
+
```
unjust: these can be added to swarm-animation (which has its own animation constants). I don't want to get in the way of this getting merged. we need to think of a way to get this project to pull in animation.scss from swarm-animate pretty soon.

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+								{triggerLabel}
+						</button>
+					</Chunk>
+
+					<Chunk
+						role='tabpanel'
+						aria-labelledby={`label-${name}`}
+						aria-hidden={!this.state.open}
+						className={classNames.content}
+						ref={(el) => { this.animator = el; }}>
```
unjust: what is `this.animator`? reading about refs in React now...

src/accordionPanel.story.jsx
```diff
@@ -0,0 +1,26 @@
+
```
akdetrick: I'd recommend creating one story for all accordion features, `accordion.story.jsx`.

You can add a story for individual panel usage, and another for usage in a group.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+	constructor(props) {
+		super(props);
+		const activePanelsArr = [];
+
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
+			if(accordionPanel.props.isOpen) {
+				activePanelsArr.push(i);
+			}
+			return activePanelsArr;
+		});
```
mmcgahan: I don't think this `map` call is doing what you expect it to. `this.accordionPanels` will end up being an array that is the length of `this.props.accordionPanels`, where each element is a reference to the same `activePanelsArr` array. You actually never _read_ from `this.accordionPanels` (it gets re-assigned at the top of `renderAccordionPanels`), so I think you are intending to just `reduce` `this.props.accordionPanels` into an `activePanelsArr`

```js
const activePanelsArr = this.props.accodionPanels.reduce((panels, panel, i) => {
  if (panel.props.isOpen) {
    panels.push(i);
  }
  return panels;
}, []);
```

That said, it looks like you're actually storing an array of _indexes_ into the active panels, not the panels themselves, so the `activePanelsArr` naming is confusing. I think the show/hide logic is getting a little convoluted, though, so I'll leave a separate comment about options for cleaning it up.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+		this.state = {
+			activePanels: activePanelsArr || []
```
mmcgahan: `activePanelsArr` is always truthy here, even when it's empty.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+			activePanelArr.push(i);
+		}
+
+		if (!this.props.multiSelectable) {
+			_.pullAll(activePanelArr, filtered);
+		}
+
+		this.setState({
+			activePanels: activePanelArr
+		});
```
mmcgahan: this function is very difficult to follow - there's a lot of index matching and searching and filtering, but it's not clear what the end result is. I'm pretty sure that this whole component can be significantly simplified by breaking down the problem, but I'm finding it difficult to know where to start.

Maybe we can pair on this sometime soon so you can walk me through it

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+	 * Document this for PR
+	 */
+	_handleToggle(event){
+		this.setState({
+			open: !this.state.open
+		});
+	}
+
+	_updateHeight() {
+		const content = ReactDOM.findDOMNode(this.content);
```
eilinora: Using ReactDOM.findDOMNode() is considered a bad in the context of a component is considered a bad practice. Since `this.content` is already a reference to the DOM element you should be able to access the elements height. You want to go with https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect instead.

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+		const animator = ReactDOM.findDOMNode(this.animator);
+		const height = this.props.isOpen ? content.clientHeight : 0;
+		const cssHeight = `${height}px`;
+
+		if (animator.style.height === cssHeight) {
```
eilinora: What is this checking? :)

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+			this.setState({ open: nextProps.isOpen });
+		}
+	}
+
+	render() {
+		const {
+			isOpen, // eslint-disable-line no-unused-vars
+			name,
+			onTriggerClick,
+			panelContent,
```
eilinora: It looks like not all props are listed in PropTypes, can we add them in?

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+						ref={(el) => { this.animator = el; }}>
+						<div
+							className='accordionPanel-content'
+							ref={(div) => { this.content = div; }}>
+							{panelContent}
+						</div>
+					</Chunk>
+				</FlexItem>
+
+				{triggerIconShape &&
```
eilinora: As an alternative to passing in all the props for declaring the icon, instead let's pass in the <Icon/> element with the proper declarations set instead. This allows for Icon to be less coupled to this component as well

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+		this.state = {
+			open: this.props.isOpen
+		};
+
+		this._handleToggle = this._handleToggle.bind(this);
+		this._updateHeight = this._updateHeight.bind(this);
+	}
+
+	/**
+	 * Document this for PR
```
eilinora: we should probably document this as it's going into master

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+		this.state = {
+			activePanels: activePanelsArr || []
+		};
+		this._onClickTrigger = this._onClickTrigger.bind(this);
+	}
+
+	/**
+	 * Document this for PR
```
eilinora: we should probably document this as it's going into master

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+		const filtered = activePanelArr.filter(panel => i !== activePanelArr.indexOf(i));
+
+		if (index > -1){
+			activePanelArr.splice(index, 1);
+		} else {
+			activePanelArr.push(i);
+		}
+
+		if (!this.props.multiSelectable) {
+			_.pullAll(activePanelArr, filtered);
```
mmcgahan: I am strongly against importing `lodash` as a dependency - it is a very bloated library and 90% redundant with native JS array methods. In this case, I'm not exactly sure why you're pulling out the `filtered` indexes from `activePanelArr`, but I think this is the same as the more-explicit `.filter` alternative

```js
const actuallyActivePanelArr = activePanelArr.filter(panelIndex =>
  filtered.indexOf(panelIndex) === -1
);
```

Although there's some confusing mapping from `activePanelArr` to `filtered` and back into `activePanelArr`, so I'm not sure what the goal state is.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+			'list',
+			className
+		);
+
+		return (
+			<ul
+				role='tabList'
+				aria-multiselectable={multiSelectable}
+				className={classNames}
+				{...other}>
```
eilinora: nitpick but easier to read when `>` is on the next line when we do this kind of code style of indented props

src/accordionPanel.test.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+import AccordionPanel from './AccordionPanel';
+
+describe('AccordionPanel', function() {
+
+	it('exists', function() {
```
eilinora: Tests are very light for this :)

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		this.setState({
+			activePanels: activePanelArr
+		});
+
+	}
+
+	renderAccordionPanels() {
+		const toCamelCase = (str) => str.slice(0,20).replace(/(?:^\w|[A-Z]|\b\w)/g, (ltr, idx) => idx === 0 ? ltr.toLowerCase() : ltr.toUpperCase()).replace(/\s+/g, '');
```
mmcgahan: We have a [`toCamelCase` string helper](https://github.com/meetup/meetup-web-platform/blob/master/src/util/stringUtils.js#L16) in the platform repo

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
```
mmcgahan: looks like you can set a `const` rather than a property on`this` - `this.accordionPanels` is not consumed anywhere else.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,129 @@
...
+							{
+								triggerIconAlign: this.props.triggerIconAlign,
+								triggerIconShape: this.props.triggerIconShape,
+								triggerIconShapeActive: this.props.triggerIconShapeActive,
+								triggerIconSize: this.props.triggerIconSize,
+								animated: this.props.animated,
+								name: toCamelCase(accordionPanel.props.triggerLabel),
+								className: accordionPanel.props.className,
+								isOpen: this.state.activePanels.includes(i),
+								onTriggerClick: (e) => this._onClickTrigger(e, i),
```
mmcgahan: keep naming consistent `onClickTrigger` -> `_onClickTrigger`

src/accordionPanelGroup.test.jsx
mmcgahan: You check the coverage report when you run `yarn test` to make sure that all the new components have 100% test coverage.

assets/scss/_accordion.scss
```diff
@@ -0,0 +1,37 @@
...
+	padding: 0;
+}
+
+.accordionPanel-icon .svg--chevron-down .svg-icon {
+	-webkit-transform: rotate(0deg);
+	transform: rotate(0deg);
+	transition: transform $A_duration $A_curve-default;
+
+	.accordionPanel--active & {
+		-webkit-transform: rotate(180deg);
```
unjust: bourbon has one, but it is being deprecated.

assets/scss/_accordion.scss
eilinora: @unjust not 100% sure what swarm-animation is but you can checkout how we include WMC & sassquatch in mup-web. might be helpful https://github.com/meetup/mup-web/blob/master/src/assets/scss/main.scss#L11

src/AccordionPanelGroup.jsx
mperrotti: I think it's easier to read on 2 lines, but I could see moving the last angle bracket up to line 57 :)
I'd prefer just keeping line 51 to 1 line. All it's doing is making a string to use for aria.

src/AccordionPanelGroup.jsx
mperrotti: That could work - I'll try it out

assets/scss/_accordion.scss
mperrotti: @unjust - 100% agreed that this stuff should live in `swarm-animation`, but I didn't think it was ready to be used yet.
If we're ready to use `swarm-animation` - let me know and we'll do a follow-up to get that into `meetup-web-components`

src/AccordionPanel.jsx
mperrotti: I could go either way. The word "container" sounds a little clearer

src/accordionPanel.story.jsx
mperrotti: That makes sense to me

src/AccordionPanel.jsx
mperrotti: Yes, forgot to go back and update this before submitting the PR

src/AccordionPanel.jsx
mperrotti: Good catch. I'll go back and change this :)

src/AccordionPanel.jsx
mperrotti: Forgot to delete this from a different way I was trying to handle this. If you want more details, follow-up on Slack :)

src/AccordionPanel.jsx
mperrotti: Yes

src/AccordionPanel.jsx
mperrotti: Good idea. Will do

src/AccordionPanelGroup.jsx
mperrotti: Yes

src/AccordionPanelGroup.jsx
mperrotti: I've seen it done both ways. If having it on a separate line is our preferred style, I'll move it down

src/accordionPanel.test.jsx
mperrotti: Yes, going to write some tests based on Ivy's feedback

src/AccordionPanel.jsx
mperrotti: You're right - Lori had similar feedback.

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+	_handleToggle(event){
+		this.setState({
+			open: !this.state.open
+		});
+	}
+
+	_updateHeight() {
+		const content = ReactDOM.findDOMNode(this.content);
+		const animator = ReactDOM.findDOMNode(this.animator);
+		const height = this.props.isOpen ? content.clientHeight : 0;
```
mperrotti: Clever. I like it and I'm changing it :)

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+		const cssHeight = `${height}px`;
+
+		if (animator.style.height === cssHeight) {
+			return;
+		}
+
+		animator.style.height = cssHeight;
```
mperrotti: Excellent point. Will fix

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+		const {
+			isOpen, // eslint-disable-line no-unused-vars
+			name,
+			onTriggerClick,
+			panelContent,
+			triggerIconShape,
+			triggerIconShapeActive,
+			triggerIconSize,
+			triggerIconAlign,
+			triggerLabel,
```
mperrotti: @mmcgahan: that's how I originally had this, but then I can't set default props for the key values in `icon`. For example: if I want to use a default `size`, but change `shape`, values for everything but `shape` get thrown out.

If you know a way around this, let me know and I'll gladly change it

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+					'accordionPanel-animator--collapse': !this.state.open,
+					'visibility--a11yHide': !animated && !this.state.open
+				}
+			)
+		};
+
+		return(
+			<Flex
+				className={classNames.accordionPanel}
+				rowReverse={triggerIconAlign == 'left' ? 'atAll' : null}
```
mperrotti: Yup 👍

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+						<button
+							role='tab'
+							id={`label-${name}`}
+							aria-controls={`panel-${name}`}
+							aria-expanded={this.state.open}
+							aria-selected={this.state.open}
+							className='accordionPanel-label display--block span--100'
+							onClick={onTriggerClick || this._handleToggle}>
+								{triggerLabel}
+						</button>
```
mperrotti: Good point—should be pretty straightforward.

assets/scss/_accordion.scss
mperrotti: Something for @akdetrick to consider for [swarm-sasstools](https://github.com/meetup/swarm-sasstools)

src/AccordionPanelGroup.jsx
mperrotti: Never used/heard of `reduce`, thanks!

src/AccordionPanelGroup.jsx
mperrotti: Thanks, I'll use that instead

src/AccordionPanelGroup.jsx
mperrotti: Filter makes more sense, but you're right - I'm doing this in kind of a weird way.

src/AccordionPanelGroup.jsx
mperrotti: I'd love to pair on this @mmcgahan - I'll look for a slot on your calendar

src/AccordionPanelGroup.jsx
mperrotti: I see. Will do

src/AccordionPanelGroup.jsx
mperrotti: 👍

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,167 @@
...
+			isOpen, // eslint-disable-line no-unused-vars
+			name,
+			onTriggerClick,
+			panelContent,
+			triggerIconShape,
+			triggerIconShapeActive,
+			triggerIconSize,
+			triggerIconAlign,
+			triggerLabel,
+			animated,
```
mperrotti: Yes

mperrotti: Agreed

src/AccordionPanel.jsx
mmcgahan: I'd recommend "wrapper" over "container" because "container" has a different meaning in Redux-land (Container == Smart component connected to app state)

src/AccordionPanel.jsx
mmcgahan: you can set defaults in the `constructor`

```js
const triggerDefaults = {
  a: 'b',
  c: 'd',
};
// ...
constructor(props) {
  props.trigger = { ...props.trigger, ...triggerDefaults };
  // ...
}

File: assets/scss/_accordion.scss (hunk 22 above)
Comment (akdetrick) [inline assets/scss/_accordion.scss]: We will not be updating to newer versions of Bourbon for this reason. Once `swarm-sasstools` is in all of our new platform repos, I'm considering moving away from Bourbon to something like [postcss-loader](https://github.com/postcss/postcss-loader) to handle prefixing.

**Comment by mmcgahan** [comment]
bump. what's the status on this? is there a deadline?

**Comment by mperrotti** [comment]
@mmcgahan - no hard deadline, but Ivy will be using this component in Event Create.
I've addressed all feedback, I just have to write tests. This is my first time writing tests, so Ivy agreed to pair with me tomorrow afternoon.

**Comment by akdetrick** [comment]
Going with individual props for icon options so defaults can be clearly defined (and documented):
screen shot 2017-03-21 at 10 36 29 pm

File: src/AccordionPanel.jsx [hunk 29]
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * @module AccordionPanel
+ */
+class AccordionPanel extends React.Component {
+	constructor(props){
+		super(props);
+
+		this.state = {
+			open: this.props.isOpen
```
mmcgahan: keep naming consistent - `isOpen: this.props.isOpen`

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+	_handleToggle(){
+		const newState = !this.state.open;
+		const { setClickedPanel } = this.props;
+
+		if (setClickedPanel) {
+			setClickedPanel(this);
+		}
+
+		this.setState({
+			height: `${newState * this.content.getBoundingClientRect().height}px`,
```
mmcgahan: Since this height calculation is made 3 times, I'd recommend making it a function that takes `isOpen` as an argument

```js
height: getHeight(isOpen),
```

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+		this._handleToggle = this._handleToggle.bind(this);
+	}
+
+	/**
+	 * @returns {undefined}
+	 *
+	 * Updates state to toggle `AccordionPanel` open and closed
+	 */
+	_handleToggle(){
+		const newState = !this.state.open;
```
mmcgahan: keep naming consistent

```js
const  isOpen = !this.state.isOpen;
```

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+		// create valid attribute name from trigger label
+		const ariaId = label.replace(/\s+/g, '').toLowerCase();
+
+		return(
+			<Flex
+				className={classNames.accordionPanel}
+				rowReverse={iconAlign === 'left' && 'all'}
+				{...other}
+				>
```
mmcgahan: align closing `>` with opening `<Flex`

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+				<FlexItem>
+					<Chunk>
+						<button
+							role='tab'
+							id={`label-${ariaId}`}
+							aria-controls={`panel-${ariaId}`}
+							aria-expanded={this.state.open}
+							aria-selected={this.state.open}
+							className={classNames.trigger}
+							onClick={this._handleToggle}>
```
mmcgahan: align closing `>` with opening `<button` - that will make it clear you have an extra indent on `{label}`

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+								{label}
+						</button>
+					</Chunk>
+
+					<Chunk
+						role='tabpanel'
+						aria-labelledby={`label-${ariaId}`}
+						aria-hidden={!this.state.open}
+						className={classNames.content}
+						style={{height: this.state.height}}>
```
mmcgahan: align closing `>` with opening `<Chunk`

mmcgahan: cool story: if you pass Number values to `style` properties, it will automatically append `px`, which means you don't have to have those ugly template strings in your `this.state.height` assignments - just use the actual number

mmcgahan: I just realized that it looks like `this.state.height` is always derived directly from the current value of `this.state.open`, so why store the value of `height` in state at all? Could you just calculate it here each time you render?

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+						<div
+							className='accordionPanel-content'
+							ref={(div) => { this.content = div; }}>
```
mmcgahan: nitpick but `content` is _extremely vague_ for something that is referenced in separate functions - I recommend `contentEl` or something that indicates what kind of thing it is.

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+							{panelContent}
+						</div>
+					</Chunk>
+				</FlexItem>
+
+				<FlexItem
+					className='accordionPanel-icon'
+					onClick={this._handleToggle}
+					shrink
+					>
```
mmcgahan: align closing `>` with opening `<FlexItem`

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,198 @@
...
+					<Icon
+						shape={this.getIconShape()}
+						size={iconSize} />
```
mmcgahan: align closing `>` with opening `<Icon`, or just make the whole opening tag inline - it's short

```jsx
<Icon shape={this.getIconShape()} size={iconSize} />
```

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+export const ACCORDIONPANELGROUP_CLASS = 'accordionPanelGroup';
+/**
+ * @module AccordionPanelGroup
+ */
+class AccordionPanelGroup extends React.Component {
+	constructor(props) {
+		super(props);
+		this.accordionPanels = this.props.accordionPanels.map((accordionPanel, i) => {
+			return this.clonePanels(accordionPanel, i, accordionPanel.props.isOpen);
```
mmcgahan: since `clonePanels` takes a single panel as an argument, could you name it `clonePanel`?

Also, since the third argument is always a property of the first argument, you don't need to pass it in separately, which means that you can simplify this to

```js
this.props.accordionPanels.map(this.clonePanel);
```

However, you should `bind` `this.closePanel` _before_ calling it in this way.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	/**
+	 * @param clickedPanel is the `AccordionPanel` component passed in from `_handleToggle`
+	 * @returns {undefined}
+	 *
+	 * Keeps track of the clicked panel in order to support AccordionPanelGroups that only
+	 * have one panel open at a time.
+	 */
+	setClickedPanel(clickedPanel) {
```
mmcgahan: Thinking about this a bit more, and I'm wondering if you should just _always_ use `setState({ clickedPanel })`, but ignore the value when `multiSelectable` is `false` rather than trying to be smart about setting `this.setClickedPanel` to be false and then checking whether it's false before calling it... essentially make the method calling consistent, by putting all the logic into figuring out what to _do_ with the information it provides

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+			isOpen: isOpen
+		};
+		return React.cloneElement(accordionPanel, panelProps);
+	}
+
+	/**
+	 * @returns {Array} `AccordionPanel` components with the correct value for `isOpen` prop
+	 */
+	renderAccordionPanels() {
+		this.accordionPanels = this.accordionPanels.map((accordionPanel, i) => {
```
mmcgahan: pretty sure this could also be simplified to `this.accordionPanels = this.accordionPanels.map(this.clonePanels)`

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+			className
+		);
+
+		return (
+			<ul
+				role='tabList'
+				aria-multiselectable={multiSelectable}
+				className={classNames}
+				{...other}
+				>
```
mmcgahan: align closing `>` with opening `<ul`

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+					this.renderAccordionPanels().map(
+						(panel, i) => (
+						<li
+							key={i}
+							className='list-item'>
+							{panel}
+						</li>
+						)
+					)
+				}
```
mmcgahan: indentation is a little weird here - simplify

```jsx
this.renderAccordionPanels().map((panel, key) => (
  <li key={key} className='list-item'>{panel}</li>
)
```

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * @returns {Array} `AccordionPanel` components with the correct value for `isOpen` prop
+	 */
+	renderAccordionPanels() {
+		this.accordionPanels = this.accordionPanels.map((accordionPanel, i) => {
+			const isOpen = accordionPanel.props.isOpen;
+
+			/*
+			 *if ( !this.props.multiSelectable && this.state.clickedPanel === accordionPanel ) {
```
nlaz: Is this needed anymore?

src/AccordionPanelGroup.jsx
akdetrick: yeah, changing this to `clonePanel` makes sense. 

The `isOpen` argument is either provided by the props of the `AccordionPanel`, or provided by some logic in `AccordionPanelGroup`. After giving some refactoring a shot, I feel like keeping the argument makes it easier to understand when we're relying on the panel props vs. explicitly setting `isOpen`.

src/AccordionPanel.jsx
akdetrick: seemed like a good idea, but got a helpful warning:
```
Warning: a `div` tag (owner: `Chunk`) was passed a numeric string value for CSS property `height` (value: `184`) which will be treated as a unitless number in a future version of React.
```

mmcgahan: ohhh. okay I will update my knowledge.

src/accordion.test.jsx
```diff
@@ -0,0 +1,185 @@
...
+					</div>
+				} />
+		);
+	});
+	afterEach(() => {
+		panel = null;
+		openPanel = null;
+	});
+
+	it(`has the class ${ACTIVEPANEL_CLASS} when state is open`, function() {
```
akdetrick: Given that the panel content visibility is based on `height` manipulation rather than CSS class toggling, we'll have some trouble checking this in tests.

Our thinking was that covering state changes and the `ACTIVEPANEL_CLASS` would be adequate for alerting us when content visibility might break

akdetrick:
Bump; how are we feeling about this one in its current state?

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,203 @@
...
+		return `${isOpen * this.contentEl.getBoundingClientRect().height}px`;
+	}
+
+	/**
+	 * @returns {undefined}
+	 *
+	 * Updates state to toggle `AccordionPanel` open and closed
+	 */
+	_handleToggle(){
+		const newState = !this.state.isOpen;
```
mmcgahan: can this be `isOpen` to avoid a new variable alias?

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,203 @@
...
+		const { setClickedPanel } = this.props;
```
mmcgahan: nitpick, but since this var is only used once in an `if` test, I'd skip the destructuring and just write `if (this.props.setClickedPanel)`

src/AccordionPanel.jsx
```diff
@@ -0,0 +1,203 @@
...
+			height: this.getHeight(newState),
+			isOpen: newState
+		});
+
+	}
+
+	/**
+	 * @returns {undefined}
+	 *
+	 * Sets height of `AccordionPanel` to be appear open or closed when mounting
```
mmcgahan: style nitpick - JSDoc function description usually goes _above_ the `@` tags

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * @param {Object} accordionPanel - `AccordionPanel` components to clone
+	 * @param {number} i - index of the `AccordionPanel`
+	 * @param {boolean} isOpen - whether the `AccordionPanel` is open or not
+	 * @returns {Array} `AccordionPanel` components with props from `AccordionPanelGroup`
+	 */
+	clonePanel(accordionPanel, i, isOpen) {
+		const panelProps = {
+			key: i,
```
mmcgahan: trivial, but feel free to name the index argument `key` so that you don't have to alias it.

src/AccordionPanelGroup.jsx
```diff
@@ -0,0 +1,115 @@
...
+			<ul
+				role='tabList'
+				aria-multiselectable={multiSelectable}
+				className={classNames}
+				{...other}
+			>
+				{
+					this.renderAccordionPanels().map((panel, i) => (
+							<li key={i} className='list-item'> {panel} </li>
+						)
```
mmcgahan: lines 90 and 91 have an extra indent - the close paren on 91 should go on the same line as the one on 92 since their corresponding open parens are both on line 89

akdetrick:
This should be good, as long as we feel ok about https://github.com/meetup/meetup-web-components/pull/79#discussion-diff-107768714R148

PR #80, opened 2017-02-13
Sds 152/use swarm sassutils

akdetrick:
#### Related issues

Fixes https://meetup.atlassian.net/browse/SDS-152

#### Description
This allows us to begin using `swarm-sasstools` in order to move away from framework-provided component level CSS.

1. Include `swarm-sasstools` for CSS classes and Sass support instead of `sassquatch2`
2. Remove shim utilities like `responsiveBase` in favor of documented, updated versions in `swarm-sasstools`
3. Organize `scss` assets to make it easy to include component css only in other repos
4. Move component-specific Sass from `sassquatch2` into this repo to cover existing components


assets/scss/components/_avatar.scss
```diff
@@ -0,0 +1,170 @@
...
+		position: absolute;
+		right: 0;
+		text-align: center;
+		text-indent: 0%;
+		width: $space;
+	}
+}
+
+%avatar,
+.avatar {
```
nlaz: Why are both `%avatar` and `.avatar` needed here? [Looking at this guide](http://thesassway.com/intermediate/understanding-placeholder-selectors) makes me believe that only `.avatar` would be needed.

akdetrick: This is probably a vestige of sassquatch2; there will be a lot of those. As long as there aren't any harmful leftovers, we can clean them up in a later iteration

Makefile
```diff
@@ -1,5 +1,6 @@
 CI_BUILD_NUMBER ?= $(USER)-snapshot
 VERSION ?= 0.1.$(CI_BUILD_NUMBER)
+VERSION ?= 0.2.$(CI_BUILD_NUMBER)
```
mattkime: need to remove line 2!

PR #86, opened 2017-02-14
Move from npm to yarn

sadafie:
I accidentally already added a `yarn.lock` file with my Greenkeeper Monday tasks yesterday, but this completes the task. It also adds the eslint precommit hook at @eilinora 's request.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,8 @@ const config = {
 		"es6": true
 	},
 	"plugins": [
-		"react"
+		"react",
+		"meetup"
 	],
 	"rules": {
 		"array-callback-return": 2,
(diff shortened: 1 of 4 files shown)


eilinora:
@sadafie does this add the pre-commit check for eslint? :crossed_fingers: doesn't look like it

sadafie:
> does this add the pre-commit check for eslint? 

Yep, the following in `package.json` should do the trick. Let me know if it doesn't work for you.

```
"precommit": "git status --porcelain | awk 'match($1, \"[ACM]\"){print $2}' | grep '[js|jsx]$' | xargs eslint",
```

mattkime:
@sadaf the precommit hook needs `husky` added as a dependency.

eilinora:
shwoo not crazy! i had tried adding the precommit in package but was missing the `husky` requirement

sadafie:
ah okay, we can file a follow up ticket

PR #88, opened 2017-02-17
Sds 156 datetime

unjust:
#### Related issues

Fixes https://meetup.atlassian.net/browse/SDS-156

#### Description
A Datetime component that defaults to the native `datetime-local` (if supported) or enhances with [flatpickr](https://chmln.github.io/flatpickr/).
We want to use the native component on mobile browsers specifically.

screen shot 2017-03-17 at 1 00 26 pm
screen shot 2017-03-17 at 1 00 33 pm
screen shot 2017-03-17 at 1 00 50 pm

Just time input
screen shot 2017-03-01 at 4 39 05 pm

will eventually be something like 
https://meetup.atlassian.net/wiki/display/SDS/Date+and+Time+Inputs


unjust:
Question: While building this out, does it make more sense to separate the `flatpickr` enhanced DateTime from a regular DateTime, and have some container that decides between the two based on support.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * @module DateTimePicker
+ */
+class DateTimePicker extends React.Component {
+
+	constructor(props) {
+		super(props);
+		this.state = {
+			value: props.value || '',
+			isSupported: false
```
mmcgahan: `isSupported` is vague - is what supported?

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,102 @@
...
+	componentWillMount() {
+		const input = document.createElement('input'),
+			invalidValue = 'notadate';
+
+		input.setAttribute('type', 'datetime-local');
+		input.setAttribute('value', invalidValue);
+
+		// adding false to view storybook
+		// in chrome since chrome supports
+		const isInvalidValueDenied = (false && input.value !== invalidValue);
```
mmcgahan: probably simpler/clearer to just call this `isSupported`.

Might be easier to have an optional boolean prop to force 'not supported' rather than hard-coding `false &&` in development.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,102 @@
...
+	componentDidMount() {
+		if (!this.state.isSupported) {
+			const options = {
+				...this.props.datepickerOptions,
+				onChange: this.onChange,
+				altInput: true,
+				altFormat: 'M d, Y h:i K' // TODO localize
+			};
+			this.flatpickr = new Flatpickr(this.node, options);
+		}
```
mmcgahan: a preferred pattern is to exit early rather than wrapping a whole function inside an `if` block:

```js
componentDidMount() {
  if (this.state.isSupported) {
    return;
  }
  // ...
}
```

it reduces nesting without hurting understandability

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,102 @@
...
+		this.flatpickr && this.flatpickr.setDate(this.state.value);
+	}
+
+	componentWillUnmount() {
+		this.flatpickr && this.flatpickr.destroy();
+	}
+
+	// conforms with flatpickr onChange args
+	onChange(selectedDates, dateStr, instance) {
+		const newValue = (this.flatpickr) ? selectedDates[0] : selectedDates.target.value;
```
mmcgahan: just call it `value`. parens are not necessary

src/DateTimePicker.jsx
unjust: whether or not datetime-local is supported natively. I'll make this round of changes and also try to clarify some things.

assets/scss/main.scss
```diff
@@ -15,3 +15,6 @@
 // COMPONENT SPECIFIC
 //
 @import "popover";
+
+// TODO: get this imported at the component level
+@import '../../node_modules/flatpickr/dist/flatpickr.css';
```
akdetrick: We should let webpack resolve imports from `node_modules`.

```
@import "~flatpicker/dist/flatpickr.css";
```

In the webpack `sass-loader`, `~moduleName` is an alias for the path to `node_modules`. This is important because the relative path to the node modules directory may not be the same in every repo.

src/DateTimeLocalInput.jsx
```diff
@@ -0,0 +1,51 @@
...
+				value={this.state.value}
+				className={classNames}
+				onChange={this.onChange}
+				{...other} />
+		);
+
+	}
+}
+
+DateTimeLocalInput.propTypes = {};
```
akdetrick: missing propType validation for `callback` and `id`

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,120 @@
...
+			<div>
+				<label htmlFor={id}>{label}</label>
+				<Flatpickr callback={this.setDate} value={this.state.value.date} opts={datepickerOptions} />
+				<TimeInput callback={this.setTime} value={this.state.value.time} />
+			</div>
+		);
+	}
+}
+
+DateTimePicker.PropTypes = {
```
akdetrick: missing prop validation for `datepickerOptions` and `value`

src/dateTimePicker.story.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import DateTimePicker from './DateTimePicker';
+import { storiesOf } from '@kadira/storybook';
+
+storiesOf('DateTimePicker', module)
+	.add('default', () => {
```
akdetrick: use `addWithInfo` (see `popover.story.jsx` for an example). The `InfoWrapper` component gives us a nice documentation add-on

src/DateTimePicker.jsx
mmcgahan: in this case it looks like `value` is not being used as a pass-through string `value` for an input, but is instead an object - that might be a little confusing, but would definitely need propTypes to clarify that it should be an object, not a string like it would be for plain HTML elements.

unjust:
@akdetrick @mmcgahan please see the review changes I have made.

There are a few components in this review: 
- DateTimePicker: wraps a Flatpickr and TimeInput, decides whether to render a DateTimeLocalInput if supported ( just realized Flatpickr defaults to datetime-local on mobile so maybe I wont need DateTimeLocalInput and can favor that )
- TimeInput: a standalone time input
- DateTimeLocalInput: datetime-local input
- Flatpickr - the Reactified Flatpickr 

I'm still waiting on final designs so not completely set how these smaller components fit together may change.

src/DateTimeLocalInput.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import cx from 'classnames';
+
+/**
+* @module DateTimeLocalInput
```
mmcgahan: more descriptive docs

src/DateTimeLocalInput.jsx
```diff
@@ -0,0 +1,66 @@
...
+		);
+
+	}
+}
+
+DateTimeLocalInput.propTypes = {
+	id: React.PropTypes.string,
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.string,
```
mmcgahan: string or a string, eh? ;)

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,131 @@
...
+			isDateTimeLocalSupported: false
+		};
+
+		this.setDate = this.setDate.bind(this);
+		this.setTime = this.setTime.bind(this);
+		this.parseDateTime = this.parseDateTime.bind(this);
+	}
+
+	componentWillMount() {
+		this.setState({ isDateTimeLocalSupported: this.hasBrowserSupport() });
```
mmcgahan: `componentWillMount` [will be called on the server](https://facebook.github.io/react/docs/react-component.html#componentwillmount) where `document` will be undefined.

Move this to `componentDidMount`

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,131 @@
...
+		const input = document.createElement('input'),
+			invalidValue = 'notadate';
+
+		// some browsers (as Android stock browsers) pretend they support
+		// certain input types,
+		// so we set the value to an invalid value and see if browser rejects
+
+		input.setAttribute('type', 'datetime-local');
+		input.setAttribute('value', invalidValue);
+
```
mmcgahan: this seems like it would be a good general utility function - not sure if it should go here or in meetup-web-platform, but it would be good to put it somewhere reusable

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,131 @@
...
+	setDate(value) {
+		this.setState({ value: { date: value }});
+	}
+
+	setTime(value) {
+		this.setState({ value: { time: value }});
+	}
+
+	parseDateTime(value) {
+		const datetime = value.split('T'),
```
mmcgahan: this is probably fine, but it appears fragile. Any way we can use the `Date` API to work with this? Something like

```js
const datetime = new Date(value);
const newState = {
  date: datetime.getDate();
  time: `${datetime.getHours()}:${datetime.getMinutes()}`
};
```

or something - maybe that's overkill. Alternatively, you should document the expected sting format of `value` here so that it's more clear what is happening

---

After thinking about it a bit more, the parsing + getting of time strings is a very manual process that should ideally be handled entirely by JavaScript Date objects - the less you have to manipulate time string values, the better.

To do that, you can look into keeping a real `Date` object in state, and then getting/setting time and date info from a single coherent object rather than trying to combine and separate separate `date` and `time` strings.

This function would then be as simple as

```js
parseDateTime(value) {
  this.setState({ datetime: new Date(value) });
}
```

`getDateTime` would be

```js
getDateTime() { return this.state.datetime; }
```

`setDate` and `setTime` would be slightly more complicated, but in a good way IMO:

```js
setDate(value) {
  const newDate = new Date(value);
  const datetime = new Date(this.state.datetime);  // copy current state
  datetime.setFullYear(  // update only the date part
    newDate.getFullYear(),
    newDate.getMonth(),
    newDate.getDate()
  );
  this.setState({ datetime });  // apply setting
},
setTime(value) {
  const newTime = new Date(value);
  const datetime = new Date(this.state.datetime);  // copy current state
  datetime.setHours(  // update only the time part
    newDate.getHours(),
    newDate.getMinutes(),
  );
  this.setState({ datetime });  // apply setting
}
```

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,131 @@
...
+			{required},
+			className
+		);
+
+		const labelClassNames = cx({required});
+		if (this.state.isDateTimeLocalSupported) {
+
+			const dateTimeLocalOpts = {
+				min: `${datepickerOptions.minDate}T00:00:00`,
+				max: `${datepickerOptions.maxDate}T00:00:00`
```
mmcgahan: check out `new Date(datepickerOptions.maxDate).toIsoString()` - it includes UTC timezone info so it might not work as expected out of the box, but it avoids having to manually construct an ISO string

src/FlatpickrComponent.jsx
```diff
@@ -21,12 +21,13 @@ class FlatpickrComponent extends React.Component {
...
 	componentDidMount() {
 		const options = {
-			...this.props.opts,
 			onChange: this.onChange,
 			altInput: true,
 			altFormat: 'D M d, Y', // TODO localize
 			defaultDate: this.props.value
 		};
+
+		Object.assign(options, this.props.opts);
```
mmcgahan: why this change? The object spread operator is effectively the same as `Object.assign` but it enforces immutability.

If you move the `...this.props.opts` down to the bottom of the `const options` object, it will correctly override the defaults.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,159 @@
...
+ * a component that renders a calendar ui and time input
+ * defaults to datetime-local on supported mobile browsers
+ */
+class DateTimePicker extends React.Component {
+
+	constructor(props) {
+		console.log(props.value);
+		super(props);
+		this.state = {
+			datetime: props.value ? new Date(props.value).toISOString() : '',
```
mmcgahan: you're initializing `state.datetime` to a string, but `setTime` will save it as a `Date` object - it should always be one or the other, and the `Date` object is going to be easier to work with.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import cx from 'classnames';
+import Flatpickr from './FlatpickrComponent';
```
mmcgahan: yeah, probably keep it as `FlatpickrComponent` or event change filenames to make it `ReactFlatpickr` or something - that would potentially be open-sourceable for other devs to use.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,185 @@
...
+		this.setState({ datetime });
+	}
+
+	setDateTime(value) {
+		this.setState({ datetime: new Date(value) });
+	}
+
+	render() {
+		const {
+			callback,			// eslint-disable-line no-unused-vars
```
eilinora: nitpicky weird spacing :)

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,185 @@
...
+	*/
+	setDate(value) {
+		const newDate = new Date(value);
+		const datetime = new Date(this.state.datetime);
+
+		datetime.setFullYear(newDate.getFullYear(), newDate.getFullMonth(), newDate.getDate());
+		this.setState({ datetime });
+	}
+
+	parseTimeFromDateTime(datetime) {
```
eilinora: some docs around this would be great! A bit unsure why you need to split it like this.

src/DateTimeLocalInput.jsx
```diff
@@ -0,0 +1,76 @@
...
+			className
+		);
+
+		const labelClassNames = cx({required});
+		return (
+			<div>
+				<label htmlFor={id} className={labelClassNames}>{label}</label>
+				<input
+					id={id}
+					type='datetime-local'
```
eilinora: datetime-local doesn't work everywhere we support. http://caniuse.com/#search=datetime-local

eilinora:
also, also, @mperrotti and @unjust is the calendar component accessible? navigating calendars are difficult with screenreaders. Have we looked at text input option?

src/DateTimeLocalInput.jsx
unjust: right, but this component is a fallback for DateTimePicker if datetime local is supported.

src/DateTimePicker.jsx
unjust: added. our state for DateTimePicker is datetime. SInce its a combined component in some cases (date picker + time input) this is used to retrieve the time value for the time input.

unjust:
@eilinora @mperrotti this should be given an accessibilty audit/check. However the CalendarComponent (flatpickr) is given the `allowInput:true` option which allows plain text input.

eilinora:
@unjust 🙌  🎺  to `allowInput:true`

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,240 @@
...
+		this.setState({ datetime });
+	}
+
+	/**
+	* @function parseTimeFromDateTime
+	* @description gets the time string part of our datetime state
+	* @return String
+	*/
+	parseTimeFromDateTime(datetime) {
+		return (datetime.toTimeString().split(' ')[0]).split(':', 2).join(':'); // TODO localize toLocaleTimeString ?
```
eilinora: this part always hangs me up about this piece, it's probably because of the pickers we are using, or maybe how datetime-locale gets data. But would be good to add some doc's or @see references to why we need to do this. Either here or at the top of the component for future dev's

src/dateTimePicker.test.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+	it('accepts a Date as a value', function() {
+		dateTimeComponent = TestUtils.renderIntoDocument(
+			<DateTimePicker name='start_time'
+				value={new Date(dateStr)}
+				forceCalendar />);
+		expect(dateTimeComponent.state.datetime).toEqual(new Date(dateStr));
+	});
+
+	it('accepts a datetime object as value', function() {
```
eilinora: seems like a duplicate of https://github.com/meetup/meetup-web-components/pull/88/files#diff-1da4556b5f5f8d13c410ac257b8e8281R45

src/dateTimeLocal.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+		callbackSpy = jasmine.createSpy('foo');
+
+	beforeEach(() => {
+		dateTimeComponent = TestUtils.renderIntoDocument(
+			<DateTimeLocalInput
+				name='datetime'
+				value={value}
+				callback={callbackSpy}
+				min={min}
+				max={max} />);
```
mmcgahan: good to align the closing `/>` with opening `<`, and closing `)` with the start of the line that has the opening `(`

src/timeInput.story.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import TimeInput from './TimeInput';
+import { storiesOf } from '@kadira/storybook';
+
+storiesOf('TimeInput', module)
+	.add('default', () => {
```
eilinora: addInfo would be good for this :)

src/timeInput.test.jsx
```diff
@@ -0,0 +1,59 @@
...
+import TimeInput from './TimeInput';
+
+describe('TimeInput', function() {
+
+	let timeInputComponent,
+		timeInputEl;
+
+	const timeValue = '22:00',
+		newTime = '23:00',
+		callbackSpy = jasmine.createSpy('foo');
```
eilinora: jasmine? do you mean jest?  alternative writing of this would be
`callbackSpy = jest.fn()`

src/dateTimeLocal.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+	it('should set a min value', function() {
+		expect(dateTimeInputEl.min).toEqual(min);
+	});
+
+	it('should set a max value', function() {
+		expect(dateTimeInputEl.max).toEqual(max);
+	});
+
+	it('should have a required attribute when specified', () => {
+		expect(dateTimeInputEl.attributes.required).not.toBeNull();
```
mmcgahan: `toBeNull` and `toBeUndefined` are always a little risky to use because they don't check for other falsy values - in this case, the test will _always_ pass because if `dateTimeInputEl` doesn't have a `required` attribute, `dateTimeInputEl.attributes.required` will be `undefined` (not `null`!), so the test passes.

You can use `expect(dateTimeInputEl.attributes.required).not.toBeFalsy();`

src/dateTimePicker.test.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import TestUtils from 'react-addons-test-utils';
+import DateTimePicker from './DateTimePicker';
+import CalendarComponent from './CalendarComponent';
+import TimeInput from './TimeInput';
+
+fdescribe('DateTimePicker', function() {
```
mmcgahan: `fdescribe => describe`

src/dateTimePicker.test.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+	// datetime.setHours(hours, minutes);
+
+	beforeEach(() => {
+		dateTimeComponent = TestUtils.renderIntoDocument(
+			<DateTimePicker name='start_time'
+				value={datetime}
+				forceCalendar />);
+		calendarComponent = TestUtils.findRenderedComponentWithType(dateTimeComponent, CalendarComponent);
+		timeComponent = TestUtils.findRenderedComponentWithType(dateTimeComponent, TimeInput);
```
mmcgahan: `timeComponent` is only used in one test - no need to define it in `beforeEach`

`calendarComponent` is only used in one test, and that test makes its own value assignment, so the `beforeEach` assignment is redundant

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,240 @@
...
+		} = this.props;
+
+		const classNames = cx(
+			'dateTimePicker',
+			{calendarTimeComponent : !dateOnly },
+			{required},
+			className
+		),
+			labelClassNames = cx({required}),
+			timeInputName = `${name}-time`;
```
mmcgahan: new `const` declaration for each new const is another option - I've stopped doing comma-delimited variable declarations entirely, and I feel like it has had a (small) benefit in scan-ability and clean diffs (no more managing commas-vs-semicolons).

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	/**
+	* @function onBlur
+	* @param Event e
+	* @description called when either of the date or time inputs is blurred
+	* takes off focus class when neither of them are in focus
+	*/
+	onBlur(e) {
+		const timeInput = ReactDOM.findDOMNode(this.timeComponent).querySelector('input[type=time]');
```
mmcgahan: just a mild caution against inspecting child components from a parent - it's essentially reversing the one-way data flow that React is designed to support, so passing down a callback is generally preferred.

references:
- https://facebook.github.io/react/docs/refs-and-the-dom.html#dont-overuse-refs
https://facebook.github.io/react/docs/lifting-state-up.html

src/DateTimePicker.jsx
unjust: ok I'll add more!

src/DateTimePicker.jsx
unjust: if you still think this is the case, let me know where to move it, although I could have sworn we said it was component specific after all.

src/DateTimePicker.jsx
```diff
@@ -0,0 +1,240 @@
...
+						}
+					</div>
+				</div>
+			</span>
+		);
+	}
+}
+
+DateTimePicker.propTypes = {
+	callback: React.PropTypes.func,
```
unjust: I forget if this is even necessary. I think I should remove.

src/TimeInput.jsx
```diff
@@ -0,0 +1,72 @@
...
+					{...other} />
+				{ error && <p className='text--error'>{error}</p> }
+			</span>
+		);
+
+	}
+}
+
+TimeInput.propTypes = {
+	callback: React.PropTypes.func,
```
unjust: dr. mike agrees - going with `onChangeCallback`

src/dateTimeLocal.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+		dateTimeInputEl = TestUtils.findRenderedDOMComponentWithTag(dateTimeComponent, 'input');
+	});
+
+	afterEach(() => {
+		dateTimeComponent = null;
+		dateTimeInputEl = null;
+	});
+
+	it('should exist', function() {
+		expect(dateTimeInputEl).not.toBeNull();
```
unjust: right! I think I missed this one. thx

src/dateTimePicker.test.jsx
unjust: just testing you guys ;) hehehe

unjust:
whats left to do here: 
- any remaining PR feedback. maybe I didn't get the ReactDOM issues in DateTimePicker sufficiently worked out. (may be of interest to @yanyi if he wants to finish it up)
- accessibility checks
- further styling, maybe @mperrotti can jump in to customize and to a browserstack check.

src/DateTimePicker.jsx
mmcgahan: yeah, there are a couple problems with the current implementation. First, you're manipulating the DOM through `this.backgroundEl.classList` in `onFocus` and `onBlur` rather than using `this.setState` to trigger a re-render and set the right `className` in JSX. You shouldn't ever directly manipulate the DOM in React - use the component's render method and read from `this.state` and `this.props`. If you don't do that, React could re-render your component and blow away any changes that you've manually applied.

To a slightly lesser extent, the `querySelector` usage is also problematic because you are accessing the DOM independently of React - you can actually use the `ref` that you've already set up for `CalendarComponent` to access its `<input>` (`this.dateComponent.node` will get you there), and you can add a `ref` to the time `<input>` in `TimeInput` so that you can access it with something like `this.timeComponent.node` to avoid having to use `querySelector`.

mmcgahan:
Check out the coveralls report to see what your unit tests are missing, e.g. https://coveralls.io/builds/10655999/source?filename=src%2FCalendarComponent.jsx

burrata:
Accessibility: https://meetup.atlassian.net/browse/SDS-219
Style refinements/browser check: https://meetup.atlassian.net/browse/SDS-220

src/DateTimePicker.jsx
akdetrick: looks component-specific to me, as it deals specifically with `type='datetime-local'`

PR #110, opened 2017-03-06
upgrade stuff not caught by greenkeeper recently - components

mmcgahan:


mmcgahan:
Not sure how coverage changed, and getting a 500 error from coveralls when investigating the supposed change in RsvpBox coverage... inclined to ignore

PR #112, opened 2017-03-07
update tabs child validation

akdetrick:
**Fun fact:**

`React.createClass` returns an `Object`, not a `Class`, which means you can't validate in propTypes using `React.PropTypes.instanceOf(Component)`

---

This fixes the validation issue in `Tabs` propTypes.


src/Tabs.jsx
```diff
@@ -63,9 +63,17 @@ export class Tabs extends React.Component {
...
-	children: React.PropTypes.arrayOf(
-		React.PropTypes.instanceOf(TabsTab)
-	).isRequired,
+	children: (props, propName, componentName) => {
+		let error;
+		const expectedName = 'TabsTab';
+
+		React.Children.forEach(props[propName], child => {
+			if (typeof child === 'undefined' || child.type.name !== expectedName) {
+				error = new Error(`Children must be React elements of type ${expectedName}`);
```
mmcgahan: just return the error immediately rather than assigning it to a `let`.

src/Tabs.jsx
```diff
@@ -63,9 +63,17 @@ export class Tabs extends React.Component {
 	}
 }
 Tabs.propTypes = {
-	children: React.PropTypes.arrayOf(
```
mmcgahan: I think the problem might be that `children` is not a true array, so when it tries to check the type of `children[0]`, it finds an object that is not the actual child.

That means that you can use `React.Children.forEach` in your custom function, but _also_ use `instanceof TabsTab`

akdetrick: simplified this to `child.type !== TabsTab`, which seems to work

src/Tabs.jsx
```diff
@@ -63,9 +63,15 @@ export class Tabs extends React.Component {
...
-		React.PropTypes.instanceOf(TabsTab)
-	).isRequired,
+	children: (props, propName, componentName) => {
+		let error;
+		React.Children.forEach(props[propName], child => {
+			if (typeof child === 'undefined' || child.type !== TabsTab) {
+				error = new Error('Children must be React elements of type TabsTab');
```
mmcgahan: Ah, I see why you kept the `let error` setup since it is set inside a callback.

Just for fun, you can still avoid the `let error` (and mutable state) with:

```js
const validChildren = React.Children.map(
  children,
  child => child instanceof TabsTab
).every(x => x);  // check that all children are TabsTab

if (!validChildren) {
  return new Error(...);
}
```

src/Tabs.jsx
```diff
@@ -63,9 +63,15 @@ export class Tabs extends React.Component {
 	}
 }
 Tabs.propTypes = {
-	children: React.PropTypes.arrayOf(
-		React.PropTypes.instanceOf(TabsTab)
-	).isRequired,
+	children: (props, propName, componentName) => {
+		let error;
+		React.Children.forEach(props[propName], child => {
+			if (typeof child === 'undefined' || child.type !== TabsTab) {
```
mmcgahan: I recommend separating out your two `child` validation conditions - use [`React.Children.count`](https://facebook.github.io/react/docs/react-api.html#react.children.count) to return an Error about passing in _no_ children, and use the `forEach` that you have here to check the type of the children.

The separation will make it a little more clear what is needed and simplify the code a bit

```js
const children = props[propName];
if (!React.Children.count(children)) {
  return new Error(...);
}
```

src/Tabs.jsx
akdetrick: nice!

src/Tabs.jsx
akdetrick: This is good stuff. I can actually use this to require that at least 2 `TabsTab` children are present.

eilinora:
bump, is there anything outstanding on this one? Would be nice to cleanup the unnecessary error messaging that this would fix

PR #118, opened 2017-03-08
Textarea

mperrotti:
#### Related issues
Fixes: https://meetup.atlassian.net/browse/SDS-183
Will be used in EventCreate, MUP comments, and MUG comments

#### Description
A textarea that has the option to resize based on content

screen shot 2017-03-08 at 2 52 01 pm
screen shot 2017-03-08 at 2 52 16 pm


mperrotti:
Just started on this. I mostly want to know if it's ok to use [autosize](https://github.com/jackmoore/autosize) instead of writing our own JS to handle the resizing bit.

src/textarea.story.jsx
```diff
@@ -0,0 +1,45 @@
+
+import React from 'react';
+import Textarea from './Textarea';
+import { storiesOf } from '@kadira/storybook';
+
+storiesOf('Textarea', module)
+	.add('default', () => <Textarea
```
eilinora: dont forget to use the addInfo so we get some of the nice doc's for the component

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	onChange(e) {
+		this.setState({
+			value: e.target.value,
+		});
+	}
+
+	componentDidMount() {
+		if (this.props.autoheight){
```
eilinora: Does this have to be after mount vs on render? I'm assuming its because it needs to be rendered before it can do its thing, maybe?

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+			labelClassName
+		);
+
+		const heightConstraints = {
+			minHeight: minHeight,
+			maxHeight: maxHeight
+		};
+
+		return (
+			<div>
```
eilinora: probably be good to make this a fieldset

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+				<label className={labelClassNames} htmlFor={other.id}>
```
eilinora: Let's make `id` a required prop, seems weird to have this being pulled from other yet also expected, which means also be explicit in textarea `<textarea id={id}>`

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element
+	]),
+	labelClassName: React.PropTypes.string,
+	required: React.PropTypes.bool,
+	autoHeight: React.PropTypes.bool,
+	minHeight: React.PropTypes.number,
+	maxHeight: React.PropTypes.number,
+	maxLength: React.PropTypes.number
```
eilinora: super nitpicky! can we move this up the list, gets lost in reading it with the height props.

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+	name: React.PropTypes.string.isRequired,
+	error: React.PropTypes.string,
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element
+	]),
+	labelClassName: React.PropTypes.string,
+	required: React.PropTypes.bool,
+	autoHeight: React.PropTypes.bool,
+	minHeight: React.PropTypes.number,
```
eilinora: is this required if you have autoheight? if so we should do some custom prop type validation

src/textarea.test.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+import TextareaContainer from './TextareaContainer';
+
+describe('TextareaContainer', function() {
+
+	it('exists', function() {
```
eilinora: obvious, but we need more tests :)

mperrotti:
I haven't heard any feedback against using [autosize](https://github.com/jackmoore/autosize), so I'm going to move forward

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+			<div>
+				<label className={labelClassNames} htmlFor={other.id}>
+					{label}
+				</label>
+				<textarea type='text'
+					name={name}
+					required={required}
+					className={classNames}
+					onChange={this.onChange}
+					rows={autoheight ? 1 : 'auto'}
```
mmcgahan: by default a textarea will only be 1 row high? I recommend a larger default so that it doesn't look like a single-line text input

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+					style={{ ...(style || {}), ...heightConstraints }}
+					value={this.state.value}
+					{...other}
+					>
+				</textarea>
+
+				{ maxLength && <p className='text--caption align--right'>{this.state.value.length} / {maxLength}</p> }
+
+				{ error && <p className='text--error'>{error}</p> }
+				{children}
```
mmcgahan: let's leave the children out of this - it doesn't really make sense to support passing in child components of a textarea

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+	onChange(e) {
+		this.setState({
+			value: e.target.value,
+		});
+	}
+
+	componentDidMount() {
+		if (this.props.autoheight){
+			autosize(this.textarea);
```
mmcgahan: the only thing I'm wondering about this is if the 'autosize' value might sometimes get lost when the element re-renders, e.g. when navigating away from the page and hitting 'back'. Worth doing some experimenting about whether you need to re-initialize it on every `render` but hopefully this `componentDidMount` is sufficient

src/Textarea.jsx
mperrotti: Some of the designs (e.g.: mug/mup comm) want it to appear that way. In the case where you want a textarea that has `autoheight` and actually looks like a textarea, you can set a `minHeight`.

src/Textarea.jsx
mperrotti: Agreed 👍

src/Textarea.jsx
mmcgahan: hm, in that case maybe `autoheight` should be called `rows` and default to `1`. If it's `'auto'`, then you can initialize `autosize`, but otherwise allow `rows` to be configurable - that setup would keep it more consistent with the underlying HTML element API, which is nice

mperrotti: Oooh, I like that

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+					{label}
+				</label>
+				<textarea type='text'
+					name={name}
+					required={required}
+					className={classNames}
+					onChange={this.onChange}
+					rows={autoheight ? 1 : 'auto'}
+					ref={(textarea) => {this.textarea = textarea;}}
+					style={{ ...(style || {}), ...heightConstraints }}
```
mmcgahan: give `style` a default value when you first assign it (`const { style={} } = this.props`) or from a `Textarea.defaultProps` to avoid this nested `||` fallback

src/Textarea.jsx
```diff
@@ -0,0 +1,107 @@
...
+					value={this.state.value}
+					{...other}
+					>
```
mmcgahan: `<textarea>` is self-closing in React - you can just un-indent this and add a slash `/>`

src/Textarea.jsx
mperrotti: I'm pretty sure having it happen after mount is the safest way, but I'll see what happens when I put it in render

src/Textarea.jsx
mperrotti: Our `TextInput` component is wrapped in a `<div>`, so I assumed we'd want `Textarea` to be similar.
If you feel strongly about using `<fieldset>` instead, should we update `TextInput` to match?

src/Textarea.jsx
mperrotti: Nothing is required with autoheight

src/Textarea.jsx
mmcgahan: `componentDidMount` is definitely preferred if it works since it will only get called once for the lifecycle of the element, but if the autosizer loses its reference to the DOM element when React re-renders (without unmounting) the component, then you'll have to call it in the render method - I think just play around with it and see if it ever loses the `autosize` functionality in unexpected ways

src/Textarea.jsx
akdetrick: :ballot_box: 
I would vote on reserving `fieldset` for grouping more than one form field

assets/scss/components/_textarea.scss
```diff
@@ -0,0 +1,3 @@
+textarea {
+	min-height: 0;
```
akdetrick: Should we be using `0` or `$font-size * $line-height`?

mperrotti: Having a `min-height` messes up the autoresizing functionality. I submitted a [PR in swarm-sasstools](https://github.com/meetup/swarm-sasstools/pull/7), so we can exclude this file from the PR altogether

akdetrick: merged that PR. You can use `1.1.10` of `swarm-sasstools` in this branch now

src/Textarea.jsx
```diff
@@ -0,0 +1,110 @@
...
+	constructor(props) {
+		super(props);
+		this.state = {
+			value: props.value || '',
+		};
+		this.onChange = this.onChange.bind(this);
+	}
+
+	onChange(e) {
+		this.setState({
```
mmcgahan: I think the `TextInput` also has an optional `onChange` prop that can be passed in so that the parent can be notified of changes - I'd recommend going ahead and adding that here

```js
onChange(e) {
  const { onChange } = this.props;
  if (onChange) {
    onChange(e);
  }
  this.setState({ value: e.target.value });
}
```

src/Textarea.jsx
```diff
@@ -0,0 +1,114 @@
...
+		};
+		this.onChange = this.onChange.bind(this);
+	}
+
+	onChange(e) {
+		const { onChange } = this.props;
+		if (onChange) {
+			onChange(e);
+		}
+		this.setState({
```
eilinora: pretty close, i think mikes comment was slightly off.
you still want to set the state and then call onChange

https://github.com/meetup/meetup-web-components/pull/134/files#diff-f9ace01f309ab9526a85b595c571df9eR17

src/Textarea.jsx
```diff
@@ -0,0 +1,114 @@
...
+					onChange={this.onChange}
+					rows={rows == 'auto' ? 1 : rows}
+					ref={(textarea) => {this.textarea = textarea;}}
+					style={{ ...style, ...heightConstraints }}
+					value={this.state.value}
+					id={id}
+					{...other}
+				/>
+
+				{ this.props.maxLength && <p className='text--caption align--right'>{this.state.value.length} / {this.props.maxLength}</p> }
```
eilinora: sorry nitpick can you break this up a into multiple lines, getting a bit hard to read

```
{ this.props.maxLength &&
   <p className='text--caption align--right'>
      {this.state.value.length} / {this.props.maxLength}
   </p>
}
```

src/Textarea.jsx
```diff
@@ -0,0 +1,117 @@
...
+	label: React.PropTypes.oneOfType([
+		React.PropTypes.string,
+		React.PropTypes.element
+	]),
+	labelClassName: React.PropTypes.string,
+	required: React.PropTypes.bool,
+	autoHeight: React.PropTypes.bool,
+	minHeight: React.PropTypes.number,
+	maxHeight: React.PropTypes.number,
+	rows: React.PropTypes.oneOfType([
```
mmcgahan: needs an `onChange: React.PropTypes.func`

PR #120, opened 2017-03-08
Adding action to group card

jmhammons:
#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-32

#### Description

Looking to extend the GroupCard component to accept and optional 'action' (or whatever we decide to call it) element.

For example: group cards in the register onboarding flow have a (+) icon in the top right of the card so members can one-touch-join groups. If they are already a member of the group the (+) is replaced by a check icon.

Meetup classic group card with (+) icon
screen shot 2017-03-08 at 12 11 50 pm

### DIFF
--- a/src/GroupCard.jsx
+++ b/src/GroupCard.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import cx from 'classnames';
+import Flex from './Flex';
+import FlexItem from './FlexItem';
 
 /**
  *	SQ2 Group Card component
@@ -26,6 +28,24 @@ export class GroupCard extends React.Component {
 		const photoUrl = group.duotoneUrl || (group.key_photo || group.group_photo || {}).thumb_link;
 		const backgroundImage = photoUrl && `url(${photoUrl})`;
 
+		const CardContent = ({group}) => (
+			<div>
+				<h4 className='card--group-content-name'>{group.name}</h4>
+				<p className='card--group-content-members'>{group.members} {group.who}</p>
+			</div>
+		);
+
+		const CardContentWithAction = ({group, action}) => (
+			<Flex>
+				<FlexItem>
+					<CardContent group={group} />
+				</FlexItem>
+				<FlexItem shrink>
+					{action}
+				</FlexItem>
+			</Flex>
+		);
+
 		return (
 			<span
 				className={cardClassNames}
@@ -33,8 +53,12 @@ export class GroupCard extends React.Component {
 				{...other}>
 
 				<div className='card--group-content'>
-					<h4 className='card--group-content-name'>{group.name}</h4>
-					<p className='card--group-content-members'>{group.members} {group.who}</p>
+					{this.props.action ? (
+						<CardContentWithAction group={this.props.group} action={this.props.action} />
+					) : (
+						<CardContent group={this.props.group} />
+					)}
+
 				</div>
 
 			</span>
@@ -44,6 +68,7 @@ export class GroupCard extends React.Component {
 
 GroupCard.propTypes = {
 	group: React.PropTypes.object.isRequired,
+	action: React.PropTypes.object,
 };
 
 export default GroupCard;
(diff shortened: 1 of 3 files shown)


src/GroupCard.jsx
```diff
@@ -26,15 +28,37 @@ export class GroupCard extends React.Component {
...
 			<span
 				className={cardClassNames}
 				style={{ ...(style || {}), backgroundImage }}
 				{...other}>
 
 				<div className='card--group-content'>
-					<h4 className='card--group-content-name'>{group.name}</h4>
-					<p className='card--group-content-members'>{group.members} {group.who}</p>
+					{this.props.action ? (
+						<CardContentWithAction group={this.props.group} action={this.props.action} />
```
jmhammons: I'm just realizing I should probably update the test for this component.

sadafie:
@jmhammons @eilinora @nlaz ---- sounds like group card might be being worked on by two separate teams? I just want to call it out here as well, in case there needs to be some coordination. Thanks!

eilinora:
Thanks for calling this out @sadafie!! Looking at the designs again and fortunately we are working on EventCard not GroupCard

But since I'm here :) this feels like a very mup-web specific thing, and we should probably migrate this to `/app/shared/components/` or `/app/groups/shared/components` in mup-web repo

sadafie:
I agree that this should live in `mup-web` repo (sorry @jmhammons!) next to EventCard, CategoryCard, and all the other composite components, in `/src/components/`

jmhammons:
Yeah, thanks @sadafie. 

@eilinora & @akdetrick -- I'm definitely down to put this in the appropriate spot. @akdetrick and I had initially discussed this as an alteration to the `Modal` comp here (unless I misunderstood something).

Since this has come up - I should probably take this opportunity to better understand how we make this determination. Can you elaborate a little on this line between `meetup-web-components` and `mup-web`?

sadafie:
My understanding (and I'll also let others chime in here), is that the components in this repo are the fundamental building blocks. They are small, and generic, and generally don't need complex actions etc. 

The components that live in `mup-web` are the meetup website specific "composite" components -- things that take these basic components and assemble them into an UI specific for the web.

eilinora:
@sadafie +1 on everything @sadafie said. Though we may have some complex components in here :) DateTimePicker that Ivy is working on.

jmhammons:
Ah ha -- so I've steered into "meetup website specific" territory here and this is not the place for that.

And to double confirm -- is `mup-web`'s `src/components/` the right spot for these changes (over the dir ideas like `src/app/shared` mentioned above by @eilinora).

sadafie:
yep, use `src/components/`! We intentionally removed the `src/app/shared` directory awhile back and moved everything to the top-level `src/components`.  You can think of everything in there as implicitly "shared"

sadafie:
> DateTimePicker that Ivy is working on

Fair point!

jmhammons:
Roger. Thank you thank you @sadafie / @eilinora. Apologies for any confusion (I'm new here). In any case, here I go.

eilinora:
no worries @jmhammons we all are 😸

jmhammons:
Oh @sadafie @eilinora,  I forgot to mention a thought based on this conversation. In the spirit of `meetup-web-components` not containing Meetup website specific things, I wonder if we should consider tweaking...

`meetup-web-components`
- `GroupCard.jsx`

To be:
`meetup-web-components`
- `Card.jsx`

And then in `mup-web`
- `GroupCard.jsx`
- `EventCard.jsx`
- `CategoryCard.jsx`

mmcgahan:
I talked with @jmhammons and he'll close this PR and move the GroupCard over to mup-web. If admin-web or pro-web ever need a group card or event card, we can decide what we want to do with both of them at that point - there might very well be some components that are more 'advanced' than meetup-web-components currently contains, but still things that need to be shared by multiple apps. They might even need to go in meetup-web-platform, but we'll make a plan when we need to.

PR #121, opened 2017-03-09
Copy testUtils from mocks repo into compenents repo

sadafie:
We want to minimize the number of dependencies that the meetup-web-mocks repo has, so we're going to have each consumer repo maintain it's own testUtils.jsx.

### DIFF
--- a/src/modal.test.js
+++ b/src/modal.test.js
@@ -1,8 +1,12 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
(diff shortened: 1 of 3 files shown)


eilinora:
is mocks repo going away?

sadafie:
No it's not. We're trying to pair it down to a zero-depedenc repo, so it's just the JS files for the mocks and nothing else.

sadafie:
> I think it will be easier long-term to leave testUtils.js out of this repo and just use scryRenderedComponentsWithType() instead 

done

PR #128, opened 2017-03-14
Member avatar (no photo variant)

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-186
Needed for mug/mup comm work

#### Description
Displays a placeholder avatar for users without a photo

screen shot 2017-03-14 at 10 32 53 am


jmhammons:
Awesome. I need this exact variant for `/register/photo`.

src/AvatarMember.jsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import cx from 'classnames';
 import Avatar from './Avatar';
+import { getIconAsBase64Uri } from './utils/base64';
+
+const NO_PHOTO_SRC = getIconAsBase64Uri('profile');
```
mmcgahan: I'm interested in this choice as opposed to `` with an SVG `src`? base64 strings can't be cached like other external assets, so this approach saves an http request but only in the first instance.

src/avatarmember.story.jsx
```diff
@@ -29,5 +29,9 @@ storiesOf('AvatarMember', module)
 			member={MOCK_MEMBER}
 			fbFriend={boolean('Facebook Friend', true)}
 		/>
-	));
+	)).add('no photo', () => {
+		const MOCK_MEMBER_NO_PHOTO = { ...MOCK_MEMBER };  // treat the mock as immutable
+		MOCK_MEMBER_NO_PHOTO.photo = {};
```
mmcgahan: another way to write this is `const MOCK_MEMBER_NO_PHOTO = { ...MOCK_MEMBER, photo: {} };`

src/AvatarMember.jsx
mperrotti: Because the icon is used as in a `background-image` rule

mmcgahan: apparently it's possible to use external SVGs in `background-image` rules - http://stackoverflow.com/a/9193203

Are there technical blockers for that? It's not a huge deal, but I am curious

mperrotti: We're rendering the icons from an SVG sprite, so there isn't a static SVG asset to point to in the `background-image` rule

mmcgahan: you're blowing my mind @mperrotti

jmhammons:
@mperrotti @mmcgahan -- Any chance this could get merged this week? Member growth could really use this for an upcoming launch, potentially next week.

src/utils/base64.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @function getIconAsBase64Uri
+ * @param {string} icon name
+ * @returns {string} base64-encoded data uri for SVG icon
+ */
+export const getIconAsBase64Uri = name => {
+	try {
+		return require(`base64-image!swarm-icons/dist/optimized/${name}.svg`);
+	} catch(e) {
+		return new Error(`Could not construct base64 uri for icon "${name}" \n ${e}`);
```
mmcgahan: does this `catch` work in practice? since the base-64 image is created by webpack at build time, the build should fail in case of an error, but this doesn't `throw` the error so I think it's probably not doing what it's intended to do at runtime

src/AvatarMember.jsx
```diff
@@ -21,15 +24,16 @@ class AvatarMember extends React.Component {
 			{
 				'avatar--org': org,
 				'avatar--fbFriend': fbFriend,
+				'avatar--noPhoto': (member.photo || {}).photo_link == undefined
```
mmcgahan: just use `'avatar--noPhoto': !(member.photo || {}).photo_link` - no need to check that it's a particular kind of falsy.

src/AvatarMember.jsx
```diff
@@ -21,15 +24,16 @@ class AvatarMember extends React.Component {
...
 			className);
 
 		return (
 			<Avatar
 				alt={member.name}
-				src={(member.photo || {}).photo_link}
+				src={(member.photo || {}).photo_link || NO_PHOTO_SRC}
 				className={classNames}
-				{...other} />
+				{...other}/>
```
mmcgahan: newline before `/>`

mmcgahan:
ah, just saw that @akdetrick merged. @mperrotti could you submit a PR to fix up the `throw`/`catch`?

src/utils/base64.js
akdetrick: I think the try/catch was my idea, but now it's obvious that it doesn't really do anything. We should just return the bit inside the `try`

akdetrick: https://github.com/meetup/meetup-web-components/pull/138

PR #131, opened 2017-03-16
cache yarn and node modules

mattkime:
Primary motivation is to avoid 502 errors but it will be interesting to see if there are build perf improvements.


mattkime:
@mmcgahan yes, but lets give this a trial run for a week just in case there are any hidden problems.

mattkime:
Might be too much variation to call it but it looks like the travis builds took 3-4min before this change. This took about 2min to build.

PR #132, opened 2017-03-16
set normal bounds to normal bounds width

akdetrick:
fixed Sass var for normal bounds max-width

### DIFF
--- a/assets/scss/components/_bounds.scss
+++ b/assets/scss/components/_bounds.scss
@@ -13,7 +13,7 @@ padding, and sets a max-width.
 	box-sizing: border-box;
 	margin-left: auto;
 	margin-right: auto;
-	max-width: $bounds-wide;
+	max-width: $bounds;
 	overflow: hidden;
 	padding: 0; // subtract for .chunk at bottom
 	width: 100%;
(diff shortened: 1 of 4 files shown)


assets/scss/components/_bounds.scss
```diff
@@ -13,12 +13,12 @@ padding, and sets a max-width.
 	box-sizing: border-box;
 	margin-left: auto;
 	margin-right: auto;
-	max-width: $bounds-wide;
+	max-width: $bounds;
```
eilinora: i thought we wanted bounds wide, and then perhaps there is a prop add to bounds narrow

nlaz: There's still a [`bounds--wide` class](https://github.com/meetup/meetup-web-components/pull/132/files/5173cbacc24b7998d180728ef8449455e3f00127#diff-887b5fa95ea8cd21b4b4f2d18e1ae9d2R23). But I don't think our `Bounds` components supports that as a prop yet.

eilinora: true, but if the goal is to use the bounds component feels like it should be an optional param there

akdetrick: Using an optional prop instead. (see updated diff)

PR #134, opened 2017-03-20
Adding onChange prop for TextInput

eilinora:
#### Description
DoD:
- Adding onChange prop so outside React components can get updates as input changes
- Adding a `getValue` function for form submissions

### DIFF
--- a/src/TextInput.jsx
+++ b/src/TextInput.jsx
@@ -19,10 +19,6 @@ class TextInput extends React.Component {
 		this.props.onChange(e);
 	}
 
-	getValue() {
-		return this.state.value;
-	}
-
 	render() {
 		const {
 			name,
--- a/src/TextInput.jsx
+++ b/src/TextInput.jsx
@@ -16,18 +16,24 @@ class TextInput extends React.Component {
 
 	onChange(e) {
 		this.setState({ value: e.target.value });
+		this.props.onChange(e);
+	}
+
+	getValue() {
+		return this.state.value;
 	}
 
 	render() {
 		const {
 			name,
-			value,	// eslint-disable-line no-unused-vars
+			value, // eslint-disable-line no-unused-vars
 			label,
 			labelClassName,
 			className,
 			children,
 			error,
 			required,
+			onChange, // eslint-disable-line no-unused-vars
 			...other
 		} = this.props;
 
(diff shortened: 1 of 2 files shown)


src/TextInput.jsx
```diff
@@ -75,7 +81,8 @@ TextInput.propTypes = {
 		React.PropTypes.element
 	]),
 	labelClassName: React.PropTypes.string,
-	required: React.PropTypes.bool
+	required: React.PropTypes.bool,
+	onChange: React.PropTypes.func.isRequired,
```
mmcgahan: this feels like it should be optional - does a parent have to do something every time the value changes?

src/TextInput.jsx
```diff
@@ -16,18 +16,24 @@ class TextInput extends React.Component {
 
 	onChange(e) {
 		this.setState({ value: e.target.value });
+		this.props.onChange(e);
+	}
+
+	getValue() {
```
mmcgahan: no need for a 'getter' on a (pseudo) read-only public property. A parent can read `theInput.state.value` just as easily as `theInput.getValue()`.

src/TextInput.jsx
```diff
@@ -16,18 +16,20 @@ class TextInput extends React.Component {
 
 	onChange(e) {
 		this.setState({ value: e.target.value });
+		this.props.onChange(e);
```
mmcgahan: if it's not required, `this.props.onChange` might be `undefined` - make sure you have a unit test with and without an `onChange` prop

src/textInput.test.jsx
```diff
@@ -18,7 +18,8 @@ describe('TextInput', function() {
 			id: NAME_ATTR,
 			maxLength: MAX_LEN,
 			error: ERROR_TEXT,
-			required: true
+			required: true,
+			onChange,
```
mmcgahan: one of the tricky aspects of relying on `beforeEach` is that it makes each unit test almost _too_ uniform - e.g. there's no test here that _doesn't_ set the optional `onChange` prop, so the tests missed out on the case there the component would fail if an `onChange` prop wasn't provided.

I think a good general rule might be to only provide _required_ props in `beforeEach` - every optional prop should be tested in isolation.

PR #135, opened 2017-03-20
remove node_module directory caching

mattkime:


mmcgahan:
Another thing we might want to consider is moving all of the `after_success` stuff into `after_script` so that it will fail the build

mattkime:
removed `.npmrc` to bring it in line with `meetup-web-platform`

mattkime:
>Another thing we might want to consider is moving all of the after_success stuff into after_script so that it will fail the build

I agree, but I'd like to do that separately. I consider that part of launch success feedback.

PR #139, opened 2017-03-21
Adding classname for tabstab + tests

eilinora:
DoD:
- TabsTab should add `prop.className` to `className` list + tests

### DIFF
--- a/src/tabs.test.jsx
+++ b/src/tabs.test.jsx
@@ -8,17 +8,14 @@ describe('Tabs', function() {
 
 	beforeEach(() => {
 		componentBasic = TestUtils.renderIntoDocument(
-			<Tabs>
-				<TabsTab>First tab</TabsTab>
+			<Tabs className='test-class-wrapper-name'>
+				<TabsTab className='test-class-name'>First tab</TabsTab>
 				<TabsTab isSelected>Second tab</TabsTab>
 				<TabsTab>Third tab</TabsTab>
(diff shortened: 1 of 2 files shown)


src/tabs.test.jsx
```diff
@@ -51,7 +48,7 @@ describe('Tabs', function() {
 		expect(tabsUlEl.classList).toContain('tabs--bordered');
 	});
 
-	it('only applies variant classes when set by props', function() {
+	it('should *not* apply variant classes when set by props', function() {
```
mmcgahan: `'should *not* apply variant classes when *not* set by props`

src/tabs.test.jsx
```diff
@@ -61,4 +58,15 @@ describe('Tabs', function() {
...
 	});
 
+	it('appropriately adds passed prop classname to component `Tabs` component', function() {
+		const tab = TestUtils.findRenderedDOMComponentWithClass(componentBasic, 'test-class-wrapper-name');
+		expect(() => tab).not.toThrow();
+		expect(tab.tagName).toBe('NAV');
+	});
+
+	it('appropriately adds passed prop classname to component `TabsTab` component', function() {
+		const tabsTab = TestUtils.findRenderedDOMComponentWithClass(componentBasic, 'test-class-name');
```
mmcgahan: set these classes as constants to make it easier to link the test input with the test output

PR #141, opened 2017-03-22
remove some unused travis.yml, build on pr and pull

mattkime:
followup from my debugging on monday. attempting to simplify and make uniform across repos.


mattkime:
Still not seeing coveralls results in the status checks but I think I'm going to set it aside for now.

.travis.yml
```diff
@@ -7,24 +7,16 @@ cache:
   directories:
     - node_modules
 
-branches:  # pushes only get built for master
-  only:
-    - master
```
mmcgahan: Overall this PR LGTM, but I'd keep the `only: master` build - I can't think of a reason why we would be interested in push-only builds for this repo. Not a huge deal, but it doubles the chance of PR build failures because there are two builds instead of one.

mattkime: >Overall this PR LGTM, but I'd keep the only: master build - I can't think of a reason why we would be interested in push-only builds for this repo.

I'd like to keep all our repos the same otherwise differences can be potential side effects for other tools, like Coveralls.

PR #142, opened 2017-03-22
alt version of avatarmember

eilinora:
Playing with alternate version without base64


assets/scss/components/_avatar.scss
```diff
@@ -112,25 +112,30 @@ Class                     | Description
...
 		position: absolute;
-		top: 0;
+		text-align: center;
 		width: 100%;
 	}
 
-	&.avatar--person {
-		background-position-y: 6px;
+	.svg-icon {
+		fill: $C_textHint;
```
akdetrick: I'd use `@color-svg($C_textHint)` here - this will help keep fill/stroke rules in one place if we update how icon styling works

PR #147, opened 2017-03-27
set a margin-top position based on scroll position; clean up story fo…

akdetrick:
Fixed storybook rendering of modal in `modal.story.js`; added state for top positioning of modal

#### Related issues

Fixes https://meetup.atlassian.net/browse/WC-33

screen shot 2017-03-27 at 12 29 20 pm


assets/scss/components/_modal.scss
```diff
@@ -5,8 +5,7 @@ $modal-width: 440px;
 	@include transition(transform .2s ease);
 	@include centerAbsolute($modal-width);
 	min-height: 0;
-	width: $modal-width;
-	margin-top: 10vh;
+	width: $modal-width !important;
```
mperrotti: Does this need to be `!important`?

src/Modal.jsx
```diff
@@ -31,6 +36,28 @@ class Modal extends React.Component {
...
 
+	/**
+	 * @param {String} scrollPosition window scroll position
+	 * @returns {String} CSS value for setting modal margin-top
+	 */
+	getModalPosition(scrollPosition) {
+		const viewportHeight = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
+
+		if (scrollPosition < viewportHeight) {
+			return DEFAULT_MARGIN_TOP;
```
mperrotti: In Storybook, the modal isn't "snapping" to the top because it has this default margin

src/Modal.jsx
```diff
@@ -60,19 +87,27 @@ class Modal extends React.Component {
...
-					<div className='inverted'></div>
-				</div>
+				{!fullscreen && overlayShim}
 
-				<div className={modalClasses} >
+				<div
+					className={modalClasses}
+					style={{marginTop: this.state.topPosition}}
+				>
 					<div className='align--right'>
```
mperrotti: Unrelated to this commit, but can we remove the "align--right" class? The close button should be aligned left.

akdetrick: I'm thinking we limit visual changes until we do our visual design sweep in the upcoming sprint

src/Modal.jsx
akdetrick: yeah, looks like we'll need to use JS matchMedia to sort this out when setting the `margin-top` in the component...

akdetrick:
Marked as `Prelim` until we merge a `matchMedia` utility for meetup-web-components

src/Modal.jsx
mmcgahan: @akdetrick could you describe more of what you're needing that can't be done in CSS?

akdetrick: In CSS, the default `margin-top` for a snap modal is `10vh`. The CSS also removes this margin if the viewport is below the `medium` breakpoint. 

If we set `margin-top` programmatically from the component, the component also needs to be aware of when we're below the `medium` breakpoint so that we can avoid forcing a margin on the modal when the viewport is small and the modal should be full screen.

akdetrick: This won't be the only component that will need to be aware of media queries, so there's a PR open to add utils for this kind of thing: https://github.com/meetup/meetup-web-components/pull/149

assets/scss/components/_modal.scss
akdetrick: It does for the sake of storybook. The only downside is that custom modal widths will also need `!important`, but it doesn't seem like a big problem.

akdetrick:
bump; this is a bit higher priority now, as onboarding needs it

src/Modal.jsx
```diff
@@ -15,6 +18,22 @@ class Modal extends React.Component {
 		super(props);
 		this.onDismiss = this.onDismiss.bind(this);
 		this.onKeyDown = this.onKeyDown.bind(this);
+		this.getModalPosition = this.getModalPosition.bind(this);
+
+		this.state = {
+			topPosition: DEFAULT_MARGIN_TOP // matches default margin-top in CSS
+		};
+
+		if (!this.props.fullscreen && typeof window.matchMedia != 'undefined') {
```
mmcgahan: this is going to crash on the server where `window` is not defined - I recommend putting it in `componentDidMount`, which will only run in the browser

src/Modal.jsx
```diff
@@ -15,6 +18,22 @@ class Modal extends React.Component {
...
+			this.mediaQuery = window.matchMedia(MEDIA_QUERIES.medium);
+
+			this.handleMediaChange = mq => {
```
mmcgahan: `componentDidMount` currently has the same structure - break it out into a named function that can be called from both places.

PR #149, opened 2017-03-28
Utils/constants match media

akdetrick:
Adds utilities for listening to media query changes from JS; adds `designConstants`.

## example usage
```
import MediaQueryable from 'react-media-queryable';
import { MEDIA_QUERIES } from './utils/designConstants';

// ... media query listener provider

	<MediaQueryable mediaQueries={MEDIA_QUERIES}>
		<MyComponent />
	</MediaQueryable>


// ... MyComponent gets a red outline at large sizes, for example

	<div
		style={
			{ outline: this.props.mediaQuery == 'large' ? '1px solid red' : 'none' }
		}
	>

```


src/utils/designConstants.js
```diff
@@ -0,0 +1,22 @@
+export const BREAKPOINTS = {
```
mmcgahan: `src/utils/` doesn't quite feel like the right place for constants, at least for a package that is just a library of nouns - I'm wondering about `src/constants/layout.js`, since `designConstants.js` is a little ambiguous, and it would be totally fine to make a new module for colors or whatever other kinds of variables we might want in the future.

src/utils/designConstants.js
```diff
@@ -0,0 +1,22 @@
...
+	m: '640px',
+	l: '840px',
+	xl: '1024px'
+};
+
+export const MEDIA_QUERIES = {
+	small: `screen and (min-width: ${BREAKPOINTS['s']})`,
+	medium: `screen and (min-width: ${BREAKPOINTS['m']})`,
+	large: `screen and (min-width: ${BREAKPOINTS['l']})`,
+	huge: `screen and (min-width: ${BREAKPOINTS['xl']})`
```
mmcgahan: Not sure why the linter doesn't pick up the unnecessary brackets in these `BREAKPOINT` references, but they should all be dot-separated references, e.g. `BREAKPOINT.s`, `BREAKPOINT.m`, for style consistency with the rest of the codebase.

src/utils/designConstants.js
akdetrick: This is an incremental step toward keeping our constants in one place. The long term plan is to have a `swarm-constants` module that provides vars/values in JS, Sass, and more

PR #151, opened 2017-03-29
DateTimePicker render should not depend on DOM; fix 3rd party CSS import

burrata:
#### Description

- fix style dependency on flatpickr
- flatpickr depends on dom, which breaks on server-side rendering in mup-web. remove dom-dependent aspects of datetimepicker.
- profit

https://meetup.atlassian.net/browse/SDS-226

### DIFF
--- a/src/CalendarComponent.jsx
+++ b/src/CalendarComponent.jsx
@@ -26,20 +26,18 @@ class CalendarComponent extends React.Component {
 		// lazy-loading flatpickr ensures it is
 		// imported only in clientside envs.
 		const Flatpickr = require('flatpickr');
-		if (Flatpickr) {
-			const options = {
-				onChange: this.onChange,
-				onOpen: this.onOpen,
-				onClose: this.onClose,
-				altInput: true,
-				allowInput: true,
-				altFormat: 'D M d, Y', // TODO localize
-				defaultDate: this.props.value
-			};
+		const options = {
+			onChange: this.onChange,
+			onOpen: this.onOpen,
+			onClose: this.onClose,
+			altInput: true,
+			allowInput: true,
+			altFormat: 'D M d, Y', // TODO localize
+			defaultDate: this.props.value
+		};
 
-			Object.assign(options, this.props.opts);
-			this.flatpickr = new Flatpickr(this.inputEl, options);
-		}
+		Object.assign(options, this.props.opts);
+		this.flatpickr = new Flatpickr(this.inputEl, options);
 	}
 
 	componentWillUnmount() {
--- a/src/CalendarComponent.jsx
+++ b/src/CalendarComponent.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import cx from 'classnames';
-import Flatpickr from 'flatpickr';
 
 /**
  * @module CalendarComponent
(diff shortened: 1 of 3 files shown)


src/CalendarComponent.jsx
```diff
@@ -22,18 +21,25 @@ class CalendarComponent extends React.Component {
...
-			altInput: true,
-			allowInput: true,
-			altFormat: 'D M d, Y', // TODO localize
-			defaultDate: this.props.value
-		};
+		// flatpickr uses `window` on import,
+		// which breaks on server-sider render.
+		// lazy-loading flatpickr ensures it is
+		// imported only in clientside envs.
+		const Flatpickr = require('flatpickr');
```
mmcgahan: `require()` is an interesting option - this will technically include flatpickr in both the server-side and client-side bundles, but it will only be actually imported in the client. It also uses webpack-specific syntax (since `require` isn't a native feature of browser JS). The alternative, using `import('flatpickr').then()` would actually make an async HTTP request for the module on the client, which means it's slower, but would also technically be compliant with [this TC39 proposal](https://github.com/tc39/proposal-dynamic-import). No need to change, just something to be aware of.

src/CalendarComponent.jsx
```diff
@@ -22,18 +21,25 @@ class CalendarComponent extends React.Component {
...
+		if (Flatpickr) {
```
mmcgahan: do you need this 'if'? You'll get an import error if it failed to find the module.

src/DateTimePicker.jsx
```diff
@@ -29,7 +29,12 @@ class DateTimePicker extends React.Component {
 		this.onBlur = this.onBlur.bind(this);
 	}
 
-	componentWillMount() {
+	// call `hasBrowserSupport` after mounting so server
+	// and client-side renders match, then immediately
+	// force a re-render. otherwise, client-side render
+	// will default to the server render, which never has
+	// browser support.
```
mmcgahan: Nitpick, but this should be formatted as a JSDoc

```js
/**
 * call ...
 * @returns {undefined} side effect only
 */
```

src/DateTimePicker.jsx
```diff
@@ -43,6 +48,10 @@ class DateTimePicker extends React.Component {
 			return;
 		}
 
+		if (typeof document === 'undefined') {
```
mmcgahan: trivial, but you might as well combine this check with the `if (this.props.forceCalendar)`

src/DateTimePicker.jsx
```diff
@@ -160,6 +169,7 @@ class DateTimePicker extends React.Component {
 			label,
 			value,	// eslint-disable-line no-unused-vars
 			required,
+			name,
```
mmcgahan: should this `name` also be sent to `<DateTimeLocalInput>`?

src/CalendarComponent.jsx
burrata: I originally had it as an `import` but was getting a syntax error on compile because of the top-level specification. strangely, it was compiling in mup-web but didn't in here.

src/DateTimePicker.jsx
burrata: good point -- it would be false for both, ha!

src/DateTimePicker.jsx
burrata: will look!

burrata: Looked—DateTimeLocalInput doesn't seem to take a name parameter.

burrata:
@akdetrick @eilinora will self-merge to unblock my work in mup-web if that's ok with you!

PR #154, opened 2017-04-04
use dependency instead of devDependency for base64-image-loader

akdetrick:
Moves the `base64-image-loader` package to `dependencies` from `devDependencies`. I believe this will prevent a missing module problem when using `Avatar` in other repos, but could use confirmation from someone more familiar with Yarn

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "react-intl": "^2.2.3"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


mmcgahan:
Yes, you are correct - since this package _doesn't work_ without `base64-image-loader`, it is a regular dependency

PR #156, opened 2017-04-04
fix package.json

akdetrick:
The recent `TextArea` merge broke the package.json - this should fix it.


akdetrick:
waiting for checks to pass just to be sure the dependencies are listed correctly, then will merge

PR #157, opened 2017-04-04
Bug/modal position on mount

akdetrick:
The modal positioning works, but only when a `matchMedia` change event fires. This fix ensures the modal positioning happens immediately on mount, and _then_ listens for media changes.


akdetrick:
agreed on the unit test - will add in a separate PR

PR #159, opened 2017-04-05
Add change callback to datetimepicker

burrata:
#### Related issues

Fixes https://meetup.atlassian.net/browse/SDS-236

#### Description
Add an onChangeCallback prop to the DateTimePicker component.


src/DateTimePicker.jsx
```diff
@@ -132,6 +132,7 @@ class DateTimePicker extends React.Component {
 	*/
 	setDateTime(value) {
 		this.setState({ datetime: new Date(value) });
+		this.props.onChangeCallback(this.state.datetime, this.props.name);
```
mmcgahan: this function name and signature are confusing

1. it's called `onChange...`, but it's called from `setDateTime`, which can technically be called at any time, not just when the input changes
2. `onChangeCallback` and `name` are supplied as props, which means that the parent already knows the value of `name` and shouldn't need for it to be returned as an argument. You can use currying to set up an `onChangeCallback` function that only relies on `this.state.datetime`, e.g.

    ```jsx
    // get a function that can use `name` without needing it as a new arg
    const getOnChangeCallback = name => datetime => { ... }; 
    const onChangeCallback = getOnChangeCallback(name);

    <DateTimePicker onChangeCallback={onChangeCallback} name={name} ... />
    ```

It's also a little strange that `onChangeCallback` doesn't take an `event` argument the way most `onX` event handlers do, but I see that convention isn't being used for any of the other form components at the moment, so probably no biggie

mmcgahan: since `onChangeCallback` is an optional prop, you'll need to check that it exists before calling it.

src/datetimepicker.test.jsx
```diff
@@ -202,4 +202,31 @@ describe('DateTimePicker', function() {
 		});
 
 	});
+
+	describe('onChangeCallback in props', () => {
+		let changeCallbackSpy;
+
+		beforeEach(() => {
+			changeCallbackSpy = spyOn(DateTimePicker.propTypes, 'onChangeCallback');
```
mmcgahan: this isn't spying on what you expect - the `propTypes` are functions that are called to validate props, not the actual values of the props.

You'll need to create a new spy, and I recommend using `jest.fn()`:

```jsx
const changeCallbackSpy = jest.fn();

<DateTimePicker name='start_time'
  value={dateStr}
  onChangeCallback={changeCallbackSpy}  // <-- supply the spy to the component
  forceCalendar
/>
```

src/datetimepicker.test.jsx
```diff
@@ -202,4 +202,31 @@ describe('DateTimePicker', function() {
...
+					onChangeCallback={function(){}}
+					forceCalendar
+				/>
+			);
+		});
+
+		it('is called when time input changes', () => {
+			const dateTimeInputEl = TestUtils.findRenderedDOMComponentWithTag(dateTimeComponent.timeComponent, 'input');
+			TestUtils.Simulate.change(dateTimeInputEl);
+			expect(changeCallbackSpy).toHaveBeenCalled();
```
mmcgahan: since you know the expected arguments for the spy, check those as well

```js
expect(changeCallbackSpy).toHaveBeenCalledWith(new Date(dateStr), 'start_time');
```

You might have to use `expect.any(Date)` for the first arg of `toHaveBeenCalledWith`, but try checking against a specific date first.

src/datetimepicker.test.jsx
burrata: very useful to know, thank you!

src/DateTimePicker.jsx
burrata: (1) I'll move this outta `setDateTime` and call it only when datetime state are being changed.
(2) `name` should probably removed, then -- that looks like a bit of product creep. Will take it out!
(3) Will look into adding the `event` arg just to start using the most common pattern.

src/DateTimePicker.jsx
```diff
@@ -250,7 +251,8 @@ DateTimePicker.propTypes = {
 		React.PropTypes.string
 	]),
 	dateOnly: React.PropTypes.bool,
-	forceCalendar: React.PropTypes.bool
+	forceCalendar: React.PropTypes.bool,
+	onChangeCallback: React.PropTypes.func
```
akdetrick: Our convention has been to match the prop name to standard event names; in this case, `onChageCallback` would just be `onChange`

akdetrick: or perhaps not, because it's not expecting an `event` arg

src/datetimepicker.test.jsx
burrata: So, I think because of timezone issues with the test, I'm going to try to expect.any(Date) value.

src/DateTimePicker.jsx
burrata: Yeah, that's the only thing about this. This element is relying on a state change being relayed on change rather than an event happening.

src/CalendarComponent.jsx
```diff
@@ -96,7 +104,7 @@ class CalendarComponent extends React.Component {
 				id={id}
 				type='text'
 				name={name}
-				defaultValue={this.state.value}
+				defaultValue={this.updateDate()}
```
mmcgahan: It feels a little strange to update a date at the same time that you are assigning a default - those seem like two separate operations. In some cases, the function also returns `undefined` as the default - is that expected?

src/DateTimePicker.jsx
```diff
@@ -187,7 +208,7 @@ class DateTimePicker extends React.Component {
 				<DateTimeLocalInput
 					id={id}
 					label={label}
-					value={this.state.datetime}
+					value={this.getDateTime()}
```
mmcgahan: this implementation forces an awkward separation of concerns - in effect, the parent of the `DateTimePicker` is responsible for both _setting_ and _reading_ its value. Form inputs should generally be [controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components), where they are responsible for keeping track of their own value. The parent can then provide an `onChange` (or similar) function that receives the value each time it changes - at the moment it looks like essentially you are using the `onChangeCallback` function from the parent as _both_ a watcher and an indirect setter, which can make the app structure less modular and maintainable.

burrata:
@mmcgahan @akdetrick The changes in here now:
- Add `onChangeCallback` to DateTimePicker
- Add `parseNewTimeAsDate` method with `moment` to fix a bug where changing input returns invalid date
- Set default datetime as current date, remove checks on `setDate` and `setTime` methods
- Use `setDateTime` as the change callback for `DateTimeLocalInput` (akin to the methods above)

package.json
```diff
@@ -38,7 +38,8 @@
   "dependencies": {
     "autosize": "3.0.20",
     "base64-image-loader": "1.2.0",
-    "flatpickr": "2.4.7"
+    "flatpickr": "2.4.7",
+    "moment": "2.18.1"
```
mmcgahan: we shoudn't need to use `moment` - it is a bloated dependency that gets even larger when adding i18n support, and using native `Date` tools + react-intl's [Date formatting components](https://github.com/yahoo/react-intl/wiki/Components#date-formatting-components) can usually get us where we need to be.

`moment.js` makes things easy, but at a pretty significant cost.

src/DateTimePicker.jsx
```diff
@@ -98,15 +113,20 @@ class DateTimePicker extends React.Component {
 		return (datetime.toTimeString().split(' ')[0]).split(':', 2).join(':'); // TODO localize toLocaleTimeString ?
 	}
 
+	parseNewTimeAsDate(timeStr) {
```
mmcgahan: This should get some docs to explain what the expect input-output formats are.

It looks like you just need to set the 'time' value on a particular 'date' object - you should be able to use the [Date object Setters](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date) to do that - since it looks like `timeStr` will represent the local timezone, parse it and use `setHours` and `setMinutes` on a `new Date(this.state.datetime)`.

If you really really need to get and use a timezone offset, use [`getTimezoneOffset`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset), but I'm hoping you won't need to.

src/DateTimePicker.jsx
```diff
@@ -118,20 +138,28 @@ class DateTimePicker extends React.Component {
...
-		const newTime = new Date(value);
-		const datetime = new Date(this.state.datetime);
+		const datetime = this.parseNewTimeAsDate(value);
 
-		datetime.setTime(newTime.getTime());
 		this.setState({ datetime });
+
+		if (this.props.onChangeCallback) {
+			this.props.onChangeCallback(datetime);
+		}
```
mmcgahan: since the only difference between `setTime` and `setDateTime` is how the value of `datetime` is determined, `setTime` can just be

```js
setTime(value) {
  const datetime = this.parseNewTimeAsDate(value);
  return setDateTime(datetime);
}
```

src/datetimepicker.test.jsx
```diff
@@ -218,4 +216,30 @@ describe('DateTimePicker', function() {
 		});
 
 	});
+
+	describe('onChangeCallback in props', () => {
+		let changeCallbackSpy;
+
+		beforeEach(() => {
+			changeCallbackSpy = jest.fn();
```
mmcgahan: the conventional way to use `jest.fn` is to assign it once, and then call [`mockFn.mockClear()`](https://facebook.github.io/jest/docs/mock-function-api.html#mockfnmockclear) each time you want to re-test it in a test suite.

src/DateTimePicker.jsx
burrata: aight, taking out moment and using react-intl!

burrata: Actually, looking more closely at the docs, the date formatting advantages are mostly in the FormattedDate component. Using the actual API from react-intl requires injecting it into the component, which created a host of testing problems when I first attempted it. This is probably why I ended up looking into using moment instead.

The `timeStr` returned does not return time zone. It's the time value on the time input in 24hr time. Adding docs now so that's clearer.

I'm going to look into doing this without moment again. Would prefer not to inject react-intl. Will see if I can do it without that, too.

src/forms/DateTimePicker.jsx
```diff
@@ -87,26 +90,62 @@ class DateTimePicker extends React.Component {
...
 	}
 
 	/**
+	* @function parseHoursAndMinutesFromTimeString
+	* @description takes a time string and returns an array of hours and minutes
+	* @param String timeString a 24-hour time string, eg `18:45`
+	* @return Array [0] - the hours, [1] - the minutes
+	*/
+	parseHoursAndMinutesFromTimeString(timeString) {
+		return timeString.split(':'); // TODO localization?
```
mmcgahan: do you have a ticket to follow up on this? should be relatively easy to check

src/forms/DateTimePicker.jsx
```diff
@@ -87,26 +90,62 @@ class DateTimePicker extends React.Component {
...
+
+
+	/**
+	* @function parseNewTimeAsDate
+	* @description takes a time string in 24hr time (e.g. `18:35`)
+	* and converts it to a Date object with the current state's date and timezone
+	* @return Date new date with current date and the given time
+	*/
+	parseNewTimeAsDate(timeStr) {
+		const datetime = this.state.datetime;
```
mmcgahan: ES6:

```js
const { datetime } = this.state;
```

src/forms/DateTimePicker.jsx
```diff
@@ -87,26 +90,62 @@ class DateTimePicker extends React.Component {
...
+		const parsedTime = this.parseHoursAndMinutesFromTimeString(timeStr);
+
+		datetime.setHours(parsedTime[0]);
+		datetime.setMinutes(parsedTime[1]);
```
mmcgahan: check out [the signature for `setHours`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours) - you can `setMinutes` by passing the minutes as a second arg.

PR #161, opened 2017-04-07
Adding yarn install after post merge/pull + minor template tweak 

eilinora:


eilinora:
going to SV after build passes

PR #164, opened 2017-04-10
Adds `.button--reset` class and `<Button reset ...` prop

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-188

#### Description
- added `reset` button variant
- added more `inverted` stories for visual testing/examples
- fixed inverted style for `disabled` buttons

**Inverted:**
screen shot 2017-04-10 at 12 42 28 pm

**Normal:**
screen shot 2017-04-10 at 12 42 24 pm


akdetrick:
Open question...

Should `reset` buttons have bold labels by default? One alternative is to also reset the font weight and expect an extra modifer class to bring back the bold if needed:
`<Button reset className='text--bold' ... />`

mperrotti:
@akdetrick , I've already encountered a case where I need un-bold text for reset buttons - RSVP actions

assets/scss/components/_button.scss
```diff
@@ -257,6 +258,17 @@ Class                   | Description
 	background: $C_contentBG;
 }
 
+.button--reset {
+	@include buttonHover(transparent);
+	background: transparent;
```
mperrotti: `.button[disabled]` has a lot of `!important` going on, including `background`, so this one will get beaten on disabled buttons

mperrotti: Maybe we should file something separately to deal with the aggressive `.button[disabled]` rules

akdetrick: filed a separate ticket for specificity issue:
https://meetup.atlassian.net/browse/WC-42

PR #166, opened 2017-04-10
Add error styling and functionality to datetimepicker

burrata:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-239

#### Description
DateTimePicker currently does not have an error state.

screen shot 2017-04-10 at 3 19 37 pm

### DIFF
--- a/src/DateTimePicker.jsx
+++ b/src/DateTimePicker.jsx
@@ -179,6 +179,9 @@ class DateTimePicker extends React.Component {
 			{ 'field--error': error },
 			className
 		);
+		const childClasses = cx(
+			{ 'field--error': error }
+		);
 		const labelClassNames = cx({required});
 		const timeInputName = `${name}-time`;
 
@@ -194,7 +197,6 @@ class DateTimePicker extends React.Component {
 						required={required}
 						className={classNames}
 						onChangeCallback={this.setDateTime}
-						error={error}
 
 						{...other} />
 					{ error && <p className='text--error'>{error}</p> }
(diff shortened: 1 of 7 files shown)


src/datetimepicker.story.jsx
```diff
@@ -119,4 +119,21 @@ storiesOf('DateTimePicker', module)
...
+		return (<div className='span--50'>
+			<DateTimePicker
+				name='event_time'
+				label='Start at'
+				value={date}
+				error={'Whoops, something went wrong.'}
+				datepickerOptions={opts}
+				forceCalendar
+			/>
+		</div>);
```
mmcgahan: doesn't matter much because it's a storybook story, but this would be easier to read if the JSX was fully nested inside the parens

```jsx
return (
  <div ...
  </div>
);
```

src/DateTimePicker.jsx
```diff
@@ -184,15 +186,19 @@ class DateTimePicker extends React.Component {
...
-					onChangeCallback={this.setDateTime}
+				<div>
+					<DateTimeLocalInput
+						id={id}
+						label={label}
+						value={this.state.datetime}
+						required={required}
+						className={classNames}
+						onChangeCallback={this.setDateTime}
+						error={error}
```
mmcgahan: sort of a minor suggestion, but I think it would be nice if `DateTimeLocalInput`, `CalendarComponent`, and `TimeInput` had APIs that were as close to a 'pure HTML' element as possible, meaning that instead of passing `error` (which is _only_ used to set an error `classname`, you should do the classname calculation in the parent and pass the `field--error` class as a `className` prop.

That setup makes the components 'dumber', and more like direct replacements for `<input>`, which I think makes them easier to understand.

PR #167, opened 2017-04-10
Textarea hotfix

mperrotti:
#### Description
onChange wasn't added into this.props. This caused the `onChange` prop from a Textarea component to not work


src/textarea.test.jsx
```diff
@@ -6,6 +6,10 @@ jest.mock('autosize', () => {
 	return jest.fn();
 });
 
+// const externalOnChange = (e) => {
```
eilinora: looks to have some legacy code :)

src/textarea.test.jsx
```diff
@@ -93,6 +97,25 @@ describe('Textarea', function() {
...
+
+		textareaEl = TestUtils.findRenderedDOMComponentWithTag(boundComponent, 'textarea');
+		TestUtils.Simulate.change(textareaEl, { target: { value: newValue } });
+
+		expect(changeSpy).toHaveBeenCalled();
+		expect(stateSpy).toHaveBeenCalledWith({ value: newValue });
+	});
+	it('should handle onChange as a prop', function() {
+		const newValue = `${VALUE}r`;
+		const changeSpy = spyOn(Textarea.prototype, 'onChange').and.callThrough();
```
eilinora: it looks to be spying on the internal onChange vs the Textarea.props version. You could do something like...
```
		const onChange = jest.fn();
		const boundComponent = TestUtils.renderIntoDocument(
			<Textarea
					onChange={onChange}
					name={NAME_ATTR}
					label={LABEL_TEXT}
					id={ID}
					value={VALUE}
			/>);
			TestUtils.Simulate.change(textareaEl, { target: { value: newValue } });
			expect(onChange).toHaveBeenCalled();
```

src/textarea.test.jsx
```diff
@@ -93,6 +97,25 @@ describe('Textarea', function() {
...
+		const stateSpy = spyOn(Textarea.prototype, 'setState').and.callThrough();
+
+		const boundComponent = TestUtils.renderIntoDocument(<Textarea
```
eilinora: and to be obnoxiously nitpicky, can we move the `<Textarea` to a new line, it is harder to read when it's on the same line with the renderIntoDoc

mperrotti:
It looks like somebody went in and made all the changes already o.O
Thank you phantom test-fixer!

src/textarea.test.jsx
```diff
@@ -101,6 +104,24 @@ describe('Textarea', function() {
...
+				value={VALUE}
+				id={ID}
+				onChange={onChange}
+			/>
+		);
+
+		textareaEl = TestUtils.findRenderedDOMComponentWithTag(boundComponent, 'textarea');
+		TestUtils.Simulate.change(textareaEl, { target: { value: newValue } });
+
+		expect(onChange).toHaveBeenCalled();
```
mmcgahan: When using a simulated user action, it's usually good to check that the spy was _not_ called before the interaction in addition to checking that is _is_ called afterward. Otherwise, code that called the spy on initialization and then ignored the user input would still pass the unit test despite being wrong.

PR #174, opened 2017-04-11
update deploy keys

mattkime:


.gitignore
```diff
@@ -21,3 +21,4 @@ npm-debug.log
 # Generated icon library file
 /src/iconLibrary.story.jsx
 
+github_deploy_key
```
mmcgahan: code comment to explain what this is for would be useful!

PR #179, opened 2017-04-12
travis - no sudo

mattkime:
simplifying the travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 language: node_js
 node_js:
   - "7.7.1"


mattkime:
this looks good to go

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
```
mmcgahan: `sudo: false` is default - simpler to delete?

mattkime: the default is based on when the repo was created so i'd rather be explicit about it.

mmcgahan: ok. keeping outdated config rules seems like an odd preference since the current Travis environment is specifically designed to not need this rule, but I guess it's not a blocker.

PR #189, opened 2017-04-24
Layout component tweaks

mperrotti:
Creating a PR so I can have use the beta release to see what side effects these changes have in our UIs.

#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-244

#### Description
We were seeing issues, especially with Section, when using in mup-web to build Event Home.


mperrotti:
Tested these changes on MUP home, the updated registration flow, and Meetup Pro. Nothing appeared broken, but we may see some unwanted bottom borders since `<Section>` now has a bottom border and bottom padding by default (similar to how `<Section padded>` worked).

Ready for review and merge!

src/layout/section.story.jsx
```diff
@@ -18,47 +20,104 @@ storiesOf('Section', module)
...
+							<div className='chunk'>
+								<h2 className='text--display666'>These sections are padded on all sides</h2>
+							</div>
+						</Section>
+						<Section>
+							<div className='chunk'>
+								<h2 className='text--display666'>These sections are padded on all sides</h2>
+							</div>
+						</Section>
+						<Section> {/* canBeLastChild='large' className='section--notAlwaysLast'*/}
```
akdetrick: These props don't look like they should be commented out.

assets/scss/components/_section.scss
```diff
@@ -11,25 +11,25 @@ $sectionSpaceMultiple: 2;
...
-.section--padded {
-	position: relative;
-	border-top: 0;
-	&:before {
-		content: "";
-		background: $C_border;
-		height: 1px;
-		bottom: 0;
-		position: absolute;
+@include _bpModifier(section, canBeLastChild) {
```
akdetrick: I'm a little confused about this modifier; in responsive layouts, the position of the `section` in the DOM won't change. 

If the position of the element in the DOM does change via the `render` method in a component, then adding or removing the `section--noBorder` should be sufficient to handle adding or removing the section border.

mperrotti: We've actually talked about these styles - there's a lot of simplification I need to include.

akdetrick: Are the props commented out in the story still relevant? Going to take a look in a browser to get a better understanding again.

mperrotti: No, they're not. I'm going with a different interface

PR #194, opened 2017-04-27
WC-45: adding value prop to textarea

eilinora:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-45

#### Description
- adding value prop to text area

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -8,26 +8,57 @@ import autosize from 'autosize';
 class Textarea extends React.Component {
 	constructor(props) {
 		super(props);
-		this.state = {
-			value: props.value || '',
-		};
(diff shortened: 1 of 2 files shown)


src/forms/Textarea.jsx
```diff
@@ -8,29 +8,61 @@ import autosize from 'autosize';
...
+	}
+
+	/**
+	 * @param {Object} nextProps the incoming props
+	 * @return {undefined} side effect only
+	 */
+	componentWillReceiveProps(nextProps) {
+		this.setState((state, props) => ({
+			value: nextProps.value || ''
+		}));
```
mmcgahan: since it's the same as `componentWillMount`, let's make this a new method that both lifecycle methods can call. A name like `overrideValue` will make the intent more clear as well

PR #196, opened 2017-05-01
Sds 262/modal stacking

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-262

#### Description
Add `narrow` and `zIndex` prop support to allow stackable modals and a narrow modal variant.

screen shot 2017-05-01 at 7 08 37 pm

### DIFF
--- a/assets/scss/components/_modal.scss
+++ b/assets/scss/components/_modal.scss
@@ -1,31 +1,4 @@
-$modal-width: 440px;
-
-@mixin modal-dialog {
-	@include transform(translate3d(0, 0, 0));
-	@include transition(transform .2s ease);
-	@include centerAbsolute($modal-width);
-	border-radius: $defaultRadius;
-	min-height: 0;
-	width: $modal-width !important;
-
-	&.off{
-	  @include transform( translateY(100vh) );
-	}
-}
-@mixin modal-full {
-	@include transform( translate3d(0, 0, 0) );
-	@include transition(transform .3s);
-	left: 0;
-	margin: 0;
-	min-height: 100vh;
-	position: absolute;
-	top: 0;
-	width: 100%;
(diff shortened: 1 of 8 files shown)


mmcgahan:
Stackable modals sounds like providing technical support to a bad UX state - is it possible to force better UX handling in cases that would otherwise produce a stacked modal?

akdetrick:
I agree that we want to generally avoid this, but there are some modals from which we need to prompt for user action. Maybe I can remove the stacking example from storybook but leave the `zIndex` support for something the foundation team is launching...

mmcgahan:
> but there are some modals from which we need to prompt for user action

Right, a modal on top of a modal in order to prompt for user action is bad UX. Let's really not do that. Inline error messages are fine inside a constrained panel like a modal. Blocking UI on top of blocking UI indicates that there is something wrong with the design.

mperrotti:
Also really not into having a modal within a modal. I think there's more design exploration that could/should be done before we commit to this.

akdetrick:
The "foundation team" is the new team name for the part of the engagement team working on things in the new platform.

The changes in this PR would give us the flexibility to work around the fact that the event attendees content was put into a modal. I agree though, a double modal is an indication that we put content into a modal that should have been a full page. 

Marking with `Prelim` to see if we can make a product change instead...

akdetrick:
Removing `Prelim` tag. The foundation team will revisit this to make the attendees modal its own page, which will avoid modals on top of modals in the future.

Going to remove the storybook story for double modals because it's not a pattern we want to encourage, but the flexibility to use a narrow modal or to set a custom z-index isn't a bad thing to have.

akdetrick:
closing, we don't want to do this

PR #197, opened 2017-05-04
Fixed typo and test for Section

mperrotti:

### DIFF
--- a/src/layout/section.test.jsx
+++ b/src/layout/section.test.jsx
@@ -28,11 +28,13 @@ describe('Section', function() {
 	});
 
 	describe('Section noBorder', () => {
-		Object.keys(VALID_BREAKPOINTS).forEach(breakpoint => {
-			section = TestUtils.renderIntoDocument(<Section noBorder={breakpoint} />);
-			sectionNode = ReactDOM.findDOMNode(section);
-			expect(sectionNode.classList).toContain(SECTION_CLASS);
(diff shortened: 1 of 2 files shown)


src/layout/section.test.jsx
```diff
@@ -28,11 +28,13 @@ describe('Section', function() {
...
-			section = TestUtils.renderIntoDocument(<Section noBorder={breakpoint} />);
-			sectionNode = ReactDOM.findDOMNode(section);
-			expect(sectionNode.classList).toContain(SECTION_CLASS);
-			expect(sectionNode.classList).toContain(`${VALID_BREAKPOINTS[breakpoint]}_${SECTION_NOBORDER_CLASS}`);
+		it(`check that default component has '${SECTION_NOBORDER_CLASS}' class`, function() {
+			Object.keys(VALID_BREAKPOINTS).forEach(breakpoint => {
+				section = TestUtils.renderIntoDocument(<Section noBorder={breakpoint} />);
+				sectionNode = ReactDOM.findDOMNode(section);
+				expect(sectionNode.classList).toContain(SECTION_CLASS);
+				expect(sectionNode.classList).toContain(`${VALID_BREAKPOINTS[breakpoint]}_${SECTION_NOBORDER_CLASS}`);
```
akdetrick: it might be better to import this className from `Section.jsx` instead of constructing it in the test

mperrotti: This isn't just one class name, it's 3. Following the pattern from the Flex tests

akdetrick: in that case exporting the pattern would make this work... but this pattern isn't likely to change. Merge away!

PR #203, opened 2017-05-11
Wc 49/match media hoc utility

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-49

#### Description
Adds higher order component that provides boolean `isAt[breakpoint]Up` props to wrapped component. 

screen shot 2017-05-11 at 4 28 08 pm
screen shot 2017-05-11 at 4 28 15 pm
screen shot 2017-05-11 at 4 28 27 pm

### DIFF
--- a/src/utils/components/withMatchMedia.jsx
+++ b/src/utils/components/withMatchMedia.jsx
@@ -1,11 +1,11 @@
 import React from 'react';
-import { MEDIA_QUERIES } from '../utils/designConstants';
+import { MEDIA_QUERIES } from '../designConstants';
 
 /**
- * @param {String} breakpoint - name of breakpoint
- * @returns {String} - prop name (for example, isAtMediumUp)
+ * @param {String} breakpoint - name of breakpoint (for example, 'medium')
+ * @returns {String} - state property name (for example, isAtMediumUp)
  */
-export const createPropNameFromBreakpoint = (breakpoint) => {
+export const getStateNameByBreakpoint = (breakpoint) => {
 	const capitalizedBp = breakpoint[0].toUpperCase() + breakpoint.slice(1);
 	return `isAt${capitalizedBp}Up`;
 };
@@ -17,6 +17,10 @@ export const createPropNameFromBreakpoint = (breakpoint) => {
  */
 export const validateBreakpoints = (breakpoints) => {
 	const validBreakpoints = Object.keys(MEDIA_QUERIES);
+
+	if (!breakpoints) {
+		throw new Error('withMatchMedia: breakpoints array required');
+	}
 
 	breakpoints.forEach(bp => {
 		if (!validBreakpoints.includes(bp)) {
@@ -34,9 +38,7 @@ export const validateBreakpoints = (breakpoints) => {
  */
 export const withMatchMedia = (
 	InnerComponent,
-	breakpoints = () => {
-		throw new Error('withMatchMedia: breakpoints array required');
-	}
+	breakpoints
 ) => class extends React.Component {
 	/**
 	 * @constructor
@@ -44,44 +46,83 @@ export const withMatchMedia = (
 	 */
 	constructor(props) {
 		super(props);
-		validateBreakpoints(...breakpoints);
+		validateBreakpoints(breakpoints);
+
+		this.handleMediaChange = this.handleMediaChange.bind(this);
+
+		this.state = {
+			media: {}
+		};
 
 		breakpoints.forEach(bp => {
-			this.state[createPropNameFromBreakpoint(bp)] = false;
+			this.state.media[getStateNameByBreakpoint(bp)] = false;
+		});
+	}
+
+	/**
+	 * handles media change for _all_ breakpoints configured
+	 * updates state.media with matched media
+	 * @returns {undefined}
+	 */
+	handleMediaChange() {
+		const updatedMedia = {};
+		this.mediaQueries.forEach((mq, i) => {
+			updatedMedia[getStateNameByBreakpoint(breakpoints[i])] = mq.matches;
+		});
+		this.setState({
+			media: {...this.state.media, ...updatedMedia}
 		});
 	}
 
 	/**
 	 * react lifecycle method
+	 * sets up media queries and listeners
 	 * @returns {undefined}
 	 */
 	componentDidMount() {
-		// TODO: add media change callbacks for custom hanlders
-		// TODO: figure out if shit should be in state
-
 		if (typeof window.matchMedia != undefined) {
-			// let mediaQueries = [];
 
-			/*
-			 *breakpoints.forEach(bp => {
-			 *    mediaQueries.push(
-			 *        window.matchMedia(MEDIA_QUERIES[bp]);
-			 *    );
-			 *});
-			 */
+			this.mediaQueries = breakpoints
+				.map(bp => window.matchMedia(MEDIA_QUERIES[bp]));
+
+			this.mediaListeners = this.mediaQueries
+				.map(mq => mq.addListener(this.handleMediaChange));
+
+			// fire media handler on mount to populate `this.state.media`
+			this.handleMediaChange();
 		}
 	}
 
 	/**
 	 * react lifecycle method
+	 * clean up media query listeners on unmount
 	 * @returns {undefined}
 	 */
 	componentWillUnmount() {
+		this.mediaQueries.forEach(mq => {
+			mq.removeListener(this.handleMediaChange);
+		});
 	}
 
+	/**
+	 * `isAt[Breakpoint]Up` props are provided to the wrapped component in render().
+	 *
+	 * When `this.mediaListeners` detect a media change, `this.state.media` is updated
+	 * and the component will re-render with correct media-conditional prop values.
+	 *
+	 * @returns {React.element}
+	 */
 	render() {
+		const mediaProps = {};
+
+		breakpoints.forEach(bp => {
+			const propName = getStateNameByBreakpoint(bp);
+			mediaProps[propName] = this.state.media[propName];
+		});
+
 		return (
 			<InnerComponent
+				{...mediaProps}
 				{...this.props}
 			/>
 		);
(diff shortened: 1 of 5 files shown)


akdetrick:
I'd like to add tests that verify that the HOC `componentDidMount` and `componentWillUnmount` lifecycle methods add and remove matchMedia listeners, but the HOC will always return the wrapped component... Not sure how to go about testing this

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * react lifecycle method
+	 * clean up media query listeners on unmount
+	 * @returns {undefined}
+	 */
+	componentWillUnmount() {
+		this.mediaQueries.forEach((mq, i) => {
+			mq.removeListener(this.handleMediaChange);
+			this.mediaQueries.splice(i, 1);
```
akdetrick: I'm removing created `MediaQueryList` objects from this array here for the sake of testing, but have had some difficulty actually testing lifecycle methods of `withMatchMedia` itself

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,143 @@
...
+		const mediaProps = {};
+
+		breakpoints.forEach(bp => {
+			const propName = getStateNameByBreakpoint(bp);
+			mediaProps[propName] = this.state.media[propName];
+		});
+
+		return (
+			<InnerComponent
+				withMatchMediaInstance={this}
```
akdetrick: This approach seems wrong, but I can't seem to figure out any other way to directly call and test instance methods on `withMatchMedia` as it always returns the wrapped component.

src/utils/components/withMatchMedia.test.jsx
```diff
@@ -0,0 +1,87 @@
...
+ * @param {Array} breakpoints - breakpoint names to pass to withMatchMedia
+ */
+function renderWrappedComponent(breakpoints) {
+	const renderer = TestUtils.createRenderer();
+	const TestComponentWithMatchMedia = MM.withMatchMedia(
+		TestComponent,
+		breakpoints
+	);
+	renderer.render(<TestComponentWithMatchMedia />);
+	return renderer.getRenderOutput();
```
mmcgahan: instead of shallow rendering, try using:
```jsx
TestUtils.renderIntoDocument(<TestComponentWithMatchMedia />)
```

I'll have to look a bit more into how to use Shallow Rendering to get the root level component

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,143 @@
...
+	return updatedMedia;
+};
+
+/**
+ * @param {Array} breakpoints
+ * @throws {Error}
+ * @return {undefined}
+ */
+export const validateBreakpoints = (breakpoints) => {
+	const validBreakpoints = Object.keys(MEDIA_QUERIES);
```
mmcgahan: no need to redefine `validBreakpoints` every time the function runs - make it a module-scoped `VALID_BREAKPOINTS` constant, or curry the function:

```js
// look ma, pure functions!
const _getValidateBreakpoints = mediaQueries => {
  const validBreakpoints = Object.keys(mediaQueries);
  return breakpoints => {
    ...
  }
};

const validateBreakpoints = _getValidateBreakpoints(MEDIA_QUERIES);

File: src/utils/components/withMatchMedia.jsx [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+		});
+	}
+
+	/**
+	 * react lifecycle method
+	 * sets up media queries and listeners
+	 * @returns {undefined}
+	 */
+	componentDidMount() {
+		if (typeof window.matchMedia != undefined) {
```
mmcgahan: prefer the 'exit early' pattern rather than a guard around the whole body of a function - makes it easier to read and understand intent.

```js
if (!window.matchMedia) {
  return;
}
// left-aligned 'happy path'
```

src/utils/components/withMatchMedia.jsx
akdetrick: this should definitely be a pure function

akdetrick:
#### Tests
I've made an improvement that also makes things easier to test (moved state update logic to a pure function). 

#### Performance concerns
Calling `window.matchMedia` will do two things:
* serialize the passed media query as a normal CSS media query
* return a `MediaQueryList` object, that exposes the media query "matches" prop to javascript

I don't expect any more than about 5 of our standard components in `meetup-web-components` to use `withMatchMedia`. If each one utilizes every breakpoint available, we're only up to about 20 media queries.

I'm also recommending that feature containers be the components that we wrap with `withMatchMedia`. That adds 4 more media queries to a given page.

That means we're likely to see about 20-30 extra media queries on a given page (with js listeners). We have _many_ more than that in sasstools, so I'm not very concerned. As a precaution, wrapped components will clean up their listeners on unmount.

If performance becomes an issue, we can revisit this to allow all instances of `withMatchMedia` to share the same `MediaQueryList` objects.

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import { MEDIA_QUERIES } from '../designConstants';
+
+/**
+ * @param {String} breakpoint - name of breakpoint (for example, 'medium')
+ * @returns {String} - state property name (for example, isAtMediumUp)
+ */
+export const getStateNameByBreakpoint = breakpoint => {
+	const capitalizedBp = breakpoint[0].toUpperCase() + breakpoint.slice(1);
```
mmcgahan: prefer a string template instead of `+` string concatenation, and prefer string methods instead of array methods to manipulate strings (just because they're a bit more explicit about the data type)

```js
return `isAt${breakpoint.substr(1, 0).toUpperCase()}{breakpoint.substr(1)}`
// or if you rename the arg to be shorter
return `isAt${bp.substr(1, 0).toUpperCase()}{bp.substr(1)}`;
```

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * @param {Array} mediaQueries - list of matchMedia-created MediaQueryList objects
+ * @param {Array} breakpoints - array of breakpoint names that were passed to HOC
+ */
+export const getUpdatedMediaState = (mediaQueries, breakpoints) => {
+	const updatedMedia = {};
+	mediaQueries.forEach((mq, i) => {
+		updatedMedia[getStateNameByBreakpoint(breakpoints[i])] = mq.matches;
+	});
+	return updatedMedia;
```
mmcgahan: building an object from an array is a classic `.reduce` application:

```js
return mediaQueries.reduce((state, mq, i) => {
  state[getStateNameByBreakpoint(breakpoints[i])] = mq.matches;
  return state;
}, {});
```

src/utils/components/withMatchMedia.jsx
akdetrick: word 👍

mmcgahan:
Performance discussion sounds good to me, thanks for the overview

akdetrick:
bump; any blockers?

akdetrick:
Simplified the way we handle breakpoints. `withMatchMedia` will now listen for all named media queries and pass `state.media` to a `media` prop in the wrapped component.

akdetrick:
Creating a shared list of `MediaQueryList` objects between instances of `withMatchMedia` probably isn't necessary. It's always possible to add this later in a backward-compatible way if we find we need or want to do so.

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * @constructor
+	 * @param {Object} - React element props
+	 */
+	constructor(props) {
+		super(props);
+
+		// map breakpoint names to MediaQueryList objects...
+		this.mediaQueries = breakpointNames
+			.map(bp => window.matchMedia(MEDIA_QUERIES[bp]));
```
mmcgahan: this won't work on server where `window` is undefined - you'll need to move it into `componentDidMount`

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+		this.state = {
+			media: getUpdatedMediaState(this.mediaQueries, breakpointNames)
```
mmcgahan: since `breakpointNames` is set as a module constant, you don't need to pass it as an argument. Also, since you can't set `this.mediaQueries` until `componentDidMount`, you might as well just call `this.handleMediaChange()` directly instead of duplicating the state-setting logic.

akdetrick:
So I could be wrong about this, but instead of moving `this.mediaQueries` assignment to `componentDidMount`, I added a check in the constructor. I'd like to do it this way so that `state.media` is set at component instantiation rather than forcing a state update and re-render on mount.

If you think this is too weird to be worth setting default state at instantiation, I can move `this.mediaQueries` assignment back to `componentDidMount`.

src/utils/components/withMatchMedia.jsx
```diff
@@ -0,0 +1,111 @@
...
+) => class extends React.Component {
+	/**
+	 * @constructor
+	 * @param {Object} - React element props
+	 */
+	constructor(props) {
+		super(props);
+
+		// map breakpoint names to MediaQueryList objects...
+		if (window && window.matchMedia) {
```
mmcgahan: this will still throw an error on the server because it tries to 'evaluate' `window`, which will throw a `ReferenceError` in Node.

The preferred idiom is `if (typeof window !== 'undefined' && window.matchMedia)`

mmcgahan: This state assignment really needs to go into `componentDidMount` even though it will trigger a re-render. The unavoidable problem is that React expects the initially-generated browser markup to exactly match the server-generated markup, and if the `media` prop is used to affect markup in the browser but not the server, then there will be (potential) problems. Apparently they are going to change this behavior in the next major version of React, but I think we can pay the (small) cost of a re-render for now and put it in the 'correct' place in `componentDidMount`

akdetrick:
Makes sense; moved matchMedia calls back into `componentDidMount`

PR #204, opened 2017-05-11
Reduce spacing in icon buttons

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-286
Will help with: [MW-1023](https://meetup.atlassian.net/browse/MW-1023) and [MW-1027](https://meetup.atlassian.net/browse/MW-1027)

#### Description
Just reducing padding coming from `.flex-item`. The CSS selectors to do this without making changes to `Button` aren't ideal

### DIFF
--- a/assets/scss/components/_button.scss
+++ b/assets/scss/components/_button.scss
@@ -370,6 +370,10 @@ $_iconGutter: $space;
 	@include flex-wrap(nowrap);
 	@include flex-direction(row);
 	@include justify-content(space-between);
+
+	.atAll_flex--rowReverse.button--icon-wrapper & {
+		padding-left: $space/4;
+	}
 }
 
 .button-label,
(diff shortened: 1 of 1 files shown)


assets/scss/components/_button.scss
```diff
@@ -370,6 +370,10 @@ $_iconGutter: $space;
 	@include flex-wrap(nowrap);
 	@include flex-direction(row);
 	@include justify-content(space-between);
+
+	.atAll_flex--rowReverse.button--icon-wrapper & {
```
akdetrick: what about `at[specificBreakpoint]_flex--rowReverse` classes?

akdetrick: kind of an edge case, now that i think about it

mperrotti: Then the `:first-child` beats it

PR #209, opened 2017-05-17
set up gh-pages deployment

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-53

#### Description
Adds deploy script for `gh-pages` deployment to github pages.


akdetrick:
Where should I be looking for Travis configuration in order to trigger the `deploy` script on merges to master? Or is Travis the wrong tool for this job?

akdetrick:
Prelim for now; investigating if we can run the deploy script on merges to master

akdetrick:
**bump**

Change introduced to travis config in this branch will trigger gh-pages deployment on merges to master.

package.json
```diff
@@ -17,6 +17,9 @@
     "postmerge": "yarn install",
     "precommit": "git status --porcelain | awk 'match($1, \"[ACM]\"){print $2}' | grep '[js|jsx]$' | xargs eslint",
     "storybook": "start-storybook -p 9001 -c .storybook",
+    "storybook-static": "build-storybook -c .storybook -o .out",
```
mmcgahan: looks like this could be named anything, so maybe `.storybook-out` is better than `.out`?

mmcgahan: is `build-storybook` part of our existing dependencies, or do we need to add https://github.com/storybooks/storybook-deployer as a dev dependency?

package.json
```diff
@@ -17,6 +17,9 @@
...
+    "gh-pages": "gh-pages -d .out -b gh-pages",
+    "deploy": "yarn run storybook-static && yarn run gh-pages",
```
mmcgahan: `deploy` is a little ambiguous for this repo, because it's not like the rest of Meetup's deployments - maybe `publish` or something?

package.json
akdetrick: `build-storybook` is built-in with the version of storybook we're running

mattkime: I was going to say - we'd have a yarn check failure if it let the build continue without a required node module

mmcgahan: @mattkime `yarn check` only ensures that `yarn.lock` lines up with `package.json` - if a required node module isn't installed at all (e.g. if it was installed globally on a dev machine and not added as a local dependency), then `yarn check` would pass, but this feature would fail when merging to master.

.travis.yml
```diff
@@ -46,6 +46,9 @@ after_script:
     npm version $GIT_TAG -m "Version $GIT_TAG built by Travis CI - https://travis-ci.com/$TRAVIS_REPO_SLUG/builds/$TRAVIS_JOB_ID" &&
     git push --tags git@github.com:$TRAVIS_REPO_SLUG.git &&
     npm publish --tag $NPM_TAG
+  - >
+    [ "${TRAVIS_BRANCH}" = "master" ] && [ "${TRAVIS_PULL_REQUEST}" = "false" ] &&
+    yarn run publish
```
mmcgahan: actually i just realized that `publish` probably isn't a good name either because it's ambiguous with a `npm publish` :/ sorry. `publish-storybook`?

akdetrick: agreed

package.json
mattkime: in this situation the build would have failed when it tried to call this task.

PR #212, opened 2017-05-23
Stripe - `hero` cleanup

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-253

#### Description
Hides the text protection scrim when no `backgroundImage` is passed in. Removes support for `hero` prop in the `Stripe` component


mperrotti:
Doesn't break anything in `mup-web`, and `hero` doesn't even get used in pro-web. Review at will

assets/scss/components/_forms.scss
```diff
@@ -0,0 +1,194 @@
...
+---
+Faux inputs visually mimic actual input elements.
+
+```html_example
+<div class="chunk">
+	<div class="fauxInput">I'm a faux input</div>
+</div>
+```
+*/
+.fauxInput {
```
akdetrick: can we move `fauxInput` to the partial for the component in which it's used?

assets/scss/components/_forms.scss
```diff
@@ -0,0 +1,194 @@
...
+	<input id="check2" class="checkbox" value="blue" type="checkbox" />
+	Blue
+</label>
+<label for="check3" class="label--minor chunk">
+	<input id="check3" class="checkbox" value="blue" type="checkbox" checked />
+	Already checked
+</label>
+```
+*/
+.checkbox {
```
akdetrick: let's move this to a `_checkbox` partial to match the component.

src/layout/Stripe.jsx
```diff
@@ -44,11 +44,11 @@ class Stripe extends React.Component {
 				className={classNames}
 				style={styles}
 				{...other}>
-				{ hero ?
+				{ backgroundImage ?
 					(
-						<Bounds className='stripe-heroContent'>
+						<div className='stripe-aboveScrim'>
```
akdetrick: do we need to rethink the `heroContent` prop for modals with this change?

mperrotti: I don't think so. "Hero" is a pretty understood term, I didn't even intentionally name it the same thing as the `<Stripe>` class.
If you don't think this is intuitive, I'm happy to rethink

assets/scss/components/_forms.scss
mperrotti: `fauxInput` is only used on `NumberInput` right now, but it could be used for similar things in the future

mperrotti:
bump

src/layout/Stripe.jsx
```diff
@@ -21,7 +21,7 @@ class Stripe extends React.Component {
 			backgroundImage,
 			collection,
 			inverted,
-			hero,
+			// hero,
```
mmcgahan: remove commented line?

PR #211, opened 2017-05-22
new type scale in MWC

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-282

#### Description
Using new type scale for our components

#### Screenshots (if applicable)
typescale

PR #213, opened 2017-05-23
Forms visual updates

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-292

#### Description
*Please wait until issue #211 is reviewed and merged before merging this*
Updating form styles to match latest visual design for form inputs. Moving some styles from `swarm-sasstools`

screen shot 2017-05-23 at 10 36 52 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,7 @@ lib
 # Generated icon library file
 /src/media/iconLibrary.story.jsx
 
+# static build of storybook
+.storybook-out
+
 github_deploy_key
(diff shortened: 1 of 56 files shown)


mperrotti:
Found an issue with labels:
Now that they're `display: inline-block`, they push down inputs if they don't have text in them.

I'm thinking labels should be required, and we can add a prop to hide them *visually*, taking them out of document flow

screen shot 2017-05-25 at 9 19 17 am

akdetrick:
Regarding empty labels, we should consider adding to this branch a fix at the component level. If we're not requiring a `label` prop, we should only render the element if the `label` prop has been passed in: 
https://github.com/meetup/meetup-web-components/blob/master/src/forms/Textarea.jsx#L111

`{label && labelStuff}`

PR #216, opened 2017-05-24
proptypes from prop-types

mmcgahan:
`React.PropTypes` is deprecated. Long live `PropTypes` from `prop-types`

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes

Doesn't matter who does the review, but @akdetrick and @mperrotti should know about the change


akdetrick:
Build seems to be failing on a version issue in `yarn.lock`, but otherwise lgtm

mperrotti:
Looks like a pretty straightforward batch of changes. lgtm 👍

PR #217, opened 2017-05-25
Versatile toggle component

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-270
Fixes https://meetup.atlassian.net/browse/SDS-221

#### Description
High order component that passes down a boolean prop. Makes it easier to make new toggling UI components (e.g.: "like" button, RSVP toggle on event cards) without having to clobber `TogglePill` styling. Shout out to [this Gist](https://gist.github.com/heldr/4bf3bad3ab7a0e4a3de52d38e44fc0b4) for getting me unstuck.

Also takes care of [visual updates](https://meetup.atlassian.net/browse/SDS-270) made by the design team


src/forms/TogglePill.jsx
```diff
@@ -1,26 +1,25 @@
 import React from 'react';
 import cx from 'classnames';
 import Icon from '../media/Icon';
+import withToggleControl from '../utils/WithToggleControl';
 
 export const TOGGLE_PILL_CLASS = 'toggleButton';
 
 /**
  * SQ2 Toggle Pill component
```
akdetrick: nitpick; we should remove references to SQ2

src/utils/WithToggleControl.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+			this.state = { bool: props[propName] };
+			this.onToggleBool = this.toggleBool.bind(this);
+		}
+		toggleBool() {
+			this.setState({ bool: !this.state.bool });
+		}
+		render() {
+
+			const newProps = Object.assign({}, this.props, {
```
akdetrick: We favor using an es6 spread over `Object.assign`....

```
const newProps = { ...this.props, [`${propName}`]: this.state.bool, toggle: this.onToggleBool }
```

src/utils/withToggleControl.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+			isChecked,
+			toggle
+		} = this.props;
+
+		return <h1 className={WRAPPED_COMPONENT_CLASS} onClick={toggle} aria-pressed={isChecked}>Is this component checked?</h1>;
+	}
+}
+
+/**
+ * @class TestComponent
```
akdetrick: class name in doc

src/utils/withToggleControl.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+	it('provides `isChecked` props to wrapped component', () => {
+		const actualPropNames = Object.keys(wrappedComponent.props);
+		const expectedPropName = 'isChecked';
+
+		expect(actualPropNames).toContain(expectedPropName);
+	});
+
+	it('provides custom prop name for bool to wrapped component when one is specified', () => {
+		const TestComponentWithToggleControl = withToggleControl(TestComponentCustomProp, 'on');
```
akdetrick: very minor, but it would be better to define the `epxectedPropName` const first, and pass it to `withToggleControl` instead of using a string literal here

src/forms/TogglePill.jsx
```diff
@@ -1,26 +1,25 @@
...
 		super(props);
 
-		this.state = {isChecked: props.checked || false};
-
 		this.onChange = this.onChange.bind(this);
 	}
 
 	onChange(e) {
-		this.setState({isChecked: !this.state.isChecked});
+		this.props.toggle();
```
akdetrick: Seeing the `toggle` prop in use here makes me want to suggest a clearer name, like `toggleActive`. Because `withToggleControl` doesn't know what prop name will be defined, I know this makes it more difficult to come up with a clear name here. 

Supporting a custom "checked" prop name is a cool feature, but I'm now wondering what is gained. If we have a number of components using `withToggleControl`, it may be a disadvantage if each component has a different prop name to deal with the same behavior. 

Forcing the name of the prop to always be the same might make it easier for developers to understand what's going on under the hood (and to remember the interface for any components using `withToggleControl`. Because not all toggles will use checkboxes, mabye:

- `isActive` for the prop name
- `state.isActive` instead of `state.bool`
- `toggleActive` instead of `toggle`

Setting a checkbox `checked` attribute with `this.props.isActive` instead of `this.props.isChecked` seems like a small price to pay for a more consistent interface

src/utils/WithToggleControl.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+/**
+ * Provides boolean isChecked prop to wrapped component.
+ * Prop name can also be customized if need be
+ *
+ * @param {React.element} ToggleComponent - the component to wrap
+ * @returns {React.element}
+ */
+export default function withToggleControl(WrappedComponent, propName = 'isChecked') {
```
akdetrick: See my comment here about simplifying this with a consistent prop name: https://github.com/meetup/meetup-web-components/pull/217/files?diff=unified#r118526489

mperrotti:
bump

src/utils/WithToggleControl.jsx
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * @module WithToggle
+	 */
+	class WithToggle extends React.Component {
+		constructor(props) {
+			super(props);
+
+			this.state = { isActive: props['isActive'] };
```
mmcgahan: I'm sort of surprised that `Prettier` or ESLint didn't catch this, but style preference is `props.isActive`

src/utils/WithToggleControl.jsx
```diff
@@ -0,0 +1,49 @@
...
+		toggleBool() {
+			this.setState({ isActive: !this.state.isActive });
+		}
+		render() {
+
+			const newProps = {
+				...this.props,
+				isActive: this.state.isActive,
+				toggleActive: this.onToggleBool
+			};
```
mmcgahan: very minor, but I think in general it would be preferable to avoid creating a new object here and instead pass the two additional props to the `<WrappedComponent>` directly

```jsx
<WrappedComponent {...this.props} isActive={this.state.isActive} toggleActive={this.onToggleBool} />
```

src/utils/WithToggleControl.jsx
```diff
@@ -0,0 +1,49 @@
...
+		}
+		toggleBool() {
+			this.setState({ isActive: !this.state.isActive });
+		}
+		render() {
+
+			const newProps = {
+				...this.props,
+				isActive: this.state.isActive,
+				toggleActive: this.onToggleBool
```
mmcgahan: the renaming here seems awkward - why not `toggleActive: this.toggleActive` or `onToggleBool: this.onToggleBool`?

src/utils/withToggleControl.story.jsx
```diff
@@ -0,0 +1,40 @@
...
+		const toggleText = isActive ? 'The toggle is on' : 'The toggle is off';
+
+		return (
+			<div
+				className='display--block align--center'
+				onClick={toggleActive}
+			>
+				<Icon
+					shape={iconShape}
+					size='m' />
```
mmcgahan: nitpick: might be a few more line breaks than necessary here :)

```jsx
<div className='display--block align--center' onClick={toggleActive}>
  <Icon shape={iconShape} size='m' />
  <p>{toggleText}</p>
</div>
```

src/utils/WithToggleControl.jsx
mmcgahan: I think that comment got buried, but maybe it's already taken care of? It looks like `propName` isn't used any more

akdetrick: Looks like it isn't used. Let's remove the argument

mperrotti:
Thanks for the great feedback. Fixed 👍

PR #226, opened 2017-06-07
Refactor/generic dropdown component

akdetrick:
Added a generic dropdown component that can accept arbitrary JSX content (for use in the `mup-web` header). We plan on refactoring `Popover` to use this component to eliminate some duplicated logic, but that will come in a later iteration.

screen shot 2017-06-07 at 3 10 10 pm
screen shot 2017-06-07 at 3 10 24 pm

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -104,14 +104,10 @@ class Dropdown extends React.Component {
 	}
 }
 
-Dropdown.defaultProps = {
-	align: 'right'
-};
-
 Dropdown.propTypes = {
 	trigger: PropTypes.element.isRequired,
 	content: PropTypes.element.isRequired,
-	align: PropTypes.oneOf(['left', 'right']),
+	align: PropTypes.oneOf(['left', 'right']).isRequired,
 	className: PropTypes.string,
 };
 
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -1,0 +1,183 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import cx from 'classnames';
+import bindAll from '../utils/bindAll';
+
+const POPOVER_MENU_CLASS = 'popover-menu-option';
+/**
+ * @module Popover
+ */
+class Popover extends React.Component {
+	constructor(props) {
+		super(props);
+
+		bindAll(this,
+			'open',
+			'close',
+			'onClick',
+			'onKeyDown',
+			'onClick',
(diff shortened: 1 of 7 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -0,0 +1,114 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import cx from 'classnames';
+import bindAll from '../utils/bindAll';
+
+/**
+ * @module Dropdown
+ */
+class Dropdown extends React.Component {
```
eilinora: Let's make this a `PureComponent` Most SDS components _should_ be `PureComponents` vs `Components`

*General philosophy on Stateless/PureComponent/Component:*
Stateless component -- these are the component that don’t need life-cycle methods or state. These should can be used in rendering repeat element of parent component such as rendering the text list which just displays the information and doesn't have any actions to perform. Caveats, cannot contain stateful component children.

Pure component -- these are the items which have life-cycle and does a baked in shallow comparison of props in shouldComponentUpdate. Use this for components that have simple prop structures and you expect don’t need to be complex components.

Normal Components or Complex Components -- these are complex components or general components the compose many children components and are connected to redux state. These are usually the page level components and consists lot of children components which read it’s data from this complex wrapper.

src/interactive/Dropdown.jsx
```diff
@@ -0,0 +1,114 @@
...
+					onClick={this.onClick}
+				>
+					{trigger}
+				</div>
+
+				{/* TODO: fix aria attributes */}
+				<div
+					className={classNames.content}
+					aria-hidden={!isActive}
+					ref={(el) => this.contentEl}
```
eilinora: don't see this being used might be good to 💀

src/interactive/Dropdown.jsx
akdetrick: We'll need the `componentDidMount` and `componentWillUnmount` lifecycle methods here to handle dropdown dismissal when the user clicks outside the dropdown content. I'll mark this with `Prelim` until that change is made.

src/interactive/dropdown.test.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import TestUtils from 'react-addons-test-utils';
+import Section from '../layout/Section';
+import Chunk from '../layout/Chunk';
+import Button from '../forms/Button';
+
+import Dropdown from './Dropdown';
+
+const dropdownContent = (
+	<Section noSeparator>
```
mperrotti: Nitpick: use `<Section className='border--none'>` instead - it gives you the correct spacing at the bottom.

PR #228, opened 2017-06-08
SV: updating sass icons refs

eilinora:
Updating swarm-icons libs


eilinora:
Adding reviewer for reference but will merge once tests pass

PR #230, opened 2017-06-12
Number Input - Minor fixes

nlaz:
#### Description
This PR fixes some issues with the `NumberInput` component. We don't want to support `e` or `E` input values, we want to trigger `onChange` whenever the plus or minus signs get triggered, and we want to display the default value prop.

#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-1153

### DIFF
--- a/src/forms/NumberInput.jsx
+++ b/src/forms/NumberInput.jsx
@@ -42,6 +42,12 @@ class NumberInput extends React.Component {
 		return newValue;
 	}
 
+	componentWillUpdate(nextProps, nextState) {
+		if (this.props.onChange && nextState.value !== this.state.value) {
+			this.props.onChange({ target: { name: nextProps.name, value: nextState.value }});
+		}
+	}
+
 	onBlur(e) {
 		const formControls = [
 			this.fauxInputEl,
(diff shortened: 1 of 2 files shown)


src/forms/NumberInput.jsx
```diff
@@ -16,16 +16,15 @@ export const INCREMENT_BTN_CLASS = 'incrementButton';
 class NumberInput extends React.Component {
 	constructor(props) {
 		super(props);
-		this.state = {
-			value: props.value || '',
```
nlaz: This was setting value to '' whenever `props.value` was 0 because of how javascript treats falsey values.

src/forms/numberInput.test.jsx
```diff
@@ -287,7 +287,7 @@ describe('NumberInput', function() {
 			);
 
 			expect(numberInputComponent.state.value).toEqual(VALUE);
-			numberInputComponent.decrementAction();
+			numberInputComponent.decrementAction({ preventDefault: () => {} });
 			expect(numberInputComponent.state.value).toEqual(newValue);
 		});
 	});
```
akdetrick: Should we add a test to make sure the component disallows `E`/`e` in the value?

PR #232, opened 2017-06-13
support placeholder prop

akdetrick:
Fixes issue with unsupported `placeholder` prop on `TextInput`.

Also improves storybook stories with `InfoWrapper`.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -94,10 +94,7 @@ TextInput.propTypes = {
 		PropTypes.string,
 		PropTypes.element
 	]),
-	placeholder: PropTypes.oneOfType([
-		PropTypes.string,
-		PropTypes.element
-	]),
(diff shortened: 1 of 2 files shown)


src/forms/TextInput.jsx
```diff
@@ -92,6 +94,10 @@ TextInput.propTypes = {
 		PropTypes.string,
 		PropTypes.element
 	]),
+	placeholder: PropTypes.oneOfType([
+		PropTypes.string,
+		PropTypes.element
```
mmcgahan: since placeholder is an html element attribute, it probably needs to be a real string end to end... hopefully not too difficult using `withIntl(Component, messages, true)` + `intl.formatMessage`

akdetrick: good call, will change this to accept only `string`

akdetrick:
Invalid now; closing

PR #235, opened 2017-06-19
Textarea - Update autosized height on changed values

nlaz:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-1356

#### Description
This PR fixes a bug where the textarea doesn't resize on form submission. This is because the autosize library only listens to onChange events and not value changes.

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -50,6 +50,7 @@ class Textarea extends React.Component {
 	 */
 	componentWillReceiveProps(nextProps) {
 		this.overrideValue(nextProps);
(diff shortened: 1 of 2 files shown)


src/forms/Textarea.jsx
```diff
@@ -150,7 +151,6 @@ Textarea.propTypes = {
 	]),
 	labelClassName: PropTypes.string,
 	required: PropTypes.bool,
-	autoHeight: PropTypes.bool,
```
nlaz: Thought I would remove this prop because I don't see it used anywhere. Let me know if I was wrong here.

PR #236, opened 2017-06-19
set flex basis for IE to auto

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-63

#### Description
Changes `-ms-preferred-size` from `0px` to `auto` for column-oriented flexbox; prevents height calculation bug in IE11 where flex items were collapsing when in column orientation.

The `-ms-preferred-size` is the IE version of `flex-basis`. Width calculation from a basis of `0px` is fine, but IE11 has issues with flex child height inside an indeterminate height container. This change is limited to `column` oriented flex children.

#### BEFORE
screen shot 2017-06-19 at 1 35 56 pm

#### AFTER
screen shot 2017-06-19 at 3 53 23 pm

### DIFF
--- a/assets/scss/components/_flex.scss
+++ b/assets/scss/components/_flex.scss
@@ -77,6 +77,7 @@ Breakpoints        | Direction       | Description
 	@include flexParent('column', false);
 	height: 100%;
 	> .flex-item {
+		-ms-flex-preferred-size: auto;
 		width: 100%;
 		padding-left: 0;
 	}


akdetrick:
Did a fair bit of testing on all versions of IE; this vendor-prefixed rule will only apply to MS browsers for column-oriented flexes. Pretty minor change to fix the issue; going to SV this.

PR #238, opened 2017-06-22
use body click listener instead of blur event to avoid Safari issue

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-1386

#### Description
To close the Popover when a user clicks outside of it, Popover was using a `setTimeout` `0` because most browsers set the `document.activeElement` to `body` for an instant before reporting the actual focused element. **Safari is not most browsers**. 

A more stable approach is to attach a click listener to `body` on component mount ala `Dropdown`.

### DIFF
--- a/src/interactive/Popover.jsx
+++ b/src/interactive/Popover.jsx
@@ -118,7 +118,7 @@ class Popover extends React.Component {
 
 	onBodyClick(e) {
 		const isNotPopoverClick = [
-			this.contentRef,
+			this.menuRef,
 			this.triggerRef
 		].every(ref => !ref.contains(e.target));
 
@@ -186,7 +186,7 @@ class Popover extends React.Component {
 
 				<nav>
 					<ul
-						ref={(el) => this.contentRef = el}
+						ref={(el) => this.menuRef = el}
 						className={classNames.menu}
 						role='menu'
 						aria-hidden={!isActive}
--- a/src/interactive/Popover.jsx
+++ b/src/interactive/Popover.jsx
@@ -117,12 +117,12 @@ class Popover extends React.Component {
 	}
 
 	onBodyClick(e) {
-		const isNotPopoverClick = [
+		const isPopoverClick = [
 			this.menuRef,
 			this.triggerRef
-		].every(ref => !ref.contains(e.target));
+		].includes(e.target);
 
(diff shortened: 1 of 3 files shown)


src/interactive/Popover.jsx
```diff
@@ -172,11 +182,11 @@ class Popover extends React.Component {
 				className={classNames.popover}
 				aria-haspopup='true'
 				onKeyDown={this.onKeyDown}
-				onBlur={this.onBlur}
 				{...other}
 			>
 
 				<div
+					ref={(el) => this.triggerRef = el}
```
eilinora: Do you need these refs? I don't see them being used

akdetrick: they're used in `onBodyClick`

mmcgahan:
there are other ways to blur an element - particularly with the keyboard. Does this break the expected `onBlur` behavior in those cases?

src/interactive/Popover.jsx
eilinora: 👍

akdetrick:
@mmcgahan the `onBlur` method was only used to close the popover for clicks outside the popover, a task better handled by a body click listener. Where keyboard navigation is concerned, the popover will still close with `Escape`

src/interactive/Popover.jsx
```diff
@@ -135,6 +116,25 @@ class Popover extends React.Component {
 		return this.menuItems;
 	}
 
+	onBodyClick(e) {
+		const isNotPopoverClick = [
+			this.menuRef,
+			this.triggerRef
+		].every(ref => !ref.contains(e.target));
```
mmcgahan: really trivial, but `array.includes` would be much more terse here

```js
const isPopoverClick = [this.menuRef, this.triggerRef].includes(e.target);
if (!isPopoverClick) { this.closeMenu(); }
```

definitely not important, but just wanted to make sure you know about `.some` - in theory, it will bail out of the array loop when it finds a value, so for super-long arrays, it's a better option than `.every` checking for the negated condition.

akdetrick:
Added a simple test to catch rendering failure; anything more might be overlapping react-intl formatted date tests.

eilinora:
bump would love to close out a few bugs that this is causing :D

PR #239, opened 2017-07-01
remove login form

mmcgahan:
#### Description

Login is a dead donkey in the platform once https://github.com/meetup/meetup-web-platform/pull/316 is merged - we don't need this form.


akdetrick:
`pro-web` and `mup-web` import this component, but neither seems to actually use it in production. We could either remove usage from consumers first, or mark this as a breaking change.

mmcgahan:
@akdetrick yeah `mup-web` no longer imports it https://github.com/meetup/mup-web/search?utf8=%E2%9C%93&q=loginform&type= and I talked to @mikespencer about this PR and he's fine with it - I'll go ahead and set up a PR in `pro-web` to remove the `import` so that I can remove the `Prelim` label here.

PR #240, opened 2017-07-06
Optionally hide text-protection scrim on Stripes with bg img

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-324

#### Description
Adds a prop to `<Stripe>` and `<Modal>` (for the hero header) that allows engineers to hide the text protection scrim that used to automatically get added to a `<Stripe>` that has `backgroundImage` prop.

**This is a breaking change**, because any `<Stripe>` that has a `backgroundImage` prop but not an `inverted` prop will show black text and icons on top of the image and scrim.

This issue came up while building the [Event Home preview page](https://www.meetup.com/account/preview/).

### DIFF
--- a/src/layout/Stripe.jsx
+++ b/src/layout/Stripe.jsx
@@ -7,6 +7,7 @@ export const STRIPE_CLASS = 'stripe';
 export const STRIPE_COLLECTION_CLASS = 'stripe--collection';
 export const STRIPE_INVERTED_CLASS = 'stripe--inverted';
 export const STRIPE_HERO_CLASS = 'stripe--withBGImg';
+export const STRIPE_NOSCRIM_CLASS = 'stripe--noScrim';
 
 /**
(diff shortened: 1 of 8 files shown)


src/layout/Stripe.jsx
```diff
@@ -29,7 +31,9 @@ class Stripe extends React.Component {
 			{
 				[STRIPE_COLLECTION_CLASS]: collection,
 				[`${STRIPE_INVERTED_CLASS} inverted`]: inverted,
-				[`${STRIPE_HERO_CLASS} inverted`]: backgroundImage
+				[`${STRIPE_HERO_CLASS}`]: backgroundImage, // [`${STRIPE_HERO_CLASS} inverted`]
```
nlaz: Do you want to remove this comment?

mperrotti: Yup

PR #241, opened 2017-07-06
Added flushUntil prop to Section

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-302

#### Description
`<Section>` can now be flushed left and right. Supports media-conditional flushing


src/layout/Section.jsx
```diff
@@ -47,6 +51,10 @@ Section.propTypes = {
 	noSeparator: PropTypes.oneOfType([
 		PropTypes.bool,
 		PropTypes.oneOf(Object.keys(VALID_BREAKPOINTS))
+	]),
+	flushUntil: PropTypes.oneOfType([
+		PropTypes.bool,
```
akdetrick: feels weird to support a boolean because it's not immediately clear what this means:

`<Section flushUntil>...`

`<Section className="flush--horizontal">` may be more useful for the use case of flushing at all breakpoints. This doesn't exist in sasstools yet, but it's an easy addition...

mperrotti: Makes sense to me. I wanted to add a way to _always_ flush a `<Section>`, but I guess it's rare enough that we could just do it in `sasstools`

src/layout/Section.jsx
```diff
@@ -47,6 +51,9 @@ Section.propTypes = {
 	noSeparator: PropTypes.oneOfType([
 		PropTypes.bool,
 		PropTypes.oneOf(Object.keys(VALID_BREAKPOINTS))
+	]),
+	flushUntil: PropTypes.oneOfType([
```
nlaz: Since you got rid of the bool proptype, can't you shorten this to something like:`flushUntil: PropTypes.oneOf(Object.keys(VALID_BREAKPOINTS)),`?

mperrotti: Good idea. On it...

akdetrick:
My `flushUntil="all"` nitpick shouldn't block this merge; approving.

PR #242, opened 2017-07-06
Section noSeparator is default

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-301

#### Description
Based on Chris' latest designs, `<Section>` doesn't have a separator (bottom padding and border) by default. 

This is a breaking change, as it removes the `noSeparator` prop, but I could add the `noSeparator` code back in for now if we want to wait to clean up `<Section>` usage in apps that consume `meetup-web-components`.


mperrotti:
@akdetrick - When you say "the prop" in points 1 and 3, do you mean `noSeparator`?

akdetrick:
@mperrotti yes, sorry - I mean supporting a `noSeparator` prop that basically does nothing until we can clean up usage in mup-web

PR #243, opened 2017-07-06
adding placeholder rendering and shouldn't use IntlProvider in tests

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-69

#### Description
Text inputs aren't doing placeholder rendering. 

And after some discussion it is recommended not to do anything with International Provider in storybook, so I'll remove.

If I should do this another way let me know!

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
+import { injectIntl, intlShape } from 'react-intl';
 
 /**
  * @module TextInput
@@ -31,6 +32,7 @@ class TextInput extends React.Component {
 			className,
 			children,
 			error,
+			placeholder,
 			required,
 			id,
 			onChange, // eslint-disable-line no-unused-vars
@@ -51,6 +53,13 @@ class TextInput extends React.Component {
 			labelClassName
 		);
 
+		console.log(this.props);
+
+		const placeholderIntl = (placeholder) ? this.props.intl.formatMessage({
+			id: 'is_this_important',
+			defaultMessage: placeholder
+		}) : null;
+
 		return (
 			<div>
 				{label &&
@@ -58,11 +67,11 @@ class TextInput extends React.Component {
 						{label}
 					</label>
 				}
-
 				<input type={isSearch ? 'search' : 'text'}
 					name={name}
 					value={this.state.value}
 					required={required}
+					placeholder={placeholderIntl} // TODO use injectIntl for message
 					className={classNames}
 					onChange={this.onChange}
 					maxLength={maxLength}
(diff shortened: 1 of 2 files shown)


src/forms/TextInput.jsx
```diff
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
+import { injectIntl, intlShape } from 'react-intl';
```
eilinora: From slack :) We don't need to inject intl here as this should be handled by the user of the component. It is wrapped in an IntlProvider and provides any translations down to these MWC components so they can stay simple aka dumb components :)

src/forms/TextInput.jsx
```diff
@@ -1,6 +1,7 @@
...
 
 /**
  * @module TextInput
```
eilinora: While we are in here, would be great to make this a `PureComponent` vs `Component` as it just manages a simple state

src/forms/TextInput.jsx
eilinora: For reference code in lets say mup web would look like
<IntlProvider>
  <Form>
    <TextInput placeholder = {intl.formatMessage(trn.placeholder)} />
  </Form>
</IntlProvider>

unjust: ok. Let me try using injectINtl outside of this component then and passing it in. Thats what you are suggesting?

src/forms/TextInput.jsx
unjust: Can you give me an example? Is it just removing state?

Or can this wait... when we implement validation we will most likely be changing this component and any state, right?

akdetrick: @unjust See `src/forms/Button.jsx`. Your class will extend `React.PureComponent` instead of `Component`.

src/forms/TextInput.jsx
akdetrick: Components in this repo are basic building blocks for UI and should avoid reliance on translation utils.
I would argue it's not even necessary to use translation utils in the story. All the component and story need to know about `placeholder` is that it is a string.

unjust: Yea I ripped the IntlProvider our. I must have been advised to at some point but I'll change it.

src/forms/textInput.story.jsx
```diff
@@ -2,115 +2,129 @@ import React from 'react';
...
+			<div className='span--50'>
 				<TextInput label='Your name'
 					id='fullname'
 					name='name'
-					error={<FormattedMessage {...trn.error} />}
-					placeholder='Not your email' />
-			</IntlProvider>
+					error={'<span>This error is an element</span>'}
+					placeholder='not your email' />
+			</div>
```
unjust: So I would still like to test that error (PropType string or element). correctly renders an element.
I took out the FormattedMessage here and put in a span. This is how it looks now though and I wouldn't expect the span to be stringified.
screen shot 2017-07-06 at 4 09 19 pm

akdetrick: You can pass the `span` as the element:
`error={<span>This is an error</span>}`

PR #246, opened 2017-07-07
Wc 70 textinput pure component

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-70
related to  https://meetup.atlassian.net/browse/WC-69

#### Description
Lets make our inputs and form fields PureComponents. This way they will play well with Redux forms in the app as well.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -5,18 +5,14 @@ import cx from 'classnames';
 /**
  * @module TextInput
  */
-class TextInput extends React.Component {
+class TextInput extends React.PureComponent {
 
 	constructor(props) {
 		super(props);
-		this.state = {
-			value: props.value || '',
-		};
 		this.onChange = this.onChange.bind(this);
 	}
 
 	onChange(e) {
-		this.setState({ value: e.target.value });
 		if (this.props.onChange) {
 			this.props.onChange(e);
 		}
@@ -25,7 +21,7 @@ class TextInput extends React.Component {
 	render() {
 		const {
 			name,
-			value, // eslint-disable-line no-unused-vars
+			value,
 			label,
 			labelClassName,
 			className,
@@ -61,7 +57,7 @@ class TextInput extends React.Component {
 				}
 				<input type={isSearch ? 'search' : 'text'}
 					name={name}
-					value={this.state.value}
+					value={value}
 					required={required}
 					placeholder={placeholder}
 					className={classNames}
@@ -71,7 +67,7 @@ class TextInput extends React.Component {
 					disabled={disabled}
 					id={id} />
 
-				{ this.props.maxLength && <p className='text--caption align--right'>{this.state.value.length} / {this.props.maxLength}</p> }
+				{ maxLength && <p className='text--caption align--right'>{value.length} / {maxLength}</p> }
 
 				{ error && <p className='text--error'>{error}</p> }
 				{children}
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -5,7 +5,7 @@ import cx from 'classnames';
 /**
  * @module TextInput
  */
-function TextInput(props) {
+const TextInput = (props) => {
 
 	const {
 		name,
@@ -61,7 +61,7 @@ function TextInput(props) {
 			{children}
 		</div>
 	);
-}
+};
 
 TextInput.propTypes = {
 	name: PropTypes.string.isRequired,
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -22,7 +22,8 @@ const TextInput = (props) => {
 		isSearch,
 		maxLength,
 		pattern,
-		disabled
+		disabled,
+		...other
 	} = props;
 
 	const classNames = cx(
@@ -53,7 +54,9 @@ const TextInput = (props) => {
 				maxLength={maxLength}
 				pattern={pattern}
 				disabled={disabled}
-				id={id} />
+				id={id}
+				{...other}
+			/>
 
 			{ maxLength && <p className='text--caption align--right'>{value.length} / {maxLength}</p> }
 
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -5,75 +5,62 @@ import cx from 'classnames';
 /**
  * @module TextInput
  */
-class TextInput extends React.PureComponent {
+function TextInput(props) {
 
-	constructor(props) {
-		super(props);
-		this.onChange = this.onChange.bind(this);
-	}
+	const {
+		name,
+		value,
+		label,
+		labelClassName,
+		className,
+		children,
+		error,
+		placeholder,
+		required,
+		id,
+		onChange,
+		isSearch,
+		maxLength,
+		pattern,
+		disabled
+	} = props;
 
-	onChange(e) {
-		if (this.props.onChange) {
-			this.props.onChange(e);
-		}
-	}
+	const classNames = cx(
+		{ 'field--error': error },
+		className
+	);
 
-	render() {
-		const {
-			name,
-			value,
-			label,
-			labelClassName,
-			className,
-			children,
-			error,
-			placeholder,
-			required,
-			id,
-			onChange, // eslint-disable-line no-unused-vars
-			isSearch,
-			maxLength,
-			pattern,
-			disabled
-		} = this.props;
+	const labelClassNames = cx(
+		'label--field',
+		{ required, disabled },
+		labelClassName
+	);
 
-		const classNames = cx(
-			{ 'field--error': error },
-			className
-		);
+	return (
+		<div>
+			{label &&
+				<label className={labelClassNames} htmlFor={id}>
+					{label}
+				</label>
+			}
+			<input type={isSearch ? 'search' : 'text'}
+				name={name}
+				value={value}
+				required={required}
+				placeholder={placeholder}
+				className={classNames}
+				onChange={onChange}
+				maxLength={maxLength}
+				pattern={pattern}
+				disabled={disabled}
+				id={id} />
 
-		const labelClassNames = cx(
-			'label--field',
-			{ required, disabled },
-			labelClassName
-		);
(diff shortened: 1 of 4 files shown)


src/forms/TextInput.jsx
```diff
@@ -5,18 +5,14 @@ import cx from 'classnames';
...
-		this.state = {
-			value: props.value || '',
-		};
 		this.onChange = this.onChange.bind(this);
 	}
 
 	onChange(e) {
-		this.setState({ value: e.target.value });
 		if (this.props.onChange) {
 			this.props.onChange(e);
```
mikespencer: If you attached this directly to the [onChange prop](https://github.com/meetup/meetup-web-components/pull/246/files#diff-ec0100a5c023d94b8141bd5e7c37d91eR64) of the input, you could really trim down a lot of the code in this component:

```jsx
<input onChange={this.props.onChange} ... />
```

You'd be able to delete the `constructor` and `onChange` methods altogether while keeping the same functionality. You could then even go one step further and define this as a stateless functional  component instead of a class if you wanted (no need to worry about `this` or lifecycle methods):

```jsx
class TextInput extends React.PureComponent {
  render() {
    const { /* destructure props */ } = this.props;
    return <input ... />
  }
}

// vs

function TextInput({/* destructure props */}) {
  return <input ... />
}
```

unjust: hm ok. Well what do we prefer? `PureComponent` vs stateless functional component? What are the benefits?
@eilinora  should we go for the latter instead of a PureComponent?

eilinora: I agree with @mikespencer we should be trying to go for stateless functional components as much as possible. Especially at the MWC level. I think my only concern is how this change would interact if we wanted this component to work with _redux forms_. @mikespencer any concerns there?

unjust:
I made this a functional component but the most basic tests (eg it exists) are even failing. Stroybook looks good. Any ideas how the test needs to change @mikespencer

unjust:
Tests need to be changed since this is a functional component, and TestUtils.renderIntoDocument returns null for these tyes of components
"NOTE:
Because stateless functions don't have a backing instance, you can't attach a ref to a stateless function component. Normally this isn't an issue, since stateless functions do not provide an imperative API. Without an imperative API, there isn't much you could do with an instance anyway. However, if a user wants to find the DOM node of a stateless function component, they must wrap the component in a stateful component (eg. ES6 class component) and attach the ref to the stateful wrapper component."
https://github.com/facebook/react/issues/5455#issuecomment-156013171

My fix is to create a `Wrapper` class, like this example:
https://github.com/facebook/react/issues/4692#issuecomment-163029873

src/forms/TextInput.jsx
```diff
@@ -5,79 +5,62 @@ import cx from 'classnames';
 /**
  * @module TextInput
  */
-class TextInput extends React.Component {
+function TextInput(props) {
```
eilinora: can i be obnoxious and ask this to be written es6'y?
`const TextInput = props => { ... };`

src/forms/textInput.test.jsx
```diff
@@ -10,6 +11,16 @@ describe('TextInput', function() {
 		MAX_LEN = '20',
 		ERROR_TEXT = 'Too wimpy.';
 
+	// Note: since TextInput is a functional component
+	// we need to wrap it in a stateful component to use TestUtils effectively
+	const TestWrapper = createReactClass({
```
eilinora: might be good to make this a utility as we move other MWC components to stateless. i suggest in `testUtils`

src/forms/textInput.test.jsx
```diff
@@ -21,12 +32,14 @@ describe('TextInput', function() {
 			required: true,
 		};
 		textInputComponent = TestUtils.renderIntoDocument(
```
eilinora: A renderComponent function might also be helpful for future updates... e.g. https://github.com/meetup/mup-web/blob/master/src/components/events/eventTimeDisplay.test.jsx#L19

src/forms/TextInput.jsx
unjust: happy to oblige

src/forms/textInput.test.jsx
unjust: since I won't be using that in these tests I'm reluctant to add it. I will when I do though!

src/forms/TextInput.jsx
mikespencer: One thing to note with doing it this way (anonymous function) is that it can be trickier to figure out what the problem is if you are looking at an error's call stack. Personally, I prefer to reference [airbnb's style guide](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless) where they recommend explicitly naming the function in these sort of situations. Not a major thing, but something to be aware of

src/forms/TextInput.jsx
mikespencer: > I think my only concern is how this change would interact if we wanted this component to work with redux forms. @mikespencer any concerns there?

Ah right - so this won't work *as is* with redux-form. [Redux-form passes in props in a specific way](http://redux-form.com/6.8.0/docs/api/Field.md/#props) to the rendered component. I like the idea of having this component being purely presentational though with no ties to a specific library, like redux-form.

One way I'd love to see this work would be to wrap this in another component that **is** meant for consumption by redux-form, roughly something like:

```jsx
import TextInput from 'meetup-web-components';

const ReduxFormTextInput = function ReduxFormTextInput(props) {
  const translatedProps = {
    // convert the props that redux-form passes into this component into something that TextInput understands
  }
  return <TextInput {...translatedProps} />;
}
```

That way, this TextInput can be used in different situations purely for presentation and is totally decoupled from redux-form

mikespencer:
I just realised I left a [comment](https://github.com/meetup/meetup-web-components/pull/246#discussion_r126499523) in a collapsed comment, so in case it gets missed I'm just going to paste it here:
 

> I think my only concern is how this change would interact if we wanted this component to work with redux forms. @mikespencer any concerns there?

Ah right - so this won't work *as is* with redux-form. [Redux-form passes in props in a specific way](http://redux-form.com/6.8.0/docs/api/Field.md/#props) to the rendered component. I like the idea of having this component being purely presentational though with no ties to a specific library, like redux-form.

One way I'd love to see this work would be to wrap this in another component that **is** meant for consumption by redux-form, roughly something like:

```jsx
import TextInput from 'meetup-web-components';

const ReduxFormTextInput = function ReduxFormTextInput(props) {
  const translatedProps = {
    // convert the props that redux-form passes into this component into something that TextInput understands
  }
  return <TextInput {...translatedProps} />;
}
```

That way, this TextInput can be used in different situations purely for presentation and is totally decoupled from redux-form

mikespencer:
> ok @mikespencer, so the ReduxFormTextInput or something like that would live in mup-web correct?
if this review is good to go, I'll merge - just let me know!

Yeah I think that one could be debated - if the goal of meetup-web-components is to be purely presentational, then I'd say it's probably best not to have components coupled to redux-form inside here, since it assumes the consuming application is using redux-form. Although, if the majority of the repos consuming this meetup-web-components *do* use redux-form, then having a redux-form input component in here could be an easy easy way to share between applications. I guess an ideal approach would be to have another package for meetup-redux-form-components that basically wrap meetup-web-components to make them compatible with redux-form, but that's just my opinion!

Either way, I think all of the above are out of the scope of this PR, so I'm 👍 with this.

eilinora:
Oh that brings up a good point, there should be a `...other` in here on the input in case we need to pass in something this component was initially setup to support. (.e.g. if we did want to manage state of the input at the form level, etc.

src/forms/TextInput.jsx
aquillano: I was thinking the same thing. Torn between es6 syntax and a named function. @ivy @eilinora

src/utils/testUtils.js
```diff
@@ -32,3 +33,13 @@ export const hasRoleAttribute = (el, roleName) => {
 export const componentHasProperty = (component, prop, value) => {
 	expect(component && component.props && component.props[prop] === value).toBe(true);
 };
+
+// Note: functional components need to be
+// wrapped it in a stateful component to use TestUtils effectively
+export const TestWrapper = createReactClass({
+	render: function() {
```
mmcgahan: in general the preferred solution would be to use enzyme instead of `TestUtils.renderIntoDocument` in these cases - I think it's fine to add this helper for 'legacy' code, but I'd like to see a `@deprecated` code comment and (ideally) a `console.warn('Using deprecated TestWrapper for TestUtils.renderIntoDocument - covert test to use enzyme')`

unjust: ok so - how do we start using enzyme? we can do this in a separate branch and I can make a ticket for that. if you want I can add the deprecated comment but I dont know if that really helps?... as of this tool set and library this code is valid.

unjust: https://meetup.atlassian.net/browse/WC-72

src/forms/TextInput.jsx
eilinora: Hmmm, yeah we've been doing it the es6 way as we've been migrating to stateless. I see the style guide says boo to it. I vote for the es6, but I'll defer to the collective on which way we go. 😃

src/utils/testUtils.js
mmcgahan: @unjust I can go over some of the enzyme stuff with you, but I can't remember who's on your team - would @eilinora or @sadafie be able to give you a quick rundown?

For that WC-72 ticket, could you add a note about removing `TestWrapper` alongside those changes?

src/forms/TextInput.jsx
mmcgahan: I recommend using an arrow function + static `.displayName` property _or_ a named function, but not a 'bare' arrow function without a name.

src/utils/testUtils.js
unjust: sure no problem. steve and I are on the team. I can do some reading - I just would like to do this upgrade/changes in a separate ticket to make sure it corrects this problem. I'm also checking in with adam to make sure he knows.

unjust:
merging, stay tuned for enzyme

PR #248, opened 2017-07-11
Wc 72 enzyme

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-72
removes helper committed in https://meetup.atlassian.net/browse/WC-71

#### Description
Updating tests to use enzyme and snapshots.
Enzyme allows us to make calls to render and dom in an easier way than TestUtils.
We can also compare a snapshot of the rendered component instead of writing a bunch of prop tests.


src/forms/textInput.test.jsx
```diff
@@ -1,128 +1,117 @@
...
-		textInputComponent = TestUtils.renderIntoDocument(
-			<TestWrapper>
-				<TextInput
-					name={NAME_ATTR}
-					label={LABEL_TEXT}
-					value={VALUE}
-					{...formAttrs}
-				/>
-			</TestWrapper>
+		textInputComponent = mount(
```
akdetrick: More of a question than an issue... 

It doesn't look like we're testing lifecycle methods here, so is there a good reason to use `mount` over `render` here?

src/forms/textInput.test.jsx
```diff
@@ -1,128 +1,117 @@
...
 
 	afterEach(() => {
 		textInputComponent = null;
 		inputEl = null;
 	});
 
-	it('exists', () => {
-		expect(() => TestUtils.findRenderedDOMComponentWithTag(textInputComponent, 'input')).not.toThrow();
+	it('matches snapshot', () => {
+		expect(toJson(textInputComponent)).toMatchSnapshot();
```
akdetrick: When using `render` instead of `mount`, I was able to avoid the `toJson` conversion (see my question above)

src/forms/textInput.test.jsx
mmcgahan: I think in general we should prefer `shallow` over `mount`, and `mount` over `render` - I think that's ordered by increasing complexity/time-to-complete.

src/forms/textInput.test.jsx
mmcgahan: `toJson` will (probably) make the snapshots a little faster and easier to read diffs for

mmcgahan: snapshots can actually be made for any object, so they can be useful if you have a big complex object (e.g. a state tree) that needs to be organized _just so_. The 'rendering' snapshots are special because they serialize the React element into a human-readable format rather than just labeling every prop as a `[Function]` or whatever

src/forms/__snapshots__/textInput.test.jsx.snap
```diff
@@ -0,0 +1,34 @@
...
+  >
+    Super Hero
+  </label>
+  <input
+    className="field--error"
+    id="superhero"
+    maxLength={20}
+    name="superhero"
+    onChange={[Function]}
+    required={true}
```
unjust: is this the way it should be rendered `required={true}` rather than just a required `attr`?
same for maxLength...

package.json
```diff
@@ -66,13 +66,16 @@
...
+    "enzyme": "2.9.1",
+    "enzyme-to-json": "1.5.1",
     "eslint": "4.2.0",
     "eslint-loader": "1.9.0",
     "eslint-plugin-react": "7.1.0",
     "gh-pages": "1.0.0",
     "husky": "0.14.3",
     "inquirer": "3.2.0",
     "jest": "19.0.2",
+    "jest-enzyme": "3.4.0",
```
mmcgahan: I'm curious what you're using this for - I'm a little suspicious of this helper because it adds 'non-standard' test syntax - I think the underlying `enzyme` API is sufficient?

src/forms/__snapshots__/textInput.test.jsx.snap
mmcgahan: this is fine - it's rendering the React component, not the actual HTML, so it won't necessarily look like the HTML you get on the page

package.json
unjust: ok. not sure if I explicitly added that. so I should just be able to add `enzyme` to the project, no need for jest-enzyme you're saying? I'll try and remove.

src/forms/textInput.test.jsx
```diff
@@ -1,129 +1,120 @@
...
-		textInputComponent = TestUtils.renderIntoDocument(
-			<TestWrapper>
+	const formAttrs = {
+		id: NAME_ATTR,
+		maxLength: MAX_LEN,
+		required: true,
+		error: ERROR_TEXT,
+	};
+
+	describe('snapshot check', () => {
```
mmcgahan: no need for a nested `describe` if you only have one test inside it - `describe` is generally used to group a collection of related unit tests.

package.json
unjust: thanks for catching that. yarn makes it too easy to add things!

src/forms/textInput.test.jsx
```diff
@@ -1,129 +1,120 @@
...
+		it('matches snapshot', () => {
```
mmcgahan: probably better to be more specific about the inputs used to make the snapshot - it might 'match the snapshot' when given certain props, but not match the snapshot when given other props, so there's no general "matches snapshot" guarantee.

src/forms/__snapshots__/textInput.test.jsx.snap
unjust: ok I opted for `shallow` instead of `mount` here to be a little more html focused.

src/forms/textInput.test.jsx
unjust: hrm

mmcgahan: e.g. `it('renders an HTML <input> with expected attributes for mock data', () => ...`

unjust:
now I'm wrestling with the fact that there is an obsolete snapshot in our build so it fails
```Snapshot Summary
 › 1 obsolete snapshot found, run with `npm run test:base -- -u` to remove them.```

should I handle this by changing the travis build command?

**Comment by unjust** [comment]
@akdetrick @mmcgahan how are we feeling with the changes?
I opted for `shallow` rendering when possible, removed a few basic prop checks, and use `mount` to look at label and error dom.

File: src/forms/textInput.test.jsx [hunk 7]
```diff
@@ -1,129 +1,117 @@
...
+				name={NAME_ATTR}
+				label={LABEL_TEXT}
+				value={VALUE}
+				id={NAME_ATTR}
+				required
+			/>
 		);
 
-		inputEl = TestUtils.findRenderedDOMComponentWithTag(textInputComponent, 'input');
+		expect(toJson(component)).toMatchSnapshot();
```
sadafie: Just fyi I setup on a jest configuration to default to the `toJson()` snapshot serializer in `mup-web`: https://github.com/meetup/mup-web/pull/1291#discussion_r126980339

If we decide to use it, might be nice to have it in this repo as well.

unjust: ok I'll do the same!

PR #249, opened 2017-07-17
Added the ability to keep grid item heights consistent

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-348

#### Description
Sometimes we need gridlist items to be the same height. This PR lets you use flexbox for gridlist layout. This came from Chris' designs for showing a sample of members in cards on group home.

screen shot 2017-07-17 at 12 21 05 pm

### DIFF
--- a/src/layout/GridList.jsx
+++ b/src/layout/GridList.jsx
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
 
-export const GRID_AUTOHEIGHT_CLASS = 'gridList--autoheight';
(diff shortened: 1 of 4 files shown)


src/layout/GridList.jsx
```diff
@@ -2,6 +2,8 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
 
+export const GRID_AUTOHEIGHT_CLASS = 'gridList--autoheight';
```
akdetrick: stupid nitpick, but could we change the prop name and class name to camel case `autoHeight`? This fits the convention we've been using for multi word prop names.

PR #250, opened 2017-07-17
Added xxs media size to be used only for icons

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-350

#### Description
Recent designs have called for an even smaller icon size. e.g.: - icons in buttons, "like" heart on comments

### DIFF
--- a/src/media/icon.story.jsx
+++ b/src/media/icon.story.jsx
@@ -28,8 +28,11 @@ storiesOf('Icon', module)
 			<Icon shape={ICON_NAME} inverted />
 		</Inverted>
 	))
-	.add('x-Small', () => (
+	.add('xx-Small', () => (
 		<Icon shape={ICON_NAME} size='xxs' />
+	))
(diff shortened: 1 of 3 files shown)


src/media/icon.story.jsx
```diff
@@ -19,46 +19,32 @@ storiesOf('Icon', module)
...
-		<div className='margin--center'>
-			<Inverted>
-				<Icon shape={ICON_NAME} inverted />
-			</Inverted>
-		</div>
+		<Inverted>
+			<Icon shape={ICON_NAME} inverted />
+		</Inverted>
 	))
 	.add('x-Small', () => (
```
akdetrick: missing a story for extra extra small

akdetrick: actually, we're missing `xs` and `xxs` is mislabeled

mperrotti: Good catch - on it

PR #251, opened 2017-07-18
Updated checkbox visual design

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-349

#### Description
Updated checkbox style to match latest visual design updates from Chris and Max. Radios coming soon...
From Sketch:
screen shot 2017-07-18 at 3 35 50 pm

From Storybook:
screen shot 2017-07-18 at 3 36 53 pm

### DIFF
--- a/assets/scss/components/_forms.scss
+++ b/assets/scss/components/_forms.scss
@@ -193,31 +193,30 @@ $C_toggleHighlight: $C_lightPurple; // replace $C_lightPurple with $C_blue once 
 	&.checked {
 		background-color: $C_toggleHighlight;
 		border-color: transparent;
-
-		&.disabled {
-			background-color: $C_white;
-			border-color: $C_textHint;
-		}
-
-		&.focused {
-			box-shadow: 0 0 4px 1px $C_toggleHighlight;
-		}
-
 	}
 
 	&.disabled {
 		@include color-all($C_textHint);
 		border-color: $C_textHint;
+
+		&.checked {
+			background-color: $C_white;
+			border-color: $C_textHint;
+		}
 	}
 
 	&.focused {
 		border-color: $C_toggleHighlight;
+
+		&.checked {
+			box-shadow: 0 0 4px 1px $C_toggleHighlight;
+		}
 	}
 
 }
 
 .fauxToggle--checkbox {
-	border-radius: 2px; // replace with $smallRadius once https://github.com/meetup/swarm-sasstools/pull/32 merges
+	border-radius: $smallRadius;
 }
 
 .fauxToggle--radio {
--- a/assets/scss/components/_forms.scss
+++ b/assets/scss/components/_forms.scss
@@ -149,3 +149,78 @@ Faux inputs visually mimic actual input elements.
 		}
 	}
 }
+
+$toggleSize: 20px;
+$C_toggleHighlight: $C_lightPurple; // replace $C_lightPurple with $C_blue once we start to use `swarm-constants`
+
+.toggleLabel {
+	padding-bottom: $space-half;
+
+	.toggleLabel-container {
+		padding-left: $space-half;
+	}
+
+	&:hover {
+
+		.fauxToggle {
+			border-color: $C_toggleHighlight;
+
+			&.checked {
+				background-color: darken($C_toggleHighlight, 15%);
+			}
+
+			&.disabled {
+				background-color: $C_white;
+				border-color: $C_textHint;
+			}
+
+		}
+
+	}
+
+}
+
+.fauxToggle {
+	background: $C_white;
+	border: 1px solid $C_textSecondary;
+	height: $toggleSize;
+	width: $toggleSize;
+
+	.svg-icon {
+		@include color-all($C_textPrimaryInverted);
+	}
+
+	&.checked {
+		background-color: $C_toggleHighlight;
+		border-color: transparent;
+
+		&.disabled {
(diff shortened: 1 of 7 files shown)


assets/scss/components/_forms.scss
```diff
@@ -149,3 +149,78 @@ Faux inputs visually mimic actual input elements.
 		}
 	}
 }
+
+$toggleSize: 20px;
+$C_toggleHighlight: $C_lightPurple; // replace $C_lightPurple with $C_blue once we start to use `swarm-constants`
+
+.toggleLabel {
```
mperrotti: Is this CSS too crazy/specific?

mperrotti:
Tested in FF, Chrome, Safari, Mobile Chrome, Mobile Safari

assets/scss/components/_forms.scss
```diff
@@ -149,3 +149,78 @@ Faux inputs visually mimic actual input elements.
...
+	padding-bottom: $space-half;
+
+	.toggleLabel-container {
+		padding-left: $space-half;
+	}
+
+	&:hover {
+
+		.fauxToggle {
```
unjust: is `.toggle` already taken? should we keep the `faux`?

assets/scss/components/_forms.scss
```diff
@@ -149,3 +149,78 @@ Faux inputs visually mimic actual input elements.
...
+$C_toggleHighlight: $C_lightPurple; // replace $C_lightPurple with $C_blue once we start to use `swarm-constants`
+
+.toggleLabel {
+	padding-bottom: $space-half;
+
+	.toggleLabel-container {
+		padding-left: $space-half;
+	}
+
+	&:hover {
```
unjust: do we want `:hover` styles here or `:focus`?

assets/scss/components/_forms.scss
```diff
@@ -149,3 +149,78 @@ Faux inputs visually mimic actual input elements.
...
+			border-color: $C_textHint;
+		}
+
+		&.focused {
+			box-shadow: 0 0 4px 1px $C_toggleHighlight;
+		}
+
+	}
+
+	&.disabled {
```
unjust: nitpick: can we move `disabled` and `focused` above `checked` rules since `checked` has more disabled, focused nested within?

src/forms/checkbox.test.jsx
```diff
@@ -20,52 +21,106 @@ describe('Checkbox', function() {
 		checkbox = null;
 	});
 
-	it('exists', function() {
-		expect(() => TestUtils.findRenderedDOMComponentWithTag(checkboxComponent, 'input')).not.toThrow();
+	describe('basic', function() {
+		it('exists', function() {
+			expect(() => TestUtils.findRenderedDOMComponentWithTag(checkboxComponent, 'input')).not.toThrow();
```
unjust: we should change these tests to use `enzyme`. The `exists` test can use snapshotting. See https://github.com/meetup/meetup-web-components/pull/248

also you can do a `shallow` render of the checkbox component and then look for attributes and other behavior. We don't need to test all props/attributes of the element if we are confident the snapshot rendered correctly.
 https://github.com/meetup/meetup-web-components/pull/248/files#diff-2690983ef9c689ac0361949ee7bfd4d9R41

unjust:
The only blocking changes are the ones to testing. We should move away from `TestUtils` in favor of `enzyme` rendering and api.

src/forms/Checkbox.jsx
```diff
@@ -14,12 +20,22 @@ class Checkbox extends React.Component {
 			checked: props.checked || false
 		};
 		this.onChange = this.onChange.bind(this);
+		this.onBlur = this.onBlur.bind(this);
+		this.onFocus = this.onFocus.bind(this);
 	}
 
 	onChange(e) {
 		this.setState({ checked: e.target.checked });
 	}
```
unjust: we will want to move away from this being stateful into something stateless, but that can be done in a different PR.

assets/scss/components/_forms.scss
mperrotti: I'm more than happy to use something other than `toggle`. Any ideas what name might be good to describe both a `input type=checkbox` as well as a `input type=radio`?

assets/scss/components/_forms.scss
mperrotti: `:hover`

assets/scss/components/_forms.scss
mperrotti: I'll give it a try!

src/forms/Checkbox.jsx
mperrotti: Good catch. Should we file a ticket now or later?

src/forms/checkbox.test.jsx
mperrotti: Awesome - I'll give that a try. May bug you or Adam for help since I've never used `enzyme` before

assets/scss/components/_forms.scss
mperrotti: Is `fauxBoolControl` or `fauxBoolInput` too awkward sounding?

assets/scss/components/_forms.scss
akdetrick: not really, although `toggleLabel-container` does not need to be and probably shouldn't be nested in `toggleLabel`.

PR #254, opened 2017-07-28
Antialias primary button type

mperrotti:
#### Description
Not related to any specific issues. I changed primary button text from subpixel antialiasing to regular antialiasing. The subpixel antialiasing was causing the white type to appear bolder on the red background. Now the type actually appears like "Medium" weight instead of "Semibold" weight, and better matches the default (secondary) buttons.

Before:
screen shot 2017-07-28 at 12 02 11 pm

After:
screen shot 2017-07-28 at 12 02 02 pm


mperrotti:
Decided I actually prefer the slightly bolder-looking text over the slightly lighter-looking text

¯\_(ツ)_/¯

PR #259, opened 2017-08-02
Textarea: Disassociate autosize functionality from the rows param/attribute and add new autosize param

aquillano:
Textarea: Disassociate autosize functionality from the rows param/attribute and add new autosize param

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-75

#### Description
Currently the rows param either accepts an integer which is then translated to the "rows" HTML attribute, or you can set rows to "auto" which enables auto-sizing as you type but also defaults the "rows" HTML attribute to 1.
I want to enable autosize, but also set a greater "row" value.
This ticket will break "rows" and "autosize" into 2 distinct params.


src/forms/textarea.story.jsx
```diff
@@ -51,7 +51,7 @@ storiesOf('Textarea', module)
 			<InfoWrapper>
 				<Bounds><Section>
 					<Textarea
-						rows='auto'
+						autosize
```
mperrotti: Add `rows={1}` to preserve the same functionality we had before separating

aquillano:
@mperrotti I did look for existing uses, but forgot to update `mup-web:master`... I'll take a look so that this change doesn't break existing features.

mmcgahan:
Breaking change should bump the version specified in the Makefile https://github.com/meetup/meetup-web-components/blob/master/Makefile#L2

src/forms/textarea.story.jsx
aquillano: I discussed with Mike, we don't want to default to `rows={1}` because there is a case where you don't want to set `rows` at all.

PR #262, opened 2017-08-02
Added a large avatar size option

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-364

#### Description
Added a `large` avatar size. Marking as prelim because I would prefer to use `$media-l`, but have to check with Chris first


mperrotti:
Checked with Chris about using `$media-l` - he'd rather stick with what we have for now.
The design team will be taking a closer look at the `AvatarRow` component, so it's possible that will result in `<Avatar large />` changing size.

PR #265, opened 2017-08-06
Wc 76/update sasstools new colors

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-76

#### Description
Updates `swarm-sasstools`, which pulls colors from `swarm-constants` instead of `meetup-swatches`.

Marked as a **breaking change** because consumer apps may need to remap some older swatches color var names to new standard colors from `swarm-constants`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "storybook-addon-i18n-tools": "1.0.0",
     "style-loader": "0.18.2",
     "swarm-icons": "1.3.129",
-    "swarm-sasstools": "1.7.78",
+    "swarm-sasstools": "1.7.80",
     "webpack": "2.6.1"
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -88,7 +88,7 @@
     "storybook-addon-i18n-tools": "1.0.0",
     "style-loader": "0.18.2",
     "swarm-icons": "1.3.129",
-    "swarm-sasstools": "1.7.62",
+    "swarm-sasstools": "1.7.78",
```
mperrotti: We're at 1.7.80 now. Otherwise LGTM

akdetrick: may as well make the update before merging; thanks

PR #268, opened 2017-08-08
upgrade swarm-sasstools for JS dist update in swarm-constants

akdetrick:
Minor update to sasstools, which includes an updated version of `swarm-constants` that has a JS dist for react apps


akdetrick:
This is a very minor update; merging to get the additional JS dist from `swarm-constants`

PR #267, opened 2017-08-07
Radio button and radio button group

mikespencer:
#### Description
Adds:
- `<RadioButton />`
- `<RadioButtonGroup />`
- `redux-form` compatible `<RadioButtonGroup />`

PR #269, opened 2017-08-11
Wc 79 redux forms dir

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-79

#### Description
After the redux form faff, we decided to keep wrapping components in meetup-web-components.
In this branch, I'm copying our first ReduxTextInput component over to this project and creating a test.

Will change dependencies in mup-web in a follow up MW-1761


src/forms/reduxForm/__snapshots__/reduxTextInput.test.jsx.snap
```diff
@@ -0,0 +1,472 @@
...
+                      label="Super Hero"
+                      maxLength={20}
+                      name="superhero"
+                      onBlur={[Function]}
+                      onChange={[Function]}
+                      onDragStart={[Function]}
+                      onDrop={[Function]}
+                      onFocus={[Function]}
+                      required={true}
+                      value=""
```
unjust: @mikespencer not sure why the value here is blank. I passed a value prop in.

src/forms/reduxForm/reduxTextInput.test.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
unjust: I'm trying to test that our wrapping ReduxTextInput component renders out a TextInput with a snapshot containing the expected props.

After working through different errors, this test needed a Form hooked up with reduxForm, and Provider. I followed this example https://github.com/erikras/redux-form/issues/849.

unjust:
I'd like feedback for:
* dir structure and naming
* dependencies that are added
* am I testing the right thing ?
* do I need to add an index.js file or whatever to import all the wrappers ?

src/forms/reduxForm/reduxTextInput.test.jsx
mmcgahan: Yeah, you can use `shallow` to do the necessary unit test:

```jsx
import React from 'react';
import { shallow } from 'enzyme';
import ReduxTextInput from './ReduxTextInput';

describe('ReduxTextInput', function() {
	const formAttrs = {
		input: {
			label: 'Super Hero',
			name: 'superhero',
			value: 'Wonder Woman and Robin',
			maxLength: 20,
			required: true,
		},
		meta: { error: 'Did you mean Batman and Robin?' },
	};

	it('renders a required HTML <input> with expected attributes for mock data', () => {
		const component = shallow(<ReduxTextInput {...formAttrs} />);

		expect(component).toMatchSnapshot();
	});
});
```

yields this snapshot:

```js
exports[`ReduxTextInput renders a required HTML <input> with expected attributes for mock data 1`] = `
<TextInput
  error="Did you mean Batman and Robin?"
  label="Super Hero"
  maxLength={20}
  name="superhero"
  required={true}
  value="Wonder Woman and Robin"
/>
`;
```

I think the issue that you noticed about the missing `value` in the snapshot was related to the structure of your `formAttrs` not being quite right

unjust: ok. Thanks for trying that out. I think I wanted to see if the snapshot rendered the correct html all the way through, but you're right - we do have another test for `TextInput` that does that. I'll clean this up.

src/forms/reduxForm/__snapshots__/reduxTextInput.test.jsx.snap
mikespencer: There's a lot going on in this snapshot, so it's hard to tell where the issue is coming from - I'd recommend testing the `<TextInput />` in isolation and see if it's still not getting the `value` passed through.

package.json
```diff
@@ -44,7 +44,9 @@
     "autosize": "3.0.21",
     "base64-image-loader": "1.2.0",
     "flatpickr": "3.0.6",
-    "prop-types": "15.5.10"
+    "prop-types": "15.5.10",
+    "react-redux": "5.0.6",
+    "redux-form": "7.0.3"
```
mikespencer: I actually don't think you need `react-redux` and `redux-form`. I have an [open PR](https://github.com/meetup/meetup-web-components/pull/267/files#diff-ccd4c1d6611298c48e393c9ad13d8148) at the moment for a radio button and radio button group that has a redux-form wrapper that might be a good point of reference for how to accomplish this without these dependencies. Basically, I'm just mocking the props that redux-form sends through to the component instead of actually hooking everything up to redux-form. That _should_ make your tests much more simple and reliable too.

src/forms/reduxForm/ReduxTextInput.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import TextInput from '../TextInput';
+
+/**
+ * @param {Object} props - props passed in from parent and redux-form
+ * @description wraps standard TextInput web component for use with redux-form
+ * deconstructs props that redux-forms sets and sets them on TextInput
+ * @return {Component} TextInput
+ */
+const ReduxTextInput = props => {
```
mikespencer: Since this isn't a named function, it's probably safer to manually add a `displayName` to this component (may come in handy with debugging later on)

```js
ReduxTextInput.displayName = 'ReduxTextInput';
```

unjust: ok I can. I'm unfamiliar with this feature though.

src/forms/redux-form/__snapshots__/reduxTextInput.test.jsx.snap
```diff
@@ -0,0 +1,28 @@
...
+  <Field
+    component={[Function]}
+    error="Did you mean Batman and Robin?"
+    label="Super Hero"
+    maxLength={20}
+    name="superhero"
+    required={true}
+    value="Wonder Woman and Robin"
+  />
+</Connect(Form(FormComponent))>
```
mmcgahan: this snapshot isn't testing `ReduxTextInput` - it's rendering the wrapper from your unit test, which isn't needed. You should shallow-render _only_ the component being tested, nothing else.

unjust:
@mikespencer @mmcgahan if you have any more feedback let me know, if its good to go let me know too :)

PR #270, opened 2017-08-11
Wc 80 popover animation

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-80

#### Description
Adding swarm-animation dependency + css partials to mwc project to use transition classes in Popover.

If you build and run storybook, you should now see a subtle opacity animation.
screen shot 2017-08-11 at 4 19 26 pm


assets/scss/main.scss
```diff
@@ -14,3 +14,9 @@
 //
 @import "accordion";
 @import "popover";
+
+//
+// SWARM-ANIMATION
+//
+@import "~swarm-animation/src/constants";
+@import "~swarm-animation/src/fade";
```
unjust: @akdetrick  is this the way we agreed upon importing css dependencies?

akdetrick: yup

akdetrick: actually, we may want to import these in the component partials that depend on them

akdetrick: For reference, consumers of `meetup-web-components` import `assets/scss/components/_all.scss`. If an individual component needs something from `swarm-animation`, it should be imported in `_all.scss` or in the component partial

unjust: ok I will put it in `_all` then.

PR #271, opened 2017-08-14
Add toast components

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-15

#### Description
Adding a component that allows us to give users feedback on their actions, or pop up other messaging.

Tests are very much a work in progress, but I'm open to feedback on everything :)

screen shot 2017-08-14 at 12 58 59 pm
screen shot 2017-08-14 at 12 59 20 pm
screen shot 2017-08-14 at 12 59 27 pm
screen shot 2017-08-14 at 12 59 36 pm


src/interactive/Toast.jsx
```diff
@@ -0,0 +1,104 @@
...
+ */
+class Toast extends React.PureComponent {
+	constructor(props) {
+		super(props);
+
+		this.handleDismiss = this.handleDismiss.bind(this);
+	}
+
+	handleDismiss() {
+		if (this.props.setDismissedToast) {
```
eilinora: you could shorthand this as :)
`this.props.setDismissedToast && this.props.setDismissedToast(this);`

mperrotti: Into it. Will change :)

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import cx from 'classnames';
+
+export const DELAY_TIME = 3000;
+
+/**
+ * @module Toaster
+ */
+class Toaster extends React.PureComponent {
```
mmcgahan: I think this class is probably going to need to be refactored, although it might require some testing to confirm. 

The main issue is that it currently expects _all_ toasts to be provided as `props` when the component is _first mounted_ - if new toasts arrive as new props later in the component's lifecycle, they will be ignored. You'll need to use `componentWillReceiveProps` to update `this.state.toasts` in some way

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+	constructor(props) {
+		super(props);
+
+		this.timeouts = [];
+		this.dismissedToasts = [];
```
mmcgahan: not used

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+		this.state = {
+			toasts: this.props.toasts.map(this.cloneToast)
+		};
+	}
+
+	/**
+	 * @param {Object} dismissedToast - `Toast` components to remove from state
+	 * @returns undefined
+	 */
+	setDismissedToast(dismissedToast) {
```
mmcgahan: nitpick - I recommend renaming this to `dismissToast` - it's not 'setting' anything

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+			toasts: this.state.toasts.filter(toast => dismissedToast.props.id !== toast.props.id)
+		});
+	}
+
+	/**
+	 * @returns undefined
+	 *
+	 * sets timers to delay each toast's dismissal
+	 */
+	setTimer() {
```
mmcgahan: I like the specificity of your `setDismissedToast` function that takes a `toast` as an argument - I think it would be nice to define this `setTimer` method the same way rather than building the `toast` looping into the method.

it could then be called like

```js
this.state.toasts.filter(t => t.props.autodismiss).map(this.setTimer);
```

I think a single-toast-at-a-time definition will also be required in order to handle new toasts coming in after mounting (you don't want to reset the timers on all the active toasts when a new one comes in)

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+		const toastsToDismiss = this.state.toasts.filter((toast) => toast.props.autodismiss);
+
+		toastsToDismiss.forEach((toast, i) => {
+			this.timeouts.push(setTimeout(() => {
+				this.setDismissedToast(toast);
+			}, DELAY_TIME*(i+1)));
```
mmcgahan: just throwing it out there - instead of multiplying the delay time, what do you think of having a shorter delay for each subsequent toast?, e.g. instead of the timeline for 3 toasts looking like

```
-------|
---------------|
-----------------------|
```

it would look like

```
-------|
----------|
-------------|
```

the equation would be something like

```js
const BASE_DELAY = 3000;
const MARGINAL_DELAY = 1000; // each additional toast will wait this much longer than the previous one
setTimeout(() => { ... }, BASE_DELAY + (MARGINAL_DELAY * i))
```

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+		});
+	}
+
+	/**
+	 * @returns undefined
+	 *
+	 * removes timers in order to prevent toasts from being dismissed
```
mmcgahan: ... in order to prevent toasts from _automatically_ dismissing?

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+			setDismissedToast: this.setDismissedToast
+		};
+		return React.cloneElement(toast, toastProps);
+	}
+
+	/**
+	 * @returns {Array} `Toast` components
+	 */
+	renderToasts() {
+		this.state.toasts = this.state.toasts.map(this.cloneToast);
```
mmcgahan: This seems strange to me - `this.state.toasts` is already a set of cloned toasts (from the constructor), so why clone them again?

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+				<CSSTransitionGroup
+					transitionAppear
+					transitionAppearTimeout={0}
+					transitionEnterTimeout={250}
+					transitionLeaveTimeout={250}
+					transitionName='slide'>
+					{
+						this.renderToasts().map((toast, i) => (
+							toast
+						))
```
mmcgahan: I don't think this `map` is doing anything - it's just returning the same array that `this.renderToasts` does.

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+					}
+				</CSSTransitionGroup>
+			</div>
+		);
+	}
+}
+
+Toaster.propTypes = {
+	toasts: PropTypes.arrayOf(PropTypes.element).isRequired
```
mmcgahan: probably needs to be `arrayOf(Toast)` right? not just any element?

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * @param {Object} Toast - `Toast` components to clone
+	 * @param {number} i - index of the `toast`
+	 * @returns {Array} `Toast` components with props from `Toaster`
+	 */
+	cloneToast(toast, i) {
+		const toastProps = {
+			key: i,
+			id: toast.props.id || i,
```
mmcgahan: integer ids are technically valid, but have a higher risk of conflicting with other ids on the page - i recommend a fallback more like

```js
`toast-${i}`
```

assets/scss/components/_toast.scss
```diff
@@ -0,0 +1,69 @@
...
+//easing - should come from `swarm-animations` once we add it here
+$easing--standard: cubic-bezier(.4, 0, .2, 1); // stays on screen from start to finish
+$easing--enter: cubic-bezier(0, 0, .2, 1); // entering the screen
+$easing--exit: cubic-bezier(.4, 0, 0, 1); // exiting the screen
+.slide-appear {
+  opacity: 0.01;
+  transform: translateY(-100px);
+}
+
+.slide-appear.slide-appear-active {
```
mperrotti: I would agree, and this would be possible if we upgrade to the newer React Transition Group

src/interactive/Toast.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+		return (
+			<span
+				className={classNames}
+				{...other}>
+				<Flex align='center'>
+					<FlexItem>{children}</FlexItem>
+					{action && (
+						<FlexItem shrink>
+							<a href='#' onClick={action} className={TOAST_ACTION_CLASS}>
```
mperrotti: Silly mistake, this should absolutely be a button. 😳

assets/scss/components/_toast.scss
```diff
@@ -0,0 +1,69 @@
...
+
+// durations - should come from `swarm-animations` once we add it here
+$duration--short: .1s;
+$duration--medium: .25s;
+$duration--long: .5s;
+
+//easing - should come from `swarm-animations` once we add it here
+$easing--standard: cubic-bezier(.4, 0, .2, 1); // stays on screen from start to finish
+$easing--enter: cubic-bezier(0, 0, .2, 1); // entering the screen
+$easing--exit: cubic-bezier(.4, 0, 0, 1); // exiting the screen
```
unjust: pretty sure these are in now so you dont need to redefine.
https://github.com/meetup/meetup-web-components/pull/271/files#diff-56442ae3941b6e3cf1c61be73278bc30R20

assets/scss/components/_toast.scss
```diff
@@ -0,0 +1,69 @@
...
+.slide-leave {
+  opacity: 1;
+  transform: translateY(0px);
+}
+
+.slide-leave.slide-leave-active {
+  opacity: 0.01;
+  transform: translateY(-100px);
+  transition: opacity $duration--medium $easing--exit, transform $duration--medium $easing--exit;
+}
```
unjust: I put in a ticket to maybe later pull `slide` into the animation lib. we can discuss. https://meetup.atlassian.net/browse/SDS-419

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,150 @@
...
+	 * @returns undefined
+	 *
+	 * removes timers in order to prevent toasts from being automatically dismissed
+	 */
+	restartTimeouts() {
+		const toasts = this.state.toasts;
+
+		toasts.forEach((toast, i) => {
+			this.setTimer(toast);
+		});
```
mmcgahan: `toasts.forEach(this.setTimer)`

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+	componentDidMount() {
+		this.state.toasts && this.state.toasts.filter(t => t.props.autodismiss).map(this.setTimer);
+	}
+
+	componentWillReceiveProps(nextProps) {
+		const currentToasts = this.state.toasts;
+		const allToasts = currentToasts.concat(nextProps.toasts);
+
+		console.log('--------------------componentWillReceiveProps fired--------------------');
```
mmcgahan: 🔪

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+	/**
+	 * @param {Object} Toast - `Toast` components to clone
+	 * @param {number} i - index of the `toast`
+	 * @returns {Array} `Toast` components with props from `Toaster`
+	 */
+	cloneToast(toast, i) {
+		const toastProps = {
+			key: i,
+			id: toast.props.id || `toast-${i}`,
```
mmcgahan: just realized that this `id` scheme won't work - since `cloneToast` is called in the constructor, the first toast will get `#toast-0`. If a new toast is added in `componentWillReceiveProps`, it will _also_ get `#toast-0`.

I think you can/should get rid of any references to the toast `id`. It looks like it's only used for `this.dismissToast`, and I think you can just use an object equality check rather than an ID check for that, since `dismissToast` takes a `toast` that should be a reference to the same thing that is in `state`

```js
dismissToast(dismissedToast) {
  toasts: this.state.toasts.filter(toast => dismissedToast !== toast) })
}
```

src/interactive/Toaster.jsx
```diff
@@ -0,0 +1,150 @@
...
+				onMouseLeave={this.restartTimeouts}
+				{...other}>
+				<TransitionGroup>
+					{
+						this.state.toasts.map((toast, i) => (
+							<CSSTransition
+								key={`csstrns-${i}`}
+								appear
+								timeout={250}
+								classNames="slide">
```
mmcgahan: can you put this closing `>` on a newline? A scan of this JSX makes it look like `{toast}` might be a prop

PR #273, opened 2017-08-21
WC-81: Move Redux Form Textarea into MWC

aquillano:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-81

### DIFF
--- a/src/forms/redux-form/textarea.test.jsx
+++ b/src/forms/redux-form/textarea.test.jsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import ReduxFormTextarea from './Textarea';
+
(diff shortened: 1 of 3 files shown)


src/forms/redux-form/textarea.test.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import ReduxFormTextarea from './Textarea';
+
+describe('redux-form Textarea', () => {
+
+	const formAttrs = {
+		input: {
```
unjust: nitpick: maybe make these attributes more suitable for a textarea? 
maxLength 20 is short.

aquillano: 🤷‍♂️

PR #276, opened 2017-08-22
Sds 411 popover render bug

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-411

#### Description
Links were being rendered, with `opacity--0`, for the popover menu., and were still clickable.
This fix uses the `a11y` classes  to position the menu offscreen. 

I needed to make an adjustment to the a11y `static` positioning.

(**another solution could be not to render the menu items if `!isActive`.)

### DIFF
--- a/assets/scss/components/_popover.scss
+++ b/assets/scss/components/_popover.scss
@@ -33,6 +33,10 @@ $view-minWidth: 320px;
 
 .popover-container--menu {
 	min-width: $block-4;
+
+	&.visibility--a11yShow {
+		position:absolute;
+	}
 }
 
 .popover-menu-option {
--- a/src/interactive/Popover.jsx
+++ b/src/interactive/Popover.jsx
@@ -160,10 +160,10 @@ class Popover extends React.Component {
 				'popover-container',
 				'popover-container--menu',
 				{
-					'trans-fadeIn--short': !isActive,
+					'trans-fadeIn': !isActive,
 					'opacity--0': !isActive,
+					'trans-fadeOut': isActive,
 					'opacity--1': isActive,
-					'trans-fadeOut--short': isActive,
 					'popover-container--horizontal-left': (align === 'left'),
 					'popover-container--horizontal-right': (align === 'right')
 				}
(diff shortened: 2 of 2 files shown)


assets/scss/components/_popover.scss
```diff
@@ -33,6 +33,10 @@ $view-minWidth: 320px;
 
 .popover-container--menu {
 	min-width: $block-4;
+
+	&.visibility--a11yShow {
+		position:absolute;
+	}
```
unjust: without this, the menu was `position:static` and took up more width than it should.

unjust:
@eilinora @mperrotti - storybook looks good, but what would be a good way to test this on `mup-web`.

mperrotti:
@unjust - you can use the beta release. Just bump the `meetup-web-components` version to `3.0.1176-beta`

src/interactive/Popover.jsx
```diff
@@ -160,10 +160,12 @@ class Popover extends React.Component {
 				'popover-container',
 				'popover-container--menu',
 				{
-					'trans-fadeIn--short': !isActive,
+					'trans-fadeIn': !isActive,
```
mperrotti: Stupid nitpick...there's a lot of classnames here, so writing them like this might help it look less overwhelming:
```
'trans-fadeIn opacity--0 visibility--a11yHide': !isActive,
'trans-fadeOut opacity--1 visibility--a11yShow': isActive,
'popover-container--horizontal-left': (align === 'left'),
'popover-container--horizontal-right': (align === 'right')
```

unjust: ok great will do!

assets/scss/components/_popover.scss
eilinora: can we added a note in the css maybe as to why we are doing? not 💯 obvious why we are, just a suggestion definitely not a blocker

eilinora:
Overall lgtm, agree with everything Mike said. I think once you update with his comments. Also a agree with him that a little movement would be nice when it opens closes (like a subtle "reveal"). But can be done as enhancement.

PR #277, opened 2017-08-23
Wc 82 time input redux

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-82

#### Description
Time input currently manages its own state and updates it to work with DateTimePicker, but this adds complexity to the component and makes it difficult to work with redux form.

In prep for working with redux form, we should create a redux form wrapper and remove state inside time input and opt for its value coming from props.

screen shot 2017-08-23 at 10 50 35 am

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -56,6 +56,7 @@ class TimeInput extends React.Component {
 					value={value}
 					className={classNames}
 					required={required}
+					onChange={this.onChange}
 					ref={ input => this.inputEl = input }
 					{...other}
 				/>
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -21,7 +21,9 @@ class TimeInput extends React.Component {
 	*	which may be provided by a parent component such as DateTimePicker
 	*/
 	onChange(e) {
-		this.props.onChange && this.props.onChange(e.target.value);
+		this.props.onChange && this.props.onChange(e.target.value); // redux-form provides an onChange prop
+
+		// onChangeCallback currently provided by a parent component eg DateTimePicker
 		this.props.onChangeCallback && this.props.onChangeCallback(e.target.value);
 	}
 
@@ -85,7 +87,8 @@ TimeInput.propTypes = {
 		PropTypes.element
 	]),
 	required: PropTypes.bool,
-	onChangeCallback: PropTypes.func,
+	onChange: PropTypes.func,			// redux-form provides an onChange prop
+	onChangeCallback: PropTypes.func, 	// currently provided by a parent component eg DateTimePicker
 };
 
 export default TimeInput;
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -9,14 +9,16 @@ class TimeInput extends React.Component {
 
 	constructor(props) {
 		super(props);
-		// onChange comment
+
 		this.onChange = this.onChange.bind(this);
 	}
 
 	/**
 	* @function onChange
 	* @param e Event Object
-	* @description ...
+	* @description called when the input changes, in turn calls the onChange
+	* 	handler prop, if there is one (eg supplied by redux-form) and an onChangeCallback
+	*	which may be provided by a parent component such as DateTimePicker
 	*/
 	onChange(e) {
 		this.props.onChange && this.props.onChange(e.target.value);
@@ -46,6 +48,14 @@ class TimeInput extends React.Component {
 			'label--field',
 			{ required }
 		);
+
+		const errorId = `${id}-error`;
+
+		if (error) {
+			other['aria-invalid'] = true;
+			other['aria-describedby'] = errorId;
+		}
+
 		return (
 			<span>
 				{ label && <label htmlFor={id} className={labelClassNames}>{label}</label> }
@@ -60,7 +70,7 @@ class TimeInput extends React.Component {
 					ref={ input => this.inputEl = input }
 					{...other}
 				/>
-				{ error && <p className='text--error'>{error}</p> }
+				{ error && <p id={errorId} className='text--error'>{error}</p> }
 			</span>
 		);
 
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -9,26 +9,31 @@ class TimeInput extends React.Component {
 
 	constructor(props) {
 		super(props);
-		this.state = {
-			value: this.props.value || ''
-		};
+		// onChange comment
 		this.onChange = this.onChange.bind(this);
 	}
 
+	/**
+	* @function onChange
+	* @param e Event Object
+	* @description ...
+	*/
 	onChange(e) {
-		this.setState({ value: e.target.value });
+		this.props.onChange && this.props.onChange(e.target.value);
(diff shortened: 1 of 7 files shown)


src/forms/TimeInput.jsx
```diff
@@ -9,26 +9,33 @@ class TimeInput extends React.Component {
...
+	* @function onChange
+	* @param e Event Object
+	* @description called when the input changes, in turn calls the onChange
+	* 	handler prop, if there is one (eg supplied by redux-form) and an onChangeCallback
+	*	which may be provided by a parent component such as DateTimePicker
+	*/
 	onChange(e) {
-		this.setState({ value: e.target.value });
+		this.props.onChange && this.props.onChange(e.target.value);
 		this.props.onChangeCallback && this.props.onChangeCallback(e.target.value);
```
mmcgahan: `onChange` isn't in the prop-types, and it's not really clear why both are needed - can we just always use `onChange` regardless of where it comes from?

src/forms/TimeInput.jsx
```diff
@@ -9,26 +9,33 @@ class TimeInput extends React.Component {
...
-			onChangeCallback,	// eslint-disable-line no-unused-vars
 			id,
 			label,
 			name,
 			className,
 			required,
-			value,		// eslint-disable-line no-unused-vars
+			value,
+			error,
+			onChange,			// eslint-disable-line no-unused-vars
```
sadafie: What's the difference between `props.onChange` and `props.onChangeCallback`? Slightly confusing. Also, I only see `onChangeCallback` list on the propTypes object below.

src/forms/TimeInput.jsx
```diff
@@ -41,35 +48,44 @@ class TimeInput extends React.Component {
...
-	onChangeCallback: PropTypes.func,
 	name: PropTypes.string.isRequired,
 	error: PropTypes.string,
 	label: PropTypes.oneOfType([
 		PropTypes.string,
 		PropTypes.element
 	]),
-	required: PropTypes.bool
+	required: PropTypes.bool,
+	onChangeCallback: PropTypes.func,
```
sadafie: should `onChange` also be a prop here?

src/forms/redux-form/TimeInput.jsx
```diff
@@ -0,0 +1,18 @@
...
+ * @param {object} props React component props
+ * @return {React.Component} TimeInput
+ */
+const ReduxFormTimeInput = props => {
+	const { input, meta, ...other } = props;
+
+	return <TimeInput {...input} error={meta.error} {...other} />;
+};
+
+ReduxFormTimeInput.displayName = 'ReduxFormTimeInput';
```
sadafie: Just curious why the displayName is needed.

src/forms/timeInput.test.jsx
```diff
@@ -1,58 +1,77 @@
...
+	const callbackSpy = jest.fn(),
+		onChangeMock = jest.fn();
 
 	beforeEach(() => {
-		timeInputComponent = TestUtils.renderIntoDocument(
-			<TimeInput name='time' value={timeValue} onChangeCallback={callbackSpy} required />);
-		timeInputEl = TestUtils.findRenderedDOMComponentWithTag(timeInputComponent, 'input');
+		component = shallow(
+			<TimeInput
+				name='time'
```
sadafie: A slight nitpick, but one sort of consistent approach I've seen (at least in `mup-web`) is to create an object of props and use the spread operator instead of individual vars. i.e., 

```
const DEFAULT_PROPS = {
    name: 'time',
    value: '22:00',
    onChange: jest.fn(),
    required: 'required',
};
shallow(<TimeInput {...DEFAULT_PROPS} />);
```

src/forms/timeInput.test.jsx
```diff
@@ -1,58 +1,77 @@
...
+		inputEl = null;
 	});
 
-	it('exists', function() {
-		expect(timeInputEl).not.toBeNull();
+	it('renders a time html input with expected props', function() {
+		expect(component).toMatchSnapshot();
 	});
 
 	it('takes a value in HH:mm format', function() {
```
sadafie: the snapshot is explicit about this case, so you shouldn't need a separate unit test for it.

src/forms/timeInput.test.jsx
```diff
@@ -1,58 +1,77 @@
...
-		expect(onChangeSpy).toHaveBeenCalled();
-		expect(setStateSpy).toHaveBeenCalledWith({ value: newTime });
-	});
+		afterEach(() => {
+			inputEl = null;
+			onChangeSpy.mockClear();
+		});
+
+		it('calls internal onChange when value is changed', function() {
+			inputEl.simulate('change', { target: { value: newTime } });
```
sadafie: since this simulate is the same across all cases, you can move this to the `beforeEach()`

src/forms/TimeInput.jsx
unjust: I can put `onChange` in the proptypes. `onChange` prop comes from the redux form `Field` component and we want this called if present. `onChangeCallback` is called if the parent component provides it (eg DateTimePicker).
I put an explanation in the header doc.

src/forms/TimeInput.jsx
unjust: `onChange` prop comes from the redux form `Field` component. `onChangeCallback` is called if the parent component provides it (eg DateTimePicker). 
I can rename it in a later branch when I work with dateTimePicker which provides that prop.
I put an explanation in the header doc of `onChange` on the difference between the two.

src/forms/TimeInput.jsx
unjust: yes I guess I should add it.

src/forms/redux-form/TimeInput.jsx
unjust: for debugging in the console I believe. 
the file name of redux form components are just `redux-form/TimeInput.jsx`
mike spencer gave me that feedback before that we should have a displayName.

src/forms/TimeInput.jsx
mmcgahan: it sounds like they could be consolidated - the parent `Field` can provide `onChange` when it's used with redux-form, and a different parent can provide `onChange` in other contexts, e.g. DateTimePicker - are there cases where you would need to support both simultaneously?

unjust: Honestly I'm not sure. 
Right now I know it works when both are preset, but its possible I need only one. 
I'm currently working out the best way these behave with redux form (both within the date picker, or separately). Can I just define the two, and remove in a follow up if I can verify in `mup-web` that only one is actually needed?

src/forms/timeInput.test.jsx
unjust: ok I'll do something like that.

src/forms/TimeInput.jsx
mmcgahan: I'm not sure where `TimeInput` is used in mup-web - it's not directly used in `master` - so I'm wondering what the risk is with just updating the name of `onChangeCallback` to `onChange` in order to avoid tech debt?

unjust: Decision: removing the DateTimePicker `onChangeCallback` in favor of one `onChange` prop. When I try building a proper `redux-form` DateTimePicker in `mup-web` on event create, if I see they are both still necessary, I will add back in a follow up branch. ccing: @aquillano

PR #278, opened 2017-08-23
Wc 83 datetimelocal redux

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-83

#### Description
Datetimelocal input currently manages its own state and updates it to work with DateTimePicker, but this adds complexity to the component and makes it difficult to work with redux form.

In prep for working with redux form, we should create a redux form wrapper and remove state inside the component and opt for its value coming from props. Also needs a story.

screen shot 2017-08-23 at 11 33 11 am


src/forms/DateTimeLocalInput.jsx
```diff
@@ -71,7 +75,7 @@ DateTimeLocalInput.propTypes = {
 		PropTypes.string,
 		PropTypes.element,
 	]),
-	onChangeCallback: PropTypes.func
+	onChange: PropTypes.func, // provided by redux-form or by a wrapping component eg DateTimePicker
```
mmcgahan: since you're not setting the `value` based on internal `this.state` any more, I think `onChange` would be a `isRequired` property, and the parent _must_ update the `value` passed in as a `prop` in the onChange? If that's right, it would be good to document that implicit requirement - if the passed-in `onChange` prop _doesn't_ update `value`, then I think this component would lose its value every time it re-rendered

unjust:
With the last commit I added back `state`, so this could function as a (self)controlled component without `redux-form` or parent controlling the value.
But it could still function fine if it were stateless and provided a `defaultValue` instead of `value`.

src/forms/DateTimeLocalInput.jsx
unjust: after the conversation - do you think `onChange` should still be a required prop?

unjust:
also - are these changes breaking changes and should I update the version?  fwiw I don't think any parts of the app are using this input yet. its embedded in `DateTimePicker`

src/forms/DateTimeLocalInput.jsx
```diff
@@ -4,40 +4,36 @@ import cx from 'classnames';
 
 /**
  * @module DateTimeLocalInput
- * @description renders html5 datetime-local input
+ * @description renders html5 datetime-local input, 
+ * this component is supported on mobile browsers mostly, and should be used when supported
```
aquillano: @unjust How/where do we sniff if it is supported? What's the fallback?

unjust: the fallback test is in DateTimePicker right now.

aquillano:
Done looking at this, nothing else to add.

src/forms/DateTimeLocalInput.jsx
```diff
@@ -4,40 +4,36 @@ import cx from 'classnames';
...
-	* (callback used in wrapping components like datetimepicker)
+	* @function handleChange
 	* @param e Event the change event
+	* @description called when value changes and in turn calls onChange from props 
+	* (redux-form or wrapping components like datetimepicker provides this)
 	*/
-	onChange(e) {
-		this.setState({ value: e.target.value });
-		this.props.onChangeCallback && this.props.onChangeCallback(e.target.value);
+	handleChange(e) {
```
unjust: hrm. not quite sure what you mean with ‘how its used in the jsx’ here. I thought itd be easier to distinguish the internal `handleChange` from the `onChange` prop provided if they have different names.

mmcgahan: yeah, I think I wouldn't bother distinguishing them further - parent components will only be aware of the `onChange` that they pass in, and internally it's relatively easy to see the distinction between `this.onChange` and `this.props.onChange`, but that's definitely subjective so no need to change (ha ha).

mmcgahan:
Your call on marking as a breaking change and bumping version number in the `Makefile`. I don't think you need to worry about it, though.

src/forms/DateTimeLocalInput.jsx
unjust: easy revert so...

PR #279, opened 2017-08-23
Wc 84 calendar redux

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-84

#### Description
Calendar component, which uses flatpickr, currently manages its own state and updates it to work with DateTimePicker, but this adds complexity and makes it difficult to work with redux form.

In prep for working with redux form, we should create a redux form wrapper , remove any unnecessary state inside the component and opt for its value coming from props.
Also needs a story.

screen shot 2017-08-23 at 11 43 13 am


src/forms/CalendarComponent.jsx
```diff
@@ -38,25 +36,30 @@ class CalendarComponent extends React.Component {
 			defaultDate: this.props.value
 		};
 
-		Object.assign(options, this.props.opts);
+		Object.assign(options, this.props.datepickerOptions);
```
mmcgahan: I see that it's existing code, but there's generally no need for `Options.assign` when you can use object destructuring - this is the same as adding `...this.props.datepickerOptions` as the last line of the `options` definition.

src/forms/CalendarComponent.jsx
```diff
@@ -38,25 +36,30 @@ class CalendarComponent extends React.Component {
...
 	}
 
 	componentWillUnmount() {
 		this.flatpickr && this.flatpickr.destroy();
 	}
 
+	// replaces updateFlatpickr
+	// if we receive a new value from parent, update Flatpickr
+	componentWillReceiveProps(newProps) {
+		this.flatpickr.setDate(newProps.value);
```
mmcgahan: `componentWillUnmount` and the old `updateFlatpickr` method check for a falsey `this.flatpickr` - should probably do the same here unless there's a guarantee that `this.flatpickr` is always available, in which case the falsey check can be removed from `componentWillUnmount`

src/forms/calendarComponent.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+		afterEach(() => {
+			calendarComponent = null;
+		});
+
+		// because we are rendering shallow here, flatpickr isnt
+		// instantiated (because componentDidMount not called)
+		// just using direct calls to flatpickr change to test
+
+		it('calls onChange prop if one is provided, as with redux-form', function() {
```
mmcgahan: since you only have one `it()` test, I recommend _not_ using `beforeEach`/`afterEach` - they are generally recommended only when they save you a lot of setup/teardown code for many different unit tests, so in this case it would be better to put everything inside the `it()` - then you won't have to deal with the complex parent scope issues.

src/forms/calendarComponent.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+			calendarComponent.instance().onFlatpickrChange([newValue]);
+			expect(onChangePropMock).toHaveBeenCalledWith(newValue);
```
mmcgahan: in general, if you're testing that something happened as the result of a specific call (not just a render), you should check that it _has not_ happened _before_ the specific call. E.g. in this case you can't really be certain that `onChangePropMock` was called as a result of `onFlatpickrChange` - you just know that it was called sometime after it was set to `jest.fn()`.

src/forms/calendarComponent.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+		afterEach(() => {
+			cal = null;
+			onFocusMock.mockClear();
+			onBlurMock.mockClear();
+		});
+
+		it('onOpen calls the props onFocus method when it exists', () => {
+			cal.instance().onOpen();
+			expect(onFocusMock).toHaveBeenCalled();
```
mmcgahan: same - make sure you call `expect(onFocusMock).not.toHaveBeenCalled();` before line 105.

I'd also recommend making the render call inside each `it()` in order to avoid `beforeEach`/`afterEach`. DRY isn't as applicable to unit tests, which are ideally fully self-contained.

src/forms/calendarComponent.test.jsx
unjust: 👍

unjust:
addressed the recommendations so I'm merging

PR #281, opened 2017-08-24
Wc 86 textinput storybook fix

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-86

#### Description
TextInput stories that have a value are non-interactive (you cant type in the field). defaultValue should be passed when possible. 
The exception is the charcounter story, that needs a value (usually given by parent) to display the `value.length`. I put a note in the infowrapper as to why this is non-interactive.


aquillano:
LGTM

PR #285, opened 2017-08-31
Remove datetimepicker component files

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-87

#### Description
DateTimePicker was developed outside of redux-form for a previous design of Event Schedule.

The new design has changed the way this component looks, and the complexities of this with redux form require its rethinking. 

Right now its best to use the pieces (Calendar, TimeInput, Datetimelocal) to create the compound component you need for your app.


unjust:
Let me know if I need to bump the version. AFAIK no one is using this component in prod.

PR #292, opened 2017-09-12
Wc 88 calendar style changes

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-88

#### Description
style changes for our calendar component. see ticket for specs.

screen shot 2017-09-12 at 3 52 11 pm

### DIFF
--- a/assets/scss/components/_calendar.scss
+++ b/assets/scss/components/_calendar.scss
@@ -1,1 +1,55 @@
-@import "~flatpickr/dist/flatpickr"; 
+@import "~flatpickr/dist/flatpickr";
+
+/* meetup overrides */
+
+.flatpickr-calendar {
+	padding: $space-half;   
+	width: auto;
+
+	.svg--chevron-right svg {
+		fill: $C_textPrimary;
+	}
+
+	.svg--chevron-left svg {
+		fill: $C_separator;
+	}
+}
+
+span.flatpickr-weekday {
+	color: inherit;
+	font-weight: $W_bold;
+  }
+
+.flatpickr-current-month {
+	width:80%;
+	& span.cur-month {
+		font-weight: $W_normal;
+	}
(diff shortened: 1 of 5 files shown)


assets/scss/components/_calendar.scss
```diff
@@ -1 +1,55 @@
...
+  }
+
+.flatpickr-current-month {
+	width:80%;
+	& span.cur-month {
+		font-weight: $W_normal;
+	}
+}
+
+.flatpickr-day.selected,
```
mperrotti: This CSS is a little crazy. Is there a shorter/better way to select all of these things?

src/forms/CalendarComponent.jsx
```diff
@@ -32,8 +32,18 @@ class CalendarComponent extends React.Component {
 			onClose: this.onClose,
 			altInput: true,
 			allowInput: true,
-			altFormat: 'D M d, Y', // TODO localize
+			altFormat: 'D M d, Y',
 			defaultDate: this.props.value,
+			nextArrow: `<span class="svg svg--chevron-right">
```
mperrotti: Is there a reason we can't just render the `<Icon>` component instead of manually adding the markup?

unjust: these config options for `nextArrow` and `prevArrow` expect `html`
https://chmln.github.io/flatpickr/options/

so I don't think the component will work. I'm pretty sure I tried it a while back.

assets/scss/components/_calendar.scss
unjust: I know. But I took this selector right from their stylesheet. I imagine its like this for a reason.

unjust:
@mperrotti didn't follow up and approve but we discussed and I think he verbally approved.

PR #298, opened 2017-09-13
fix for safari on ios

jaclynj:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-1692

#### Description
Clicking out of popover menu was not working in Safari mobile because of an issue with Safari mobile where `click` events are not fired on typically non-interactive elements. Reference: https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile

### DIFF
--- a/src/interactive/Popover.jsx
+++ b/src/interactive/Popover.jsx
@@ -128,6 +128,10 @@ class Popover extends React.Component {
 	}
 
 	componentDidMount() {
+		// fix for safari on ios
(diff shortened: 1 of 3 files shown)


src/interactive/Popover.jsx
```diff
@@ -128,6 +128,10 @@ class Popover extends React.Component {
 	}
 
 	componentDidMount() {
+		// fix for safari on ios
+		if( navigator.userAgent.match(/(iPad|iPhone|iPod)/i) ) {
```
eilinora: tests :)

PR #302, opened 2017-09-22
AvatarMember - Support thumbnails on small avatars

nlaz:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-424

#### Description
This PR updates the `AvatarMember` component to support thumbnails on small sizes in order to optimize for the smaller sized images.

#### Screenshots
screen shot 2017-09-22 at 2 46 47 pm

### DIFF
--- a/src/media/AvatarMember.jsx
+++ b/src/media/AvatarMember.jsx
@@ -14,9 +14,9 @@ export const AVATAR_PERSON_NOPHOTO_CLASS = 'avatar--noPhoto';
  */
 class AvatarMember extends React.PureComponent {
 	render() {
-		const { member, org, fbFriend, className, small, ...other } = this.props;
+		const { member, org, fbFriend, className, big, large, xxlarge, ...other } = this.props;
 
-		const photoLink = small ? 'thumb_link' : 'photo_link';
+		const photoLink = big || large || xxlarge ? 'photo_link' : 'thumb_link';
 		const showNoPhoto = (member.photo || {})[photoLink] == undefined;
 
 		const classNames = cx(
@@ -36,7 +36,9 @@ class AvatarMember extends React.PureComponent {
 				alt={member.name}
 				src={showNoPhoto ? noPhotoImage : member.photo[photoLink]}
 				className={classNames}
-				small={small}
+				big={big}
+				large={large}
+				xxlarge={xxlarge}
 				{...other}
 			/>
 		);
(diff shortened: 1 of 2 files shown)


src/media/AvatarMember.jsx
```diff
@@ -14,22 +14,17 @@ export const AVATAR_PERSON_NOPHOTO_CLASS = 'avatar--noPhoto';
...
-			member,
-			org,
-			fbFriend,
-			className,
-			...other
-		} = this.props;
+		const { member, org, fbFriend, className, small, ...other } = this.props;
 
-		const showNoPhoto = (member.photo || {}).photo_link == undefined;
+		const photoLink = small ? 'thumb_link' : 'photo_link';
```
eilinora: I feel like we should make the default the small one as this it the most likely way it will be used.

nlaz: @eilinora when should the `photo_link` be used then? On `large` versions?

eilinora: Yeah there are a couple of sizes... so large and above :)

nlaz: Hmm yeah that's probably a good idea. I'll wait for @mperrotti to be back to run this by him since it's a sorta small but still breaking change.

mperrotti: @nlaz - All we're doing is asking for the smaller image asset for smaller avatars?
If so, LGTM :)

src/media/AvatarMember.jsx
```diff
@@ -39,20 +34,24 @@ class AvatarMember extends React.PureComponent {
 		return (
 			<Avatar
 				alt={member.name}
-				src={showNoPhoto ? noPhotoImage : member.photo.photo_link}
+				src={showNoPhoto ? noPhotoImage : member.photo[photoLink]}
 				className={classNames}
+				big={big}
```
eilinora: we shouldn't need to specify these as they _should_ come through with `...other` did you observe something different?

nlaz: Yeah I saw something different. They don't come through because of the destructuring statement `const { ... , big, large, xxlarge, ...other } = this.props;`. There's a more clear example here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

nlaz: But I could do something like:

```
const { member, org, fbFriend, className, ...other } = this.props;
const { big, large, xxlarge } = other;
```
How does that feel to you?

eilinora: yes yes your right! leave as is

PR #305, opened 2017-10-03
Added a ToggleSwitch component

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-103_add_toggleSwitch

#### Description
Adding a form input that looks like a native mobile toggle switch

screen shot 2017-10-03 at 9 22 15 am
screen shot 2017-10-03 at 9 22 10 am

### DIFF
--- a/assets/scss/components/_toggleSwitch.scss
+++ b/assets/scss/components/_toggleSwitch.scss
@@ -1,0 +1,67 @@
+$trackWidth: $media-l;
+$knobSize: $trackWidth/2;
+
+.toggleSwitch {
+	background-color: $C_coolGrayLight;
+	border-radius: 999px;
+	height: $media-s;
+	position: relative;
+	width: $media-l;
+
+	&:hover {
+		background-color: darken($C_coolGrayLight, 5%);
+	}
+
+	&:focus {
+		background-color: $C_coolGrayLight;
+	}
+
+	&:active {
+		background-color: darken($C_coolGrayLight, 10%);
+	}
+}
+
+.toggleSwitch--active {
+	background: $C_darkGray;
+
+	&:hover {
+		background-color: darken($C_darkGray, 5%);
+	}
+
+	&:focus {
+		background-color: $C_darkGray;
+	}
+
+	&:active {
+		background-color: darken($C_darkGray, 10%);
+	}
+}
+
+.button.toggleSwitch--disabled[disabled] {
+	background: $C_lightGray;
+
+	&:hover, &:focus, &:active {
+		background-color: $C_lightGray;
+	}
+}
+
+.toggleSwitch-knob {
+	background: white;
+	border-radius: 999px;
+	box-shadow: 0 0 2px $C_coolGrayMediumTransp;
+	height: $knobSize;
+	left: 0;
+	position: absolute;
+	top: 0;
+	transform: translateX(0);
+	transition: transform $duration--short $easing--standard;
+	width: $knobSize;
+}
+
+.toggleSwitch-knob--active {
+	transform: translateX($knobSize);
+}
+
+.toggleSwitch-knob--disabled {
+	box-shadow: 0 0 2px $C_border;
+}
--- a/src/forms/ToggleSwitch.jsx
+++ b/src/forms/ToggleSwitch.jsx
@@ -29,12 +29,18 @@ class ToggleSwitch extends React.Component {
 	}
 
 	toggle() {
-		!this.props.disabled && this.props.toggleActive();
+		if (this.props.disabled) {
+			return;
+		} else {
+			this.props.toggleActive();
+			this.props.isActive
+				? this.props.offCallback && this.props.offCallback()
+				: this.props.onCallback && this.props.onCallback();
+		}
 	}
 
 	render() {
 		const {
-			children,
 			className,
 			disabled,
 			isActive,
@@ -43,6 +49,8 @@ class ToggleSwitch extends React.Component {
 			labelledBy,
 			name,
 			toggleActive, // eslint-disable-line no-unused-vars
+			onCallback, // eslint-disable-line no-unused-vars
+			offCallback, // eslint-disable-line no-unused-vars
 			...other
 		} = this.props;
 
@@ -99,7 +107,6 @@ class ToggleSwitch extends React.Component {
 							label='Toggle switch label'/>
 					</span>
 				</Button>
-				{children}
 			</div>
 		);
 	}
--- a/src/forms/ToggleSwitch.jsx
+++ b/src/forms/ToggleSwitch.jsx
@@ -32,15 +32,16 @@ class ToggleSwitch extends React.Component {
 		!this.props.disabled && this.props.toggleActive();
 	}
 
-	componentWillReceiveProps(nextProps) {
-		nextProps.isActive
-			? this.props.onCallback && this.props.onCallback()
-			: this.props.offCallback && this.props.offCallback();
+	componentDidUpdate(prevProps) {
+		if (prevProps !== this.props) {
+			this.props.isActive
+				? this.props.onCallback && this.props.onCallback()
+				: this.props.offCallback && this.props.offCallback();
+		}
 	}
 
 	render() {
 		const {
-			children,
 			className,
 			disabled,
 			isActive,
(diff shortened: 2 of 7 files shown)


src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,119 @@
...
+ */
+class ToggleSwitch extends React.Component {
+	constructor (props) {
+		super(props);
+
+		this.toggle = this.toggle.bind(this);
+	}
+
+	toggle() {
+		!this.props.disabled && this.props.toggleActive();
```
aquillano: The name of this function is confusing to me. I see that comes from `WithToggleControl` so maybe that's not easily changed.

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,119 @@
...
+			className,
+			disabled,
+			isActive,
+			label,
+			labelClassName,
+			labelledBy,
+			name,
+			toggleActive, // eslint-disable-line no-unused-vars
+			...other
+		} = this.props;
```
aquillano: Seems like this could use additional props, like `onCallback` and `offCallback` that are called in `componentWillReceiveNewProps` (maybe?)  when the state of the switch changes. Thinking about the need to expand/collapse an `<AccordionPanel />` as the state of the `<ToggleSwitch />` is changed.

aquillano: If you think that's a correct observation, I don't think it matters if that change happens in this PR or in the `Accordion` update PR.

assets/scss/components/_toggleSwitch.scss
```diff
@@ -0,0 +1,67 @@
...
+	&:focus {
+		background-color: $C_darkGray;
+	}
+
+	&:active {
+		background-color: darken($C_darkGray, 10%);
+	}
+}
+
+.button.toggleSwitch--disabled[disabled] {
```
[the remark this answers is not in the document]
mperrotti: You're right, but I'm doing it this way to beat the `[disabled]` styles that come from `button--reset`

src/forms/ToggleSwitch.jsx
mperrotti: It's definitely not the easiest thing to change at this point, but if you thing it's really that confusing, we can open a separate ticket

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,119 @@
...
+				className
+			),
+			knob: cx(
+				`${KNOB_CLASS} flex flex--center flex--alignCenter`,
+				{
+					[KNOB_ACTIVE_CLASS]: isActive,
+					[KNOB_DISABLED_CLASS]: disabled
+				}
+			),
+			label: cx(
```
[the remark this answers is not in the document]
mperrotti: That works for me

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,119 @@
...
+				>
+					<span className={classNames.knob}>
+						<Icon
+							shape={iconShape}
+							color={disabled ? C_TEXT_HINT : activeIconColor}
+							size='xxs'
+							label='Toggle switch label'/>
+					</span>
+				</Button>
+				{children}
```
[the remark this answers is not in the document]
mperrotti: I had the same thought myself and was going to remove this, but kept it in just incase somebody ever needed to pass children in for some weird unforeseen reason.
Will remove

src/forms/ToggleSwitch.jsx
mperrotti: Good idea, I'll add something to support this

src/forms/ToggleSwitch.jsx
aquillano: Don't mess with it.

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,127 @@
...
+		super(props);
+
+		this.toggle = this.toggle.bind(this);
+	}
+
+	toggle() {
+		!this.props.disabled && this.props.toggleActive();
+	}
+
+	componentWillReceiveProps(nextProps) {
```
aquillano: Now I'm wondering if putting this in `componentDidUpdate` -- it still receives `prevProps`. Also, either lifecycle function will get call many times so I think you should be comparing the current and past state of `this.props.isActive` and only calling the appropriate callback when isActive has actually changed. It will prevent a callback from getting called multiple times.

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+		this.toggle = this.toggle.bind(this);
+	}
+
+	toggle() {
+		!this.props.disabled && this.props.toggleActive();
+	}
+
+	componentDidUpdate(prevProps) {
+		if (prevProps !== this.props) {
```
aquillano: This is getting into the gray area of my React knowledge... Couldn't multiple props have changed? I guess in this component we'd only expect `isActive` to be changing, but I still wonder if `prevProps.isActive !== props.isActive` would be mo' better on any level?

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,123 @@
...
+						className={`${LABEL_CLASS} ${labelClassName} margin--right`}
+						id={labelId}
+					>
+						{label}
+					</span>
+				}
+				<Button
+					reset
+					disabled={disabled}
+					onClick={this.toggle}
```
unjust: small nitpick - `toggle` is confusing cos its a noun and a verb. `this.onClick` ? you can ignore me.

src/forms/toggleSwitch.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+import { shallow, mount } from 'enzyme';
+
+const mockOnCallback = jest.fn();
+const mockOffCallback = jest.fn();
+
+import ToggleSwitch, {
+	TOGGLE_SWITCH_ACTIVE_CLASS,
+	TOGGLE_SWITCH_DISABLED_CLASS,
+	TOGGLE_SWITCH_CLASS,
+	// KNOB_CLASS,
```
unjust: ✂️

src/forms/ToggleSwitch.jsx
```diff
@@ -0,0 +1,123 @@
...
+		const activeIconColor = isActive ? C_TEXT_PRIMARY : C_TEXT_SECONDARY;
+
+		return (
+			<div {...other}>
+				{ label &&
+					<span
+						className={`${LABEL_CLASS} ${labelClassName} margin--right`}
+						id={labelId}
+					>
+						{label}
```
unjust: why not just a `label` el?

src/forms/toggleSwitch.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+			expect(shallowWrapper()).toMatchSnapshot();
+		});
+
+		it('renders checked toggle switch', function() {
+			expect(shallowWrapper({checked: true})).toMatchSnapshot();
+		});
+
+		it('has a label only if one is given', () => {
+			const component = mountWrapper({label: 'Turn it on?'});
+			expect(component.find(`.${LABEL_CLASS}`)).not.toBe(null);
```
unjust: I was going to ask if it is more future proof to just look for the `label` tag rather than a class? `component.find('label')` but then I see we are not using a `label` element

src/forms/toggleSwitch.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+		it('calls `toggle()` and updates aria-checked when clicked', function() {
+			const component = mountWrapper();
+			const btnNode = component.find(`.${TOGGLE_SWITCH_CLASS}`);
+
+			expect(btnNode.props()['aria-checked']).toBe(false);
+			btnNode.simulate('click');
+			expect(btnNode.props()['aria-checked']).toBe(true);
+		});
+
+		it('calls `onCallback()` and `offCallback`', function() {
```
unjust: 💯

src/forms/ToggleSwitch.jsx
mperrotti: Didn't realize `<button>` was a label-able element until I looked up the spec: https://www.w3.org/TR/html5/forms.html#category-label

Good catch, will change!

src/forms/toggleSwitch.test.jsx
mperrotti: Will change this after doing your last suggestion

src/forms/ToggleSwitch.jsx
mperrotti: We don't _currently_ have any other props that might change, but I'll make it more specific just incase we ever do :)

PR #306, opened 2017-10-04
WC-105: Remove duplicate onClick in bindAll; Add preventDefault to onClick handler

aquillano:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-105

#### Description
I may have been the first to notice this because I have this component within a `form` element.

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -32,6 +32,10 @@ class Dropdown extends React.PureComponent {
 	onClick(e) {
 		e.preventDefault();
 		this.toggleContent();
+
+		if (this.props.onClick) {
+			this.props.onClick(e);
+		}
 	}
 
 	onKeyDown(e) {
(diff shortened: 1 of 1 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -31,6 +30,7 @@ class Dropdown extends React.PureComponent {
 	}
 
 	onClick(e) {
```
mperrotti: I'd recommend doing this like we do for `<TogglePill>` so we can still let engineers specify their own stuff to happen `onClick`:
````
	onClick(e) {
        e.preventDefault();
		this.props.toggleContent();

		if (this.props.onClick) {
			this.props.onClick(e);
		}
	}
````

mperrotti: Then just do this after we set up props at the beginning of the `render()` method:
`delete other.onClick;`

This way, a new `onClick` prop isn't passed into the element, but whatever the engineer said should happen `onClick` will happen in addition to `this.toggleContent`

PR #307, opened 2017-10-04
SDS-430: add validateBeforeTouched option to ReduxFormTextInput

burrata:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-430

#### Description
For field-level validations, redux-form automatically validates on first render unless you check for validation on whether the field has been touched.

It seems as though this is intended behavior.
https://github.com/erikras/redux-form/issues/120
>This is the purpose of the touched flags. In most UI cases, you should only show an error if a field has been touched (gone to and then blurred). There is also a visited flag, which is set to true onFocus. There's a working example of how the flags work here.
>Does that help?

Docs for `meta.touched`: https://redux-form.com/7.0.4/docs/api/field.md/#-code-meta-touched-boolean-code-

This ticket is to adjust the current `ReduxFormTextInput` implementation to take a flag that, when passed in, does not show the initial validation error until after the field has been touched. This is to prevent having to check for this in all form implementations we have.

Add venue form on first load, with these fields being required:
screen shot 2017-10-04 at 5 04 30 pm

### DIFF
--- a/src/forms/redux-form/TextInput.jsx
+++ b/src/forms/redux-form/TextInput.jsx
@@ -8,9 +8,10 @@ import TextInput from '../TextInput';
  * @return {Component} TextInput
  */
 const ReduxFormTextInput = props => {
-	const { meta, input, ...other } = props;
+	const { meta, input, validateAfterTouched, ...other } = props;
+	const error = (!validateAfterTouched || meta.touched) ? meta.error : null;
 
-	return <TextInput error={meta.error} {...input} {...other} />;
+	return <TextInput error={error} {...input} {...other} />;
 };
 
 ReduxFormTextInput.displayName = 'ReduxFormTextInput';
--- a/src/forms/redux-form/TextInput.jsx
+++ b/src/forms/redux-form/TextInput.jsx
@@ -8,8 +8,8 @@ import TextInput from '../TextInput';
  * @return {Component} TextInput
  */
 const ReduxFormTextInput = props => {
-	const { meta, input, validateAfterTouched, ...other } = props;
-	const error = (!validateAfterTouched || meta.touched) ? meta.error : null;
+	const { meta, input, validateBeforeTouched, ...other } = props;
+	const error = (validateBeforeTouched || meta.touched) ? meta.error : null;
 
 	return <TextInput error={error} {...input} {...other} />;
 };
--- a/src/forms/redux-form/textInput.test.jsx
+++ b/src/forms/redux-form/textInput.test.jsx
@@ -4,6 +4,7 @@ import ReduxFormTextInput from './TextInput';
 
 describe('redux-form TextInput', function() {
 
+	const MOCK_ERROR = 'Did you mean Batman and Robin?';
 	const formAttrs = {
 		input: {
 			label: 'Super Hero',
@@ -13,7 +14,7 @@ describe('redux-form TextInput', function() {
 			required: true
 		},
 		meta: {
-			error: 'Did you mean Batman and Robin?'
+			error: MOCK_ERROR
 		}
 	};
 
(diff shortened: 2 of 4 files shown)


src/forms/redux-form/textInput.test.jsx
```diff
@@ -13,15 +14,47 @@ describe('redux-form TextInput', function() {
...
+	});
 
+	describe('validateAfterTouched', () => {
+		it('always renders the error when `validateAfterTouched` is not true', () => {
+			const component = shallow(<ReduxFormTextInput {...formAttrs} />);
+			expect(component.prop('error')).toBe(MOCK_ERROR);
+		});
+		describe('when true', () => {
+			it('does not render the error when touched is false', () => {
+				const props = {
```
unjust: just pull these props out and set `props.validateAfterTouched = true` or `false` in each test

unjust:
I wonder if there are other ways to prevent this behavior rather than adding into `TextInput`. (I'm hesitant to embed an extra property here.)

Right now, in the event schedule form I think we don't see this issue because we avoid it - by not checking required fields in the [`validate`](https://redux-form.com/7.0.4/docs/api/reduxform.md/#-code-validate-values-object-props-object-gt-errors-object-code-optional-) function of the form.

Maybe we can look into leveraging the `initialRender` property of 
`shouldValidate` instead so when `validate` gets called we skip certain fields. https://redux-form.com/7.0.4/docs/api/reduxform.md/#-code-shouldvalidate-params-boolean-code-optional-

@mikespencer have you seen this issue with your forms on pro? how have you avoided it

mperrotti:
@mikespencer - We follow semver

mikespencer:
> @mikespencer have you seen this issue with your forms on pro? how have you avoided it

@unjust We've been using form components (eg [Input](https://github.com/meetup/pro-web/blob/master/src/app/shared/components/fields/Input.jsx)) that we created a while ago in pro-web before redux-form compatible form components were available here. We will need to update this on our end eventually, just haven't got around to it yet. These components use the `touched && error` check, though.

Seems a common pattern elsewhere too: [redux-form-material-ui abstract this logic into a reusable function](https://github.com/erikras/redux-form-material-ui/blob/master/src/mapError.js#L9)

burrata:
@unjust are you good with this change, per @mikespencer's thoughts above?

@mperrotti should I update the semver if I merge?

src/forms/redux-form/TextInput.jsx
```diff
@@ -8,9 +8,10 @@ import TextInput from '../TextInput';
  * @return {Component} TextInput
  */
 const ReduxFormTextInput = props => {
-	const { meta, input, ...other } = props;
+	const { meta, input, validateAfterTouched, ...other } = props;
+	const error = (!validateAfterTouched || meta.touched) ? meta.error : null;
```
unjust: Should this be `validateBeforeTouched` instead? Just thinking what the most common use case might be so we don't have to add an extra prop every time. I would think we would in most cases, want always to check `touched` before validating.

src/forms/redux-form/textInput.test.jsx
unjust: just pull these props out and set `props.meta.touched = true` or `false` in each test. I htink its more legible whats going on but you can disagree.

src/forms/redux-form/TextInput.jsx
unjust: also if we can add some documentation somewhere as to what this property is there for, that would be 💯

mikespencer: > Should this be validateBeforeTouched instead?

Actually I think that makes more sense too 👍

burrata: I'll make this update to both my prs!

unjust:
I approve. Ready to moige.

PR #310, opened 2017-10-05
Wc 106 redux form toggle

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-106

#### Description
The problem: Redux Form's implementation for checkboxes sets the value as `true` or `false`, unlike expected html behavior that separates `name`, `value`, and `checked`. 
https://github.com/erikras/redux-form/issues/2922

If we want our TogglePill's to be `checked`, we need to read the `input.value` prop from Redux Form (which is true or false), and pass that to `isActive`.

### DIFF
--- a/src/forms/redux-form/TogglePill.jsx
+++ b/src/forms/redux-form/TogglePill.jsx
@@ -21,8 +21,6 @@ const ReduxFormTogglePill = props => {
 		...other
 	} = props;
 
-	console.log(other, input);
-
 	return <TogglePill {...input} isActive={input.value == 'true' && !other.useRadio} {...other} />;
 };
 
--- a/src/forms/redux-form/TogglePill.jsx
+++ b/src/forms/redux-form/TogglePill.jsx
@@ -13,14 +13,17 @@ import TogglePill from '../TogglePill';
  * @return {Component} TogglePill
  */
 const ReduxFormTogglePill = props => {
+	// not adding meta error to TogglePill prop for now
+	// since TogglePills should function in a group with one error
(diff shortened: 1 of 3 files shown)


src/forms/redux-form/TogglePill.jsx
```diff
@@ -0,0 +1,30 @@
...
+const ReduxFormTogglePill = props => {
+	// not adding meta error to TogglePill prop for now
+	// since TogglePills should function in a group with one error
+	const {
+		meta, // eslint-disable-line no-unused-vars
+		input,
+		...other
+	} = props;
+
+	return <TogglePill {...input} isActive={input.value == 'true' && !other.useRadio} {...other} />;
```
unjust: reading `other.useRadio` since I do not know if there are similar idiosyncracies for Redux Form  radio values. I think if we come up against the same thing for the radios, we can remove this check.

src/forms/redux-form/TogglePill.jsx
```diff
@@ -0,0 +1,30 @@
...
+ * deconstructs props that redux-forms sets and sets them on TogglePill
+ * 
+ * NOTE: redux-form expects checkboxes value to be true/false and will set them as so
+ * https://github.com/erikras/redux-form/issues/2922, therefore using the value here to 
+ * set isActive
+ *  
+ * @return {Component} TogglePill
+ */
+const ReduxFormTogglePill = props => {
+	// not adding meta error to TogglePill prop for now
```
burrata: 👍

PR #314, opened 2017-10-06
SDS-435: Update selectinput to use `error` 

burrata:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-435

#### Description
We love standardization. Also, we need this for ReduxForm.

screen shot 2017-10-06 at 4 49 33 pm


mperrotti:
Changes make sense to me. Follow-up ticket filed and will be addressed when Adam returns and we can make an informed decision on whether we want an array or a string: https://meetup.atlassian.net/browse/SDS-436

PR #319, opened 2017-10-09
Popover `.inverted` bug fix

mperrotti:
#### Description
Popover menu options will always be colored `$C_textPrimary` by default

### DIFF
--- a/assets/scss/components/_popover.scss
+++ b/assets/scss/components/_popover.scss
@@ -51,6 +51,7 @@ $view-minWidth: 320px;
 	& > span,
 	& > div,
 	& > button {
+		color: $C_textPrimary;
 		display: block;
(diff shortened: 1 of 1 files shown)


assets/scss/components/_popover.scss
```diff
@@ -51,6 +51,7 @@ $view-minWidth: 320px;
 	& > span,
 	& > div,
 	& > button {
+		color: $C_textPrimary;
```
aquillano: I am using a button element, so this solved my issue, but what about anchor elements?

mperrotti: Anchors are also in this selector. See line 50

aquillano: Doh, Github hid the top part of the rule. Looks good!

PR #322, opened 2017-10-10
TextInput with icon

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-432

#### Description
Adds the option to have text inputs with icons. Chose to use a `background-image` instead of `<Icon>` in order to have the least changes to `<TextInput>` as possible.

screen shot 2017-10-10 at 4 14 16 pm

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -4,7 +4,6 @@ import cx from 'classnames';
 import { MEDIA_SIZES } from '../utils/designConstants';
 
 export const FIELD_WITH_ICON_CLASS = 'field--withIcon';
-export const DEFAULT_ICON_SIZE = 'xs';
 
 /**
  * @module TextInput
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -1,6 +1,10 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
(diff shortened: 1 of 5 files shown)


src/forms/textInput.test.jsx
```diff
@@ -112,5 +117,11 @@ describe('TextInput', function() {
 			const errorEl = component.find('.text--error').getDOMNode();
 			expect(errorEl.textContent).toEqual(ERROR_TEXT);
 		});
+
+		// it('should show an icon when iconShape prop is specified', () => {
```
emilyeng: can this be removed?

src/forms/TextInput.jsx
```diff
@@ -38,13 +45,26 @@ const TextInput = (props) => {
 		labelClassName
 	);
 
+	const iconSize = props.iconSize || 'xs';
```
emilyeng: you can pull `iconSize` out above with the other props (line ~29)

src/forms/textInput.test.jsx
mperrotti: Yea, my mistake

src/forms/TextInput.jsx
mperrotti: Keeping as-is in the interest of not having to define a new variable for icon size

PR #324, opened 2017-10-11
CalendarComponent datepickerOptions prop updates

mikespencer:
#### Related issues
Relates to https://meetup.atlassian.net/browse/MUP-15668

#### Description
Updating values via the `datepickerOptions` prop does not actually update the flatpickr config in `CalendarComponent`. This currently only gets set on `componentDidMount`. This causes a problem when trying to set a date range based off of the value of another date picker, for example. This PR fixes that and allows you to dynamically set the config of the flatpicker by passing an updated `datepickerOptions` prop.

`Prettier` seems to have made the diff larger than necessary - sorry about that.

### DIFF
--- a/src/forms/calendarComponent.story.jsx
+++ b/src/forms/calendarComponent.story.jsx
@@ -3,34 +3,38 @@ import CalendarComponent from './CalendarComponent';
 import { storiesOf } from '@kadira/storybook';
 import { InfoWrapper } from '../utils/storyComponents';
 import { decorateWithLocale } from '../utils/decorators';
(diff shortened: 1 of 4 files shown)


src/forms/calendarComponent.story.jsx
```diff
@@ -3,34 +3,38 @@ import CalendarComponent from './CalendarComponent';
...
-						name='event_time'
-						label='Start at'
-						value={new Date()}
-						datepickerOptions={{ allowInput: true }}
-					/>
-				</div>
-			</InfoWrapper>
-		)
-	)
+	.addDecorator(withKnobs)
```
unjust: 💯

PR #326, opened 2017-10-12
Bugfix - character count layout

mperrotti:
#### Description
Moved error element outside of `.inputContainer` so when it get's absolutely positioned, it's always inside the input box

BAD (before):
screen shot 2017-10-12 at 11 30 44 am

GOOD (after):
screen shot 2017-10-12 at 11 30 27 am

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -113,29 +113,30 @@ class Textarea extends React.Component {
 		};
 
 		return (
-			<div className="inputContainer">
-				{label &&
-					<label className={classNames.label} htmlFor={id}>
-						{label}
-					</label>
-				}
+			<div>
+				<div className="inputContainer">
+					{label &&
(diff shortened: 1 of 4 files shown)


src/forms/Textarea.jsx
```diff
@@ -113,29 +113,30 @@ class Textarea extends React.Component {
...
-					style={{ ...style, ...heightConstraints }}
-					id={id}
-					value={this.state.value}
-					{...other}
-				/>
-
-				{ maxLength && <p className='text--tiny text--secondary align--right charCount'>{parseInt(maxLength - this.state.value.length)}</p> }
-
+			<div>
+				<div className="inputContainer">
```
unjust: is `inputContainer` class providing anything? should we remove? should we move it one level up ?

mperrotti: It's there on purpose - it does `position: relative;`, keeping the character count at the bottom right of the input

PR #329, opened 2017-10-13
Accordion panel changes

unjust:
The PR has moved to: https://github.com/meetup/meetup-web-components/pull/330

#### Related issues

There is no bug for this yet.

#### Description
When using the `AccordionPanelGroup` in `mup-web` problems with re-rendering and props came up where the accordion panel would open and immediately rerender and close.

The problem seemed to be in AccordionPanelGroup's `render/clonePanels` method, which relied on `props` from each panel to dictate whether or not the panel should be rendered open.
This prop was often still `false` even after being clicked open, it never gets set as `true` by any parent.

I'm thinking that the reason it looks like it works in storybook and not `mup-web`, is that storybook doesnt have as may parents/containers that can trigger re-renders like our app.

The solution I implemented was to have the `AccrodionPanelGroup` keep track of `openPanels` in its state. `clickId` prop is given to each panel by the panelGroup in order to identify whether or not the panel is in its `openPanels`.

The code needs some cleanup, but I would like feedback on whether or not this is correct.
It works in storybook and in `mup-web` event schedule.

- [ ] update tests

### DIFF
--- a/src/interactive/AccordionPanel.jsx
+++ b/src/interactive/AccordionPanel.jsx
@@ -17,11 +17,7 @@ export const ACTIVEPANEL_CLASS = 'accordionPanel--active';
 class AccordionPanel extends React.Component {
 	constructor(props){
 		super(props);
-
-		this.state = {
-			isOpen: this.props.isOpen
-		};
-
+		this.state = {};
 		this._handleToggle = this._handleToggle.bind(this);
 	}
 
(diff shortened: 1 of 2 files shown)


src/interactive/AccordionPanel.jsx
```diff
@@ -66,15 +66,34 @@ class AccordionPanel extends React.Component {
...
 		if (nextProps.isOpen !== this.state.isOpen) {
+			const isOpen = nextProps.isOpen;
 			this.setState({
-				isOpen: nextProps.isOpen,
-				height: this.getHeight(nextProps.isOpen)
+				isOpen: isOpen,
+				height: this.getHeight(isOpen)
 			});
 		}
 	}
```
unjust: I read that you shouldn't update state in `componentWillUpdate`, so using `componentWillReceiveProps`

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -28,30 +30,62 @@ class AccordionPanelGroup extends React.Component {
...
-	 * @param {number} i - index of the `AccordionPanel`
-	 * @param {boolean} isOpen - whether the `AccordionPanel` is open or not
+	 * @param {number} i - index of the `AccordionPanel` used as the key
 	 * @returns {Array} `AccordionPanel` components with props from `AccordionPanelGroup`
 	 */
-	clonePanel(accordionPanel, key, isOpen) {
-		const panelProps = {
-			key,
+	clonePanel(accordionPanel, i, arr, initialize) {
+		const isOpen = (initialize) ? accordionPanel.props.isOpen : this.isPanelInState(accordionPanel);
```
unjust: this `initialize` arg might be awkward and I'm sure this can use some cleanup. What I'm trying to do is have the AccordionPanelGroup assign `clickId` props when cloning the panels at first, and only save them into state on initialization.

unjust:
@mmcgahan  thanks! making some changes. I think AccordionPanel might be able to use `prop`.isOpen` rather than `state` in order to calculate height.

PR #330, opened 2017-10-16
Wc 111 accordion panel group state changes

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-111

#### Description
When using the `AccordionPanelGroup` in `mup-web` problems with re-rendering and props came up where the accordion panel would open and immediately rerender and close.

The problem seemed to be in AccordionPanelGroup's `render/clonePanels` method, which relied on `props` from each panel to dictate whether or not the panel should be rendered open.
This prop was often still `false` even after being clicked open, it never gets set as `true` by any parent.

I'm thinking that the reason it looks like it works in storybook and not `mup-web`, is that storybook doesnt have as may parents/containers that can trigger re-renders like our app.

The solution I implemented was to have the `AccrodionPanelGroup` keep track of `openPanels` in its state. `clickId` prop is given to each panel by the panelGroup in order to identify whether or not the panel is in its `openPanels`.

The code needs some cleanup, but I would like feedback on whether or not this is correct.
It works in storybook and in `mup-web` event schedule.

- [x] update tests


src/interactive/AccordionPanel.jsx
```diff
@@ -34,42 +32,41 @@ class AccordionPanel extends React.Component {
...
+	_handleToggle(e){
+		e.preventDefault();
+		e.stopPropagation();
 
+		const isToggledOpen = !this.props.isOpen;
+		this.props.setClickedPanel && this.props.setClickedPanel(this.props.clickId, isToggledOpen);
 		this.setState({
-			height: this.getHeight(!this.state.isOpen),
-			isOpen: !this.state.isOpen
+			height: this.getHeight(isToggledOpen)
```
mmcgahan: I think you can/should skip the `setState` here - it will be handled by the parent update of `isOpen` in `componentWillReceiveProps`.

In fact, do you need to keep track of `height` in `this.state` at all? Structurally, it would be nice if `height` _always_ corresponded to `this.props.isOpen` - it would never get out of sync, and the component wouldn't have to have _any_ internal `state`, and wouldn't have to implement `componentDidMount` or `componentWillReceiveProps`

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -90,7 +118,9 @@ class AccordionPanelGroup extends React.Component {
 			>
 				{
 					this.renderAccordionPanels().map((panel, i) => (
```
mmcgahan: there's something funky happening here - `this.accordionPanels` is set to a _cloned_ list of panels in the constructor, and then `this.renderAccordionPanels` does a clone of the clone inside `render()` - you should only need to clone once, and then I think you can just use `this.accordionPanels` on this line instead of `this.renderAccordionPanels()`.

Related: the`clonePanel` method should _not_ be setting a `key` - it's not needed, since each individual panel is wrapped inside an `<li>` with a `key` prop here in the `render()` method.

I think the net render result will be the same, but you can simplify the implementation by stripping out a couple things.

src/interactive/AccordionPanel.jsx
```diff
@@ -34,42 +32,41 @@ class AccordionPanel extends React.Component {
...
-	 *
-	 * Updates state to toggle `AccordionPanel` open and closed
 	 */
-	_handleToggle(){
-		if (this.props.setClickedPanel) {
-			this.props.setClickedPanel(this);
-		}
+	_handleToggle(e){
+		e.preventDefault();
+		e.stopPropagation();
```
mperrotti: I think we just want to `preventDefault` here - unless there's a reason we're also stopping propagation?

src/interactive/AccordionPanelGroup.jsx
unjust: got it. yea I thought that was funky too but didnt question it enough.

src/interactive/AccordionPanel.jsx
unjust: I need to try it in mup-web and make sure the form doesnt attempt submission.

src/interactive/AccordionPanel.jsx
unjust: so I was thinking the same, however - the `contentEl` that is required to calculate height is a `ref` that it set after a height needs to be calculated.

setting the height in `state` in `componentDidMount` allows for the `ref` to be there and also since its in `state` triggers a necessary re-render.

I tried just using `props`, but if a panel is open by default, the ref isnt set in time with render, and the height will get calculated wrong, and since no state is getting updated in `componentDidMount` there is no re-render with the correct height.

thoughts?

src/interactive/AccordionPanelGroup.jsx
unjust: well I realize the clone is necessary to set the appropriate `isOpen` prop from state during render, unless there is another way. I refactored the code into `initPanel` and `cloneAccordionPanels` to be a little less awkward and hopefully more clear what is going on .

src/interactive/AccordionPanel.jsx
mmcgahan: I'm probably not seeing all the moving pieces, but I think you can still avoid state, but use `componentDidMount` to call [`this.forceUpdate()`](https://reactjs.org/docs/react-component.html#forceupdate) - I think that will allow the `ref` to be set on initial render, and allow `height` to be calculated correctly the second time around.

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -11,55 +11,78 @@ class AccordionPanelGroup extends React.Component {
...
 		const panelProps = {
-			key,
 			indicatorAlign: this.props.indicatorAlign,
 			indicatorIcon: this.props.indicatorIcon,
 			indicatorIconActive: this.props.indicatorIconActive,
 			indicatorSwitch: this.props.indicatorSwitch,
-			className: accordionPanel.props.className,
-			setClickedPanel: this.setClickedPanel,
-			isOpen: accordionPanel.props.isOpen
+			className: panel.props.className,
```
mmcgahan: this line shouldn't be necessary - cloning the panel will retain any original props that are not overridden

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -75,6 +98,9 @@ class AccordionPanelGroup extends React.Component {
 			...other
 		} = this.props;
 
+		// gives us the correct isOpen prop from state
+		this.cloneAccordionPanels();
```
mmcgahan: this makes sense to me now.

src/interactive/AccordionPanel.jsx
```diff
@@ -30,49 +25,33 @@ class AccordionPanel extends React.Component {
 	 * @returns {Number} panel height
 	 */
 	getHeight(isOpen) {
+		if (!this.contentEl) {
+			return 'auto';
```
unjust: @mperrotti is this a kosher default? or is `0px` better.

mperrotti: Since we animate height - `0px` is better. You can't animate from `auto`

PR #340, opened 2017-10-23
add disabled attribute to Select input options

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-120

screen shot 2017-10-23 at 11 07 35 am
screen shot 2017-10-23 at 11 07 30 am

### DIFF
--- a/src/forms/selectInput.story.jsx
+++ b/src/forms/selectInput.story.jsx
@@ -57,7 +57,7 @@ storiesOf('SelectInput', module)
 			required
 			options={[
 				{ label: 'Geoffrey', value: 'geoffrey', disabled: true },
-				{ label: 'Doctor Horse, MD Junior', value: 'drhorse', disabled: false },
+				{ label: 'Doctor Horse, MD Junior', value: 'drhorse' },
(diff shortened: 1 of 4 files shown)


unjust:
@mperrotti don't want to block this but should I add some style somewhere for a more disabled look?

src/forms/selectInput.story.jsx
```diff
@@ -49,6 +49,19 @@ storiesOf('SelectInput', module)
...
 	)
+	.add('with disabled options', () =>
+		(<SelectInput
+			label='Select a name for your horse'
+			id='horsename'
+			name='horsename'
+			required
+			options={[
+				{ label: 'Geoffrey', value: 'geoffrey', disabled: true },
+				{ label: 'Doctor Horse, MD Junior', value: 'drhorse', disabled: false },
```
mperrotti: Instead of saying `disabled: false`, I think we can just leave it off

unjust: ok. I just wanted to make sure passing attribute as false didnt do anything weird. I'll take it out though.

PR #342, opened 2017-10-24
Clone accordionPanels to get fresh props

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-121

#### Description
Props on accordion panels in accordion panel group become stale, since we are cloning the `this.accordionPanels` which get set on init, rather than whats coming in updated from props.

As a result the panels and panel content weren't getting fresh props and not updating.

relates a bit to https://github.com/meetup/meetup-web-components/pull/330/


aquillano:
LGTM

PR #343, opened 2017-10-24
Upgrade to React 16

mmcgahan:
There's a lot of noise from Prettier, but the main things were converting from `react-addons-test-utils` (which is deprecated and doesn't work with React 16) to `react-dom/test-utils` and updating the Enzyme adapter to the React 16 version.

I also added some missing peer dependencies and rebuild the yarn.lock file to update transitive deps

#### Important

One thing I have _not_ done is update `@kadira/storybook...` deps to the latest `@storybook/...` packages. The older packages do not officially support React 16 but I don't think that will be a problem in the short term. I'd like @mperrotti and @akdetrick to make a plan for implementing that upgrade sometime in the not-too-distant future so that we don't end up breaking our storybook as we write more React-16-savvy components


willh-meetup:
🙀

PR #345, opened 2017-10-25
Wc 122 accordion panel click callback

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-122

#### Description
adding a callback fro when a panel is clicked. We need a way to know if a panel is open/active for Event Schedule.


unjust:
@mperrotti - what were you thinking for stroybook? theres no change in visual output or behavior.
let me know what you're thinking. I did write a test.

mperrotti:
@unjust - just use `action` from Storybook. I just want it to be more obvious that the callback functionality is there without having to look at `AccordionPanel.jsx`. You can even just add that prop to all the panels in existing stories if you prefer not to create a whole new story

PR #349, opened 2017-10-30
Reset form styles for date and time inputs

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-2475

#### Description
Date, time, and datetime inputs were getting default `<select>` styling in mobile browsers. This PR fixes that, but I'm stumped on how we can safely show the down chevron when the input behaves as a `<select>`.

screen shot 2017-10-30 at 5 44 15 pm


mperrotti:
Spoke with Natalia, she prefers shipping without an arrow. Ready for review!

PR #350, opened 2017-10-31
Toggle Pills in radioButtonGroups should be responsive and felx to full width

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-124

#### Description

Toggle pills in radio button group do not fill width when you tell the group to switch to coulmn at smaller widths. 

Right now:

screen shot 2017-10-30 at 9 09 21 pm

we want it to be like this:
screen shot 2017-10-30 at 8 54 05 pm

- [x] need story
- [x] add test?

### DIFF
--- a/assets/scss/components/_togglePill.scss
+++ b/assets/scss/components/_togglePill.scss
@@ -7,6 +7,10 @@
 	@include shadowOnHover();
 	margin: $space-quarter 0;
 	-webkit-tap-highlight-color: rgba(0,0,0,0);
+
+	.radioButtonGroup & {
+		@include display(flex);
+	}
 }
 
 .toggleButton-label--large {
@@ -14,7 +18,7 @@
 	padding: $space*1.5 $space;
 }
 
-.toggleButton-input{
+.toggleButton-input {
 	&:checked + .toggleButton-label {
 		@include buttonColor(
 			$bgColor: $C_blue,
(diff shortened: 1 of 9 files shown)


assets/scss/components/_togglePill.scss
```diff
@@ -7,14 +7,18 @@
 	@include shadowOnHover();
 	margin: $space-quarter 0;
 	-webkit-tap-highlight-color: rgba(0,0,0,0);
+
+	.radioButtonGroup & {
```
mperrotti: I'd prefer to keep things as neatly separated as possible. Instead of this, I'd recommend adding a class of `span--100` or `display--flex` to each TogglePill in the group.

src/forms/RadioButtonGroup.jsx
```diff
@@ -37,9 +38,23 @@ export default class RadioButtonGroup extends PureComponent {
...
-		const { direction, children, name, className } = this.props;
+		const {
+			direction,
+			switchDirection,
+			children,
+			name,
+			className
+		} = this.props;
+
+		const classNames = cx(
```
mperrotti: If we remove the new CSS in `_togglePill.scss`, we can get rid of the classname stuff

assets/scss/components/_togglePill.scss
unjust: :thumbsup: :thumbsup:  :thumbsup:

PR #351, opened 2017-11-01
Better support for <TimeInput>

mperrotti:
Re: Prelim tag -
- [x] Get general feedback from @unjust about this approach
- [x] Limit what numbers can be entered via keyboard (e.g. "25:00" would not work)
- [x] Handle disabled state
- [x] Handle error state
- [ ] Make sure this will work with `redux-forms`
- [x] Tests
- [x] Stories

#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-2490

#### Description
Adding support for [a few important browsers](http://caniuse.com/#feat=input-datetime) don't support `input[type=time]`

Will add when bug is completed

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -65,8 +65,8 @@ class TimeInput extends React.Component {
 				...partialState
 			};
 
-			const v = `${formatDigits(stateValues.hours) % (this.props.is24Hr ? 24 : 12)}:${formatDigits(stateValues.minutes)}`;
-			this.props.onChange(v);
+			const value = `${formatHours(stateValues.hours) % (this.props.is24Hr ? 24 : 12)}:${formatDigits(stateValues.minutes)}`;
+			this.props.onChange(value);
 		}
 	}
 
@@ -79,7 +79,6 @@ class TimeInput extends React.Component {
 	onTimeInputChange(e) {
 		const { value } = e.target;
 		this.setState(() => ({ value: value }));
-
 		this.props.onChange && this.props.onChange(value);
 	}
 
@@ -91,7 +90,6 @@ class TimeInput extends React.Component {
 	*/
 	onNumberChange(e) {
 		const { value, id } = e.target;
-
 		this.setState(() => ({ [id]: value }));
 	}
 	/**
@@ -102,7 +100,6 @@ class TimeInput extends React.Component {
 	*/
 	onMeridianChange(e) {
 		const { value } = e.target;
-
 		this.setState(() => ({ meridian: value }));
 	}
 
@@ -114,17 +111,15 @@ class TimeInput extends React.Component {
 	*/
 	onBlur(e) {
 		const { value, min, max, id } = e.target;
+		const constrainedVal = this.constrainValue(value,min, max);
+		this.onChange({ [id]: formatDigits(constrainedVal) });
+	}
 
+	constrainValue(value, min, max) {
 		if (max || min) {
-			const constrainedVal = Math.max(Math.min(value, max), min);
-			if (constrainedVal == value) {
-				this.setState(() => ({ [id]: formatDigits(value) }));
-				this.onChange();
-			} else {
-				this.setState(() => ({ [id]: formatDigits(constrainedVal) })); // try doing onChange stuff in this setstate callback
-				this.onChange({ [id]: formatDigits(constrainedVal) });
-			}
+			value = Math.max(Math.min(value, max), min);
 		}
+		return value;
 	}
 
 	componentDidMount() {
@@ -265,16 +260,6 @@ class TimeInput extends React.Component {
 									}
 								</Flex>
 							</div>
-							{
-								<input
-									type="text"
-									className={HIDDEN_INPUT_CLASS}
-									id={id}
-									name={name}
-									value={`${formatHours(this.state.hours, this.state.meridian)}:${this.state.minutes}`}
-									onChange={this.onChange}
-									/>
-								}
 						</div>
 				}
 
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -6,7 +6,10 @@ import Flex from '../layout/Flex';
 import FlexItem from '../layout/FlexItem';
 import SelectInput from './SelectInput';
 
-let supportsTime = true;
+export const FAUX_INPUT_CLASS = 'fauxInput';
+export const MERIDIAN_CLASS = 'timeInput-meridian';
+
+// let supportsTime = true;
 const formatDigits = (number) => `0${number}`.slice(-2);
 
 /**
@@ -18,11 +21,13 @@ class TimeInput extends React.Component {
 		super(props);
 
 		this.state = {
-			hours: props.hours || '01',
+			supportsTime: true,
+			hours: props.hours || '12',
 			minutes: props.minutes || '00',
-			meridian: props.meridian || props.isMilitaryTime ? false : 'PM'
+			meridian: props.meridian || props.is24Hr ? false : 'AM'
 		};
 
+		this.onBlur = this.onBlur.bind(this);
 		this.onHoursChange = this.onHoursChange.bind(this);
 		this.onMinutesChange = this.onMinutesChange.bind(this);
 		this.onMeridianChange = this.onMeridianChange.bind(this);
@@ -42,6 +47,8 @@ class TimeInput extends React.Component {
 	onHoursChange(e) {
 		const { value } = e.target;
 
+		console.log(formatDigits(value));
+
 		this.setState(() => ({ hours: formatDigits(value) }));
 	}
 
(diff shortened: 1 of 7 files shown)


src/forms/TimeInput.jsx
```diff
@@ -23,6 +39,29 @@ class TimeInput extends React.Component {
...
+
+	onMeridianChange(e) {
+		const { value } = e.target;
+
+		this.setState(() => ({ meridian: value }));
+	}
+
+	componentDidMount() {
+		supportsTime = this.inputEl.type === 'time'; /* comment OUT this line to test in browsers that support time */
+		// supportsTime = false; /* UNcomment this line to test in browsers that support */
```
unjust: you could also have a prop `forceTimeInput` to get the same effect

src/forms/TimeInput.jsx
```diff
@@ -56,24 +98,85 @@ class TimeInput extends React.Component {
...
+											onChange={this.onMeridianChange}
+											options={[
+												{ label: 'PM', value: 'PM' },
+												{ label: 'AM', value: 'AM' }
+											]}
+										/>
+									</FlexItem>
+								}
+							</Flex>
+							<input
```
unjust: also I think this input would have the actual `name` of the time input - since we would want to get the value here.

src/forms/TimeInput.jsx
```diff
@@ -10,6 +17,15 @@ class TimeInput extends React.Component {
 	constructor(props) {
 		super(props);
 
+		this.state = {
+			hours: props.hours || '01',
```
unjust: nitpick maybe default to `12:00 AM` `00:00`

src/forms/TimeInput.jsx
```diff
@@ -48,6 +87,9 @@ class TimeInput extends React.Component {
 
 		const errorId = `${id}-error`;
 
+		const formatHours = (hours, meridian) => meridian ? (this.state.hours % 12 + 12 * (meridian === 'PM')) : (this.state.hours % 24);
```
unjust: replace `this.state.hours` with the `hours` arg

src/forms/TimeInput.jsx
```diff
@@ -10,6 +17,15 @@ class TimeInput extends React.Component {
 	constructor(props) {
 		super(props);
 
+		this.state = {
+			hours: props.hours || '01',
+			minutes: props.minutes || '00',
+			meridian: props.meridian || props.isMilitaryTime ? false : 'PM'
```
unjust: so this `meridian`/`isMilitaryTime` prop in a way would be dictated by locale right? - since some use 24hr time and some 12hr.
it'd be nice if we could find a way to pass in locale and use that to dictate all of this.

`props.is24Hr` might be better than military?

mperrotti: I thought about doing that, but since all the locale stuff is in mup-web, I was thinking we could check locale in there and pass a true or false from that information. If that doesn't make sense, I'm happy to put the logic into the component

Agreed on the prop name change though 👍

src/forms/TimeInput.jsx
```diff
@@ -2,6 +2,27 @@ import PropTypes from 'prop-types';
...
+
+export const HOURS_INPUT_CLASS = 'timeInput-hours';
+export const MINUTES_INPUT_CLASS = 'timeInput-minutes';
+export const MERIDIAN_INPUT_CLASS = 'timeInput-meridian';
+export const HIDDEN_INPUT_CLASS = 'timeInput-hidden';
+
+const formatDigits = (number) => `0${number}`.slice(-2);
+const getTimeParts = (time, part) => {
+	const splitTime = time.split(':');
+	const timeParts = new Object();
```
mmcgahan: just assign the whole object in one go.

```js
const [ hours, minutes ] = time.split(':');
const parts = {
  hours,
  minutes,
  meridian: hours < 12 ? 'AM': 'PM',
};
return parts[part];
```

also, avoid `new Object()` - use the `{}` object literal.

src/forms/TimeInput.jsx
```diff
@@ -23,6 +54,66 @@ class TimeInput extends React.Component {
...
+	* 	in turn changes the state and updates the value of the inputs
+	*/
+	onNumberChange(e) {
+		const { value, id } = e.target;
+
+		id === 'hours' &&
+			this.setState(() => ({ hours: value }));
+
+		id === 'minutes' &&
+			this.setState(() => ({ minutes: value }));
```
mmcgahan: would this work instead of the two ternaries?

```js
this.setState(() => ({ [id]: value });
```

src/forms/TimeInput.jsx
```diff
@@ -23,6 +54,66 @@ class TimeInput extends React.Component {
...
+		if (value > max){
+			e.target.value = max;
+
+			id === 'hours' &&
+				this.setState(() => ({ hours: formatDigits(max) }));
+
+			id === 'minutes' &&
+				this.setState(() => ({ minutes: formatDigits(max) }));
+
+		} else if (value < parseInt(min)){
```
mmcgahan: `parseInt` should usually have a second arg of `10` (the 'radix') to avoid being parsed with a non-decimal base

mmcgahan: why use `parseInt` for `min` but not for `max`?

src/forms/TimeInput.jsx
```diff
@@ -23,6 +54,66 @@ class TimeInput extends React.Component {
...
+			e.target.value = min;
+
+			id === 'hours' &&
+				this.setState(() => ({ hours: formatDigits(min) }));
+
+			id === 'minutes' &&
+				this.setState(() => ({ minutes: formatDigits(min) }));
+		}
```
mmcgahan: I think the repetitive code could be eliminated:

```js
if (max || min) {
  // this is guaranteed to be an input that is tracked in state[id]
  const constrainedVal = Math.max(Math.min(value, max), min);
  if (constrainedVal !== value) {
    this.setState(() => ({ [id]: constrainedVal }));
  }
}
```

src/forms/TimeInput.jsx
```diff
@@ -41,42 +134,139 @@ class TimeInput extends React.Component {
...
 			{ required }
 		);
 
 		const errorId = `${id}-error`;
 
+		const formatHours = (hours, meridian) => {
+			const newHours = meridian ? (hours % 12 + 12 * (meridian === 'PM')) : hours % 24 ;
+			return(
+				formatDigits(newHours)
+			);
```
mmcgahan: extra parens here

src/forms/TimeInput.jsx
```diff
@@ -41,42 +134,139 @@ class TimeInput extends React.Component {
...
+		);
+
 		const labelClassNames = cx(
 			'label--field',
 			{ required }
 		);
 
 		const errorId = `${id}-error`;
 
+		const formatHours = (hours, meridian) => {
```
mmcgahan: this function should be defined outside of the component, not every time `render` runs

src/forms/TimeInput.jsx
mperrotti: Just forgot to bring it up there. Thanks!

src/forms/TimeInput.jsx
```diff
@@ -23,6 +58,50 @@ class TimeInput extends React.Component {
...
+	* @description called when either the hour or minute input loses focus,
+	* 	and ensures the value entered is not out of the min/max range
+	*/
+	onBlur(e) {
+		const { value, min, max, id } = e.target;
+
+		if (max || min) {
+			const constrainedVal = Math.max(Math.min(value, max), min);
+			if (constrainedVal !== value) {
+			this.setState(() => ({ [id]: constrainedVal }));
```
unjust: indent

src/forms/TimeInput.jsx
```diff
@@ -54,29 +152,102 @@ class TimeInput extends React.Component {
...
+									}
+								</Flex>
+							</div>
+							<input
+								type="hidden"
+								className={HIDDEN_INPUT_CLASS}
+								id={id}
+								name={name}
+								value={`${formatHours(this.state.hours, this.state.meridian)}:${this.state.minutes}`}
+								onChange={this.onChange}
```
unjust: so I don't think `onChange` fires with a hidden input, there needs to be another way to tell redux-form that the values have changed and fire `this.props.onChnage`

unjust:
Messing with this some more - we need a way to tell redux-form there was a change in value somehow, since `onChange` on the hidden field wont work.
 
https://gist.github.com/unjust/0bd1eae6e1e66b5529ee01c7eea63f97

I wanted to trigger `onChange` as a `setState` callback in the `numberOnChange` etc, but couldnt get it work.
Instead I am dispatching the change to redux-form in `onBlur`.

The component also needs to make sure its state for values is in sync with the `value` prop from the parent/redux-form. so when it receives new value props, it needs to `setState` to keep state up to date with any value the parent can be trying to set.

I also am playing with returning false in `componentShouldUpdate` if there were no state changes.


https://gist.github.com/unjust/0bd1eae6e1e66b5529ee01c7eea63f97

mperrotti:
Closing this, instead we're reviewing at: https://github.com/meetup/meetup-web-components/pull/362

PR #362, opened 2017-11-06
Better support for <TimeInput>

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-129

#### Description
Adding support for [a few important browsers](http://caniuse.com/#feat=input-datetime) that don't support `input[type=time]`

Native input[type=time]:
screen shot 2017-11-06 at 3 09 29 pm

Fallback with text inputs and `<select>`:
screen shot 2017-11-06 at 3 08 45 pm


src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
 	* @description called when the input changes, in turn calls the onChange
```
unjust: need to update doc

src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
-	onChange(e) {
-		this.props.onChange && this.props.onChange(e);
+	onChange(partialState) {
+		if (this.props.onChange) {
+			const stateValues = {
+				...this.state,
+				...partialState
+			};
+
+			const v = `${formatDigits(stateValues.hours) % (this.props.is24Hr ? 24 : 12)}:${formatDigits(stateValues.minutes)}`;
```
unjust: formatDigits or formatHours? this was something I rushed

src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
+
+	componentDidMount() {
+		this.setState(() => ({ supportsTime: !this.props.forceTextInput && this.inputEl.type === 'time' }));
+	}
+
+	componentWillReceiveProps(nextProps) {
+		if (this.props.value !== nextProps.value) {
+			const props = nextProps;
+			this.setState({
+				hours: props.value && (formatDigits(getTimeParts(props.value, 'hours') % (props.is24Hr ? 24 : 12)) || '12').toString(),
```
unjust: again - formatHours? 
how bout a format method like `parseTimeValue` that takes a value and gives back the state object.
we can use this in the constructor too

``` 
parseTimeValue(value, is24Hr) {
    return {
        hours: ...
        minutes: ...
   }
} ```

File: src/forms/TimeInput.jsx (hunk 2 above)
Comment (mperrotti) [reply src/forms/TimeInput.jsx]: Good catch, it should be formatHours. I'll get that back in there now...

File: src/forms/TimeInput.jsx [hunk 4]
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
+			}
+		}
+	}
+
+	componentDidMount() {
+		this.setState(() => ({ supportsTime: !this.props.forceTextInput && this.inputEl.type === 'time' }));
+	}
+
+	componentWillReceiveProps(nextProps) {
+		if (this.props.value !== nextProps.value) {
```
unjust: put a comment here how we need to update state to stay in sync with new props from the parent

src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
+			const props = nextProps;
+			this.setState({
+				hours: props.value && (formatDigits(getTimeParts(props.value, 'hours') % (props.is24Hr ? 24 : 12)) || '12').toString(),
+				minutes: props.value && (formatDigits(getTimeParts(props.value, 'minutes')) || '00').toString(),
+			});
+		}
+	}
+
+	shouldComponentUpdate(nextProps, nextState) {
```
unjust: I'm forgetting why I put this here. Maybe it was from another version I tried. this may be able to be removed.

src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
+	* @param e Event Object
+	* @description called when either the hour or minute input loses focus,
+	* 	and ensures the value entered is not out of the min/max range
+	*/
+	onBlur(e) {
+		const { value, min, max, id } = e.target;
+
+		if (max || min) {
+			const constrainedVal = Math.max(Math.min(value, max), min);
+			if (constrainedVal == value) {
```
unjust: is this the only place to constrain the value? you can factor it out into just its own mothod that returns a value
```constrainValue(value, min, max) ...```

if anything - I think this might be more leg:
```
let newValue = value;
if (max || min) { // constrain values
    newValue = Math.max(Math.min(value, max), min);
}
const newState = { [id]: formatDigits(newValuel) };
this.setState(() => (newState)); // try doing onChange stuff in this setstate callback
this.onChange(newState);
```

File: src/forms/TimeInput.jsx (hunk 5 above)
Comment (mikespencer) [reply src/forms/TimeInput.jsx]: seems like a premature optimisation to me... if I were you I think I'd chop it out. Also, we'd probably need to do a deep equals comparison here because i'm guessing `this.state` is not the same object reference as `nextState`, right?

File: src/forms/TimeInput.jsx (hunk 6 above)
Comment (mmcgahan) [reply src/forms/TimeInput.jsx]: I really like the idea to break `constrainValue` out into a function.

In terms of signature, I'd recommend

```js
constrainValue({ min=0, max=Infinity }, value)
```

that will allow you to have an _optional_ `min` and `max` param, and then you can pass in `value` as a second argument to make it easier to curry the function like

```js
const constrainTo100 = constrainValue.bind(null, { max:100 });
const contrainedValue = constrainTo100(value);
```

(not that you'd ever need to do that kind of currying, but it helps explain why having `value` as the _last_ argument is better than having it as the _first_ arg.

src/forms/TimeInput.jsx
```diff
@@ -19,8 +50,91 @@ class TimeInput extends React.Component {
...
+
+	/**
+	* @function onTimeInputChange
+	* @param e Event Object
+	* @description called when the browser-native time input changes, in turn calls the onChange
+	* 	handler prop, if there is one provided (eg supplied by redux-form or DateTimePicker)
+	*/
+	onTimeInputChange(e) {
+		const { value } = e.target;
+		this.setState(() => ({ value: value }));
```
mikespencer: Just curious - any particular reason for doing this instead of just `this.setState({ value: value });`?

mperrotti: I got feedback in a PR a while back that said to do it this way, so that's how I've been doing it ever since. Either way works though :)

mmcgahan: the callback form of `setState` is now [the default syntax for React](https://reactjs.org/docs/react-component.html#setstate), and the React maintainers have suggested that it might end up being the only supported syntax in future versions of React because it enables some performance optimizations that apparently are more difficult with the object syntax. It's definitely a 'loose' recommendation to use the callback rather than the object arg form, but it's guaranteed future-proof in a way that the object form is not, and has a tiny benefit of making it clear that the state update will be done asynchronously - you can get yourself into trouble thinking about `setState` as a synchronous call, and the object form doesn't give you any indication that it's otherwise.

I would recommend `{ value }` instead of `{ value: value }`, however :)

src/forms/TimeInput.jsx
```diff
@@ -2,6 +2,40 @@ import PropTypes from 'prop-types';
...
+import FlexItem from '../layout/FlexItem';
+import SelectInput from './SelectInput';
+
+export const HOURS_INPUT_CLASS = 'timeInput-hours';
+export const MINUTES_INPUT_CLASS = 'timeInput-minutes';
+export const MERIDIAN_INPUT_CLASS = 'timeInput-meridian';
+
+export const HOURS = 'hours';
+export const MINUTES = 'minutes';
+export const MERIDIAN = 'meridian';
```
mmcgahan: I think I would avoid abstracting the keys into consts - they're intuitively named and don't need to be used outside of this component code, but fine to leave in if you found it useful to use them. Sort of makes object shorthand and destructuring more awkward.

src/forms/TimeInput.jsx
```diff
@@ -2,6 +2,40 @@ import PropTypes from 'prop-types';
...
+ * @returns {String|Object} the only parsed value or all the parsed parts if no part defined
+ */
+export const getTimeParts = (time, part) => {
+	const [ hours, minutes ] = time.split(':');
+	const parts = {
+		[HOURS]: hours,
+		[MINUTES]: minutes,
+		[MERIDIAN]: hours < 12 ? 'AM': 'PM',
+	};
+	return (part) ? parts[part] : parts;
```
mmcgahan: this makes the return value more difficult to reason about - 'string OR object' vs. always returning the whole object and letting the caller key into it.

I recommend simplifying the return value to just be the object since it doesn't really cost the caller anything to key into the return value itself

```js
const { hours } = getTimeParts(time);
// vs
const hours = getTimeParts(time, HOURS);
```

src/forms/TimeInput.jsx
```diff
@@ -2,6 +2,40 @@ import PropTypes from 'prop-types';
...
+};
+const formatHours = (hours, meridian) => {
+	const newHours = meridian ? ((hours % 12) + (12 * (meridian === 'PM'))) : hours % 24 ;
```
mmcgahan: Looks like you might not need to use `(hours % 12)` - just `hours`, unless you really want to support `formatHours(15, 'AM') === '03'`.

src/forms/TimeInput.jsx
```diff
@@ -10,17 +44,136 @@ class TimeInput extends React.Component {
...
+	onChange(partialState) {
+		if (this.props.onChange) {
+			const stateValues = {
+				...this.state,
+				...partialState
+			};
+
+			const value = `${formatHours(stateValues[HOURS], stateValues[MERIDIAN])}:${formatDigits(stateValues[MINUTES])}`;
+			this.props.onChange(value);
+		}
```
mmcgahan: nitpick, you can unindent the body of this function by bailing early `if (!this.props.onChange) { return; }`

src/forms/TimeInput.jsx
```diff
@@ -10,17 +44,136 @@ class TimeInput extends React.Component {
...
 	* @param e Event Object
-	* @description called when the input changes, in turn calls the onChange
+	* @description called when the browser-native time input changes, in turn calls the onChange
 	* 	handler prop, if there is one provided (eg supplied by redux-form or DateTimePicker)
 	*/
-	onChange(e) {
-		this.props.onChange && this.props.onChange(e);
+	onTimeInputChange(e) {
+		const { value } = e.target;
+		this.setState(() => ({ value: value }));
```
mmcgahan:
```js
this.setState({ value });
```

src/forms/timeInput.test.jsx
```diff
@@ -1,49 +1,250 @@
...
+				const min = 20;
+				const result = TimeInput.prototype.constrainValue(min, undefined, value);
+				expect(result).toEqual(min);
+			});
+			it('returns the constrained value if max', () => {
+				const value = '10';
+				const max = 5;
+				const result = TimeInput.prototype.constrainValue(undefined, max, value);
+				expect(result).toEqual(max);
+			});
```
mmcgahan: probably good to add a test for the 'no change' case when constraints _are_  provided 
- e.g. 'returns unchanged value if within constraints'

src/forms/timeInput.test.jsx
```diff
@@ -1,49 +1,250 @@
...
 	});
 
-	it('calls internal onChange when value is changed', function() {
-		inputEl.simulate('change', { target: { value: newTime } });
-		expect(onChangeSpy).toHaveBeenCalled();
+	describe('TimeInput, with input[time] support', () => {
+		beforeEach(() => {
+			onChangeSpy = jest.spyOn(TimeInput.prototype, 'onChange');
+			onTimeInputChangeSpy = jest.spyOn(TimeInput.prototype, 'onTimeInputChange');
+			component = mount(<TimeInput {...props} />);
```
mmcgahan: do you need to `mount` vs `shallow`?

unjust: yes

mmcgahan: Looks like it's not too tricky to enable `shallow` rendering for this:
https://github.com/meetup/meetup-web-components/blob/shallow-time-input/src/forms/timeInput.test.jsx#L119-L121
https://github.com/meetup/meetup-web-components/blob/shallow-time-input/src/forms/timeInput.test.jsx#L133

I know there's time pressure on this so it's not a blocker, but `mount` almost _always_ over-tests the component, so it should be avoided. The other uses of `mount` a little more tricky to test using `shallow`, but mainly because they make assumptions about about how `mount` renders elements - e.g. your `blur` simulate calls don't provide an argument, even though your `onBlur` handler expects an argument.

unjust: ok. totally agree - sorry for the short response previously. I just swapped `mount` out for `shallow` and things failed - without much time to rewrite tests.  I'll put in a defect to fix tests.

PR #363, opened 2017-11-08
Media PropType

cranberyxl:
Start of exporting proptypes for properties that are injected by mwc code


mmcgahan:
Actually just remembered that `Media` only has 3 props, which could be specified directly. No biggie

PR #364, opened 2017-11-08
WC-135: formatHours fixes, no hanging

aquillano:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-135

#### Description
Fixes time calc math. Prevents infinite loop.

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -27,7 +27,7 @@ export const getTimeParts = (time) => {
 	};
 };
 const formatHours = (hours, meridian) => {
-	const newHours = meridian ? (parseInt(hours) + (12 * (meridian === 'PM'))) : hours % 24 ;
+	const newHours = meridian ? (parseInt(hours % 12) + (12 * (meridian === 'PM'))) : hours % 24 ;
 	return formatDigits(newHours);
 };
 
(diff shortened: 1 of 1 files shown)


src/forms/TimeInput.jsx
```diff
@@ -58,8 +58,9 @@ class TimeInput extends React.Component {
 	 */
 	parseValueIntoState(value, is24Hr) {
 		const timeParts = getTimeParts(value);
+
 		return {
-			hours: value && (formatDigits(timeParts.hours % (is24Hr ? 24 : 12)) || '12'),
+			hours: value && (formatDigits(timeParts.hours % (is24Hr ? 24 : 12) || (is24Hr ? '00' : '12'))), // TODO test midnight at is24
```
mmcgahan: this is hard to read - probably should be setting more named `const`s before assembling them, but not a blocker

src/forms/TimeInput.jsx
```diff
@@ -27,7 +27,7 @@ export const getTimeParts = (time) => {
 	};
 };
 const formatHours = (hours, meridian) => {
-	const newHours = meridian ? (parseInt(hours) + (12 * (meridian === 'PM'))) : hours % 24 ;
+	const newHours = meridian ? (parseInt(hours % 12) + (12 * (meridian === 'PM'))) : hours % 24 ;
```
swozniak: also not a blocker, but a little too much going on to pick this up at a glance

swozniak: always best practice to include the radix with parseInt()

i.e. `parseInt(hours % 12, 10)`

swozniak:
Couldn't reproduce previous issues in `mup-web` in Safari or Firefox (setting the start Hours to 12 or higher no longer advances the date, which had the side effects of rendering as 00:00 and possibly introducing an infinite loop by advancing the date past the end date).

src/forms/TimeInput.jsx
aquillano: Agreed.

mmcgahan:
@aquillano not sure if you saw my comment, but it's important to add unit tests that ensure the updated behavior doesn't regress - whatever the new code does that the old code _doesn't_ should be validated with a unit test

PR #365, opened 2017-11-08
Wc 136 timeinput callback

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-136

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -188,6 +188,7 @@ class TimeInput extends React.Component {
 			disabled,
 			is24Hr,
 			onChange,	// eslint-disable-line no-unused-vars
+			suppressError, // eslint-disable-line no-unused-vars
 			...other
 		} = this.props;
 
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -69,12 +69,13 @@ class TimeInput extends React.Component {
 
 	/**
 	* @function onChange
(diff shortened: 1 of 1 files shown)


src/forms/TimeInput.jsx
```diff
@@ -69,12 +69,13 @@ class TimeInput extends React.Component {
 
 	/**
 	* @function onChange
-	* @param partialState the information to update state with
+	* @param {Object} e the information to update state with
```
swozniak: Doc comment just looks like a copy/paste from the next line... but why do we need to pass in this event & through to the callback? Where's the callback needed?

aquillano: Same question.

unjust: The callback is used in RsvpTimes in `mup-web`.
RsvpTimes flips another value (`rsvp_times_active`) when an onChange occurs.
It needs other info from the event, not just the value.

unjust: actually i may have a way to not need the event passed to the callback... but do you think its harmful?

mmcgahan: IMO anythin called `onChange` should _always_ be given an `Event` argument, regardless of whether it's used, to stay consistent with native onChange event handler behavior

swozniak: Definitely don't think it's harmful (and will echo @mmcgahan's sentiment), just wanted to know where it was coming from!

PR #368, opened 2017-11-12
Fix prop warning in AccordionPanel

mperrotti:
#### Description
The warning```Expected `onClick` listener to be a function, instead got a value of `boolean` type.``` should no longer be printing in the console


src/interactive/AccordionPanel.jsx
```diff
@@ -39,7 +39,7 @@ class AccordionPanel extends React.Component {
 	 * @param {Event} e - the event object
 	 * @returns {undefined}
 	 */
-	_handleToggle(e){
+	_handleToggle(e, isInidcatorContainer){
```
unjust: spelling

PR #371, opened 2017-11-16
dont set maxlen of field

unjust:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-139

#### Description
display errors but allow people to overtype in input. remove maxlength attr.

screen shot 2017-11-16 at 11 41 14 am

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -73,7 +73,6 @@ const TextInput = (props) => {
 					placeholder={placeholder}
 					className={classNames}
 					onChange={onChange}
-					maxLength={maxLength}
 					pattern={pattern}
 					disabled={disabled}
 					id={id}


nataliaines:
Looks great!! Thanks!

src/forms/TextInput.jsx
```diff
@@ -73,7 +73,6 @@ const TextInput = (props) => {
 					placeholder={placeholder}
 					className={classNames}
 					onChange={onChange}
-					maxLength={maxLength}
```
swozniak: Just to be clear, we still want to support `maxLength` as a prop, but not enforce it on the input element itself? 

Could be confusing for a future dev to encounter, so I wouldn't mind seeing this behavior commented. Looks fine to me otherwise.

unjust: exactly. this is the way textarea behaves and design wants it consistent.

PR #377, opened 2017-11-21
Wc 143/avatar img tag

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-143

#### Description
Given our available options, adding a hidden image tag to `Avatar` is the least disruptive change that will allow printing of member photos.

Later, we _could_ replace the background image with actual image tags with some api changes to use the photo scaler for square cropping.

screen shot 2017-11-21 at 5 23 00 pm
screen shot 2017-11-21 at 5 23 26 pm

### DIFF
--- a/assets/scss/components/_avatar.scss
+++ b/assets/scss/components/_avatar.scss
@@ -5,30 +5,6 @@ name: avatar
 category: UI Components
 ---
 
-```html_example
-<ul class="inlineblockList">
-	<li>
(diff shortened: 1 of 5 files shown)


mperrotti:
+1 for following up with real img tags from the photo scaler.
If we wanted to do this sooner/without API changes, we could use `object-fit: cover` and come up with a fallback for IE and Edge

PR #380, opened 2017-11-28
Wc 145/rm inline base64 loader

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-145

#### Description
Fixes no photo avatar and icon text inputs in `mup-web`.

Removes base64 inline loader; refactors `Icon` and `TextInput` components to use actual `Icon` component instead of inline loading a base64 background image.

Shared components should not rely on inline webpack loaders, as it makes them less portable across applications.

screen shot 2017-11-28 at 1 02 57 pm
screen shot 2017-11-28 at 1 03 00 pm
screen shot 2017-11-28 at 1 03 10 pm

### DIFF
--- a/src/media/AvatarMember.jsx
+++ b/src/media/AvatarMember.jsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
 import Avatar from './Avatar';
+import Icon from './Icon';
 
 export const AVATAR_PERSON_CLASS = 'avatar--person';
 export const AVATAR_PERSON_ORG_CLASS = 'avatar--org';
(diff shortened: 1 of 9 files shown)


src/media/AvatarMember.jsx
```diff
@@ -30,16 +31,22 @@ class AvatarMember extends React.PureComponent {
...
-			<Avatar
-				alt={member.name}
-				src={showNoPhoto ? noPhotoImage : member.photo[photoLink]}
-				className={classNames}
-				{...other}
-			/>
-		);
+		const allProps = {
+			alt: member.name,
+			src: !showNoPhoto && member.photo[photoLink],
```
mmcgahan: I believe you'll get a proptype error when you pass `src=false` in the `showNoPhoto` case
https://github.com/meetup/meetup-web-components/blob/master/src/media/Avatar.jsx#L84

src/media/AvatarMember.jsx
```diff
@@ -30,16 +31,22 @@ class AvatarMember extends React.PureComponent {
...
+			className: classNames,
+		};
+
+		if (showNoPhoto) {
+			return <Avatar {...allProps} {...other}>{noPhotoIcon}</Avatar>;
```
mmcgahan: minor, but instead of 2 return statements, I would recommend adding `children: showNoPhoto && noPhotoIcon` as part of `allProps` and just having a single return statement

akdetrick:
bump; resolved conflicts

PR #384, opened 2017-12-05
Wc 146/char counter a11y

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-146

#### Description
Fields with character counters currently allow screen readers to tab to the characters remaining element, which reads a number out of context.

To fix this and provide better hints, tab navigation to the character counter has been disabled, and the `maxlength` attribute will be used on text areas or text inputs in which the `maxLength` prop is passed. This attribute will alert screen readers to "invalid data" when the max length is exceeded. Depending on the screen reader, the `maxlength` value may also be read to provide a hint.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -91,7 +91,7 @@ const TextInput = (props) => {
 					disabled={disabled}
 					id={id}
 					style={inputStyles}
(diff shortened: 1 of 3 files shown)


mperrotti:
Just needs tests to pass

PR #385, opened 2017-12-06
Wc 148/aria form errors

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-148

#### Description
Accessibility enhancements for form errors. Because we conditionally rendered field errors without an `aria-live` region, assistive technology was not informed of visual UI changes and error states.

This branch introduces an `ErrorList` component and adds attributes to all base form components in order to assist accessibility.

* error container always renders with `aria-live` attribute, which alerts screen readers when field errors appear
* form fields in error state now receive an `aria-invalid="true"` attribute
* when an `id` prop is passed to a form component, it is now used to link the error container to the related field with `aria-describedby` when there is an error

**No significant visual difference from previous error rendering**
screen shot 2017-12-06 at 4 40 01 pm

### DIFF
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -2,6 +2,10 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
 import ErrorList from './ErrorList';
+import {
+	getFieldErrorProps,
+	getErrorListProps,
+} from './util/errorProps';
 
 /**
  * @module CalendarComponent
@@ -113,7 +117,6 @@ class CalendarComponent extends React.Component {
 	render() {
 		const {
 			className,
-			id,
 			name,
 			error,
 			suppressError,
@@ -149,7 +152,7 @@ class CalendarComponent extends React.Component {
 			<div>
 				<span>
 					{label && (
-						<label htmlFor={id} className={classNames.label}>
+						<label htmlFor={name} className={classNames.label}>
 							{label}
 						</label>
 					)}
@@ -160,15 +163,16 @@ class CalendarComponent extends React.Component {
 					}
 					<input
 						type="text"
-						id={id}
+						id={name}
 						name={name}
 						defaultValue={value}
 						className={classNames.field}
 						ref={input => (this.inputEl = input)}
 						{...other}
+						{...getFieldErrorProps(name, !!error)}
 					/>
 				</span>
-				{!suppressError && <ErrorList errors={error} />}
+				{!suppressError && <ErrorList {...getErrorListProps(name, error)} />}
 			</div>
 		);
 	}
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
+import ErrorList from './ErrorList';
 
 /**
  * @module CalendarComponent
@@ -167,8 +168,7 @@ class CalendarComponent extends React.Component {
 						{...other}
 					/>
 				</span>
-				{!suppressError &&
-					error && <p className="text--error text--small">{error}</p>}
+				{!suppressError && <ErrorList errors={error} />}
 			</div>
 		);
 	}
--- a/src/forms/ErrorList.jsx
+++ b/src/forms/ErrorList.jsx
@@ -1,7 +1,8 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import cx from 'classnames';
 
-export const ERROR_CLASSNAMES = 'text--error text--small';
+export const ERROR_CLASSNAME = 'text--error text--small';
 
 const ErrorList = (props) => {
 	const {
@@ -23,7 +24,7 @@ const ErrorList = (props) => {
 		>
 			{
 				errorList.map((err, key) => (
-					<li key={key} className={ERROR_CLASSNAMES}>{err}</li>
+					<li key={key} className={cx(ERROR_CLASSNAME, 'text--small')}>{err}</li>
 				))
 			}
 		</ul>
--- a/src/forms/ErrorList.jsx
+++ b/src/forms/ErrorList.jsx
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export const ERROR_CLASSNAMES = 'text--error text--small';
+
+const ErrorList = (props) => {
+	const {
+		errors,
+		...other
+	} = props;
+
+	const errorList = Array.isArray(errors) ? errors : [errors];
+
+	return (
+		<ul
+			aria-role="alert"
+			aria-live="assertive"
+			{...other}
+		>
+			{
+				errorList.map((err) => (
+					<li className={ERROR_CLASSNAMES}>{err}</li>
+				))
+			}
+		</ul>
+	);
+};
+
+ErrorList.propTypes = {
+	errors: PropTypes.oneOfType([
+		PropTypes.string,
+		PropTypes.array
+	])
+};
+
+export default ErrorList;
--- a/src/forms/ErrorList.jsx
+++ b/src/forms/ErrorList.jsx
@@ -13,13 +13,12 @@ const ErrorList = (props) => {
 
 	return (
 		<ul
-			aria-role="alert"
 			aria-live="assertive"
 			{...other}
 		>
 			{
-				errorList.map((err) => (
-					<li className={ERROR_CLASSNAMES}>{err}</li>
+				errorList.map((err, key) => (
+					<li key={key} className={ERROR_CLASSNAMES}>{err}</li>
 				))
 			}
 		</ul>
--- a/src/forms/ErrorList.jsx
+++ b/src/forms/ErrorList.jsx
@@ -5,11 +5,16 @@ export const ERROR_CLASSNAMES = 'text--error text--small';
 
 const ErrorList = (props) => {
 	const {
+		errorId,
 		errors,
 		...other
 	} = props;
 
 	const errorList = Array.isArray(errors) ? errors : [errors];
+
+	if (errorId) {
+		other['id'] = errorId;
+	}
 
 	return (
 		<ul
@@ -26,6 +31,7 @@ const ErrorList = (props) => {
 };
 
 ErrorList.propTypes = {
+	errorId: PropTypes.string,
 	errors: PropTypes.oneOfType([
 		PropTypes.string,
 		PropTypes.array
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
 import autosize from 'autosize';
+import ErrorList from './ErrorList';
 
 /**
  * @module Textarea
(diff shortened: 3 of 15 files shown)


akdetrick:
The `errorProps.js` utility functions help keep things a bit more DRY, but usage seems a bit awkward. Would a universal base form component be a better way to share `errorId` and prop handling across the different components?

akdetrick:
I'm thinking that maybe the `errorProp` utility functions are doing the job of a higher order component... Would an HOC be the preferred method of sharing this prop assignment logic across form components?

akdetrick:
Marking this as `Prelim` to refactor `errorProp` utilities into a HOC, `WithErrorList`

src/forms/CalendarComponent.jsx
```diff
@@ -1,12 +1,13 @@
...
 import React from 'react';
 import cx from 'classnames';
+import withErrorList from '../utils/components/withErrorList';
 
 /**
  * @module CalendarComponent
  * inits flatpickr js date picker over a text input
 */
-class CalendarComponent extends React.Component {
+export class CalendarComponent extends React.Component {
```
akdetrick: (I'm exporting unwrapped form components like this for tests)

src/forms/ErrorList.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+	const errorList = Array.isArray(errors) ? errors : [errors];
+
+	if (errorId) {
+		other['id'] = errorId;
+	}
+
+	return (
+		<ul
+			aria-live="assertive"
```
mperrotti: Would adding `aria-role="alert"` be too severe?

src/forms/ErrorList.jsx
```diff
@@ -0,0 +1,42 @@
...
+			{...other}
+		>
+			{
+				errorList.map((err, key) => (
+					<li key={key} className={cx(ERROR_CLASSNAME, 'text--small')}>{err}</li>
```
mmcgahan: I'm pretty sure you don't need `key` in React16, but fine to leave as-is

src/forms/ErrorList.jsx
```diff
@@ -0,0 +1,42 @@
...
+export const ERROR_CLASSNAME = 'text--error text--small';
+
+const ErrorList = (props) => {
+	const {
+		errorId,
+		errors,
+		...other
+	} = props;
+
+	const errorList = Array.isArray(errors) ? errors : [errors];
```
mmcgahan: IMO `errors` should always be an array - no need to be super adaptable here, and it's a plural prop so sending a single string would be a little confusing to future maintainers

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,103 @@
...
+	 * @module WithErrorList
+	 */
+	class WithErrorList extends React.PureComponent {
+		/**
+		 * @constructor
+		 * @param {Object} - React element props
+		 */
+		constructor(props) {
+			super(props);
+		}
```
mmcgahan: no need for `constructor` here

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,103 @@
...
+			} = this.props;
+
+			return this.props.suppressError ? <WrappedComponent /> : (
+				<div>
+					<WrappedComponent {...this.props} {...getFieldErrorProps(id, !!error)} />
+					<ErrorList {...getErrorListProps(id, error)} />
+				</div>
+			);
+		}
+	}
```
mmcgahan: make this a pure functional component

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,103 @@
...
+		 */
+		render() {
+			const {
+				id,
+				error,
+			} = this.props;
+
+			return this.props.suppressError ? <WrappedComponent /> : (
+				<div>
+					<WrappedComponent {...this.props} {...getFieldErrorProps(id, !!error)} />
```
mmcgahan: the 'getters' in this component make it feel extra verbose and non-linear - I think it would be simpler to read and use if everything was just in the `render` function directly:

```jsx
const fieldProps = { 'aria-invalid': Boolean(error) }
const listProps = { errors: error }; // this looks wonky to me
if (id) {
  const errorId = `${id}-error`;
  fieldProps['aria-describedby'] = errorId;
  listProps.errorId = errorId;
}
// ...
```

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,103 @@
...
+					<ErrorList {...getErrorListProps(id, error)} />
```
mmcgahan: this `error` argument is consumed as `errors`, which makes for an ambiguous data type - is it an array or a string? Better to lock it down instead of implicitly supporting both

src/utils/components/withErrorList.test.jsx
```diff
@@ -0,0 +1,66 @@
...
+const MOCK_ERROR_MSG = 'You are not Keith Hernandez';
+
+/**
+ * @class TestComponent
+ */
+class TestComponent extends React.Component {
+	render() {
+		return <h1>Hello world</h1>;
+	}
+}
```
mmcgahan: pure functional alternative:

```jsx
const TestComponent = () => <h1>Hello world</h1>;
```

src/forms/CalendarComponent.jsx
```diff
@@ -145,31 +143,27 @@ class CalendarComponent extends React.Component {
...
+						{helperText}
+					</div>
+				}
+				<input
+					type="text"
+					id={name}
+					name={name}
+					defaultValue={value}
+					className={classNames.field}
+					ref={input => (this.inputEl = input)}
```
mmcgahan: I think there's an AirBnB eslint rule that disallows variable assignment inside return values - might want to clean that up now

src/forms/ErrorList.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+	if (errorId) {
+		other['id'] = errorId;
```
mmcgahan: ESLint might not be set up on this, but we're using a rule that will error when using square brackets unnecessarily in mup-web - this is equivalent to `other.id = errorId`

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param {string} id - the id of the form field
+ * @param {string|array} errors - single or multiple error messages
+ * @returns {object} props object to apply to ErrorList
+ */
+export const getErrorListProps = (id, errors) => {
+	const props = { errors };
+
+	if (id) {
+		props['errorId'] = getErrorId(id);
```
mmcgahan: unnecessary square brackets

src/forms/ErrorList.jsx
akdetrick: we could add `role="alert"` to the list items, but it would be redundant with the surrounding `assertive` live region

src/forms/CalendarComponent.jsx
akdetrick: added a file-level disable for that rule

mmcgahan: I recommend fixing the code rather than disabling the rule - it's useful outside of this line.

akdetrick:
I ended up with enforcing an `error` string on `ErrorList` and `withErrorList`.

It would be nice to support an `array` instead, but `redux-form` will always pass a single `error`. We can add support for multiple errors to `ErrorList` later if need be (or if we begin using a different form manager).

akdetrick:
bump

akdetrick:
Adding `Prelim` tag until an issue with `TimeInput` is resolved. The `withErrorList` HOC appears to cause issues in the `mup-web` schedule form.

akdetrick:
Addressed issues with `withErrorList` props not being passed down to wrapped component; tested in `mup-web` and we're looking good again.

src/forms/Textarea.jsx
```diff
@@ -138,35 +139,32 @@ class Textarea extends React.Component {
...
+				}
+				<textarea
+					type='text'
+					name={name}
+					required={required}
+					className={classNames.textarea}
+					onChange={this.onChange}
+					rows={rows}
+					ref={(textarea) => {this.textarea = textarea;}}
+					style={{ ...style, ...heightConstraints }}
```
mmcgahan: nitpick, but you could just set `style.min/maxHeight` explicitly instead of creating a new `heightConstraints` object and then spreading `style` and `heightConstraints` here

src/forms/errorList.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+		<ErrorList error={MOCK_ERROR} />
+	);
+
+	it('matches snapshot for single error', () => {
+		expect(componentWithId).toMatchSnapshot();
+	});
+	it('matches snapshot for input without id', () => {
+		expect(componentWithoutId).toMatchSnapshot();
+	});
+	it('applies id attribute when `errorId` provided', () => {
```
mmcgahan: this is tested by snapshot

src/forms/errorList.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+		expect(componentWithId.props()).toHaveProperty('id');
+		expect(componentWithId.prop('id')).toEqual(MOCK_ERROR_ID);
+	});
+	it('does NOT apply id attribute when `errorId` is not provided', () => {
```
mmcgahan: tested by snapshot

src/forms/errorList.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+		expect(componentWithoutId.props()).not.toHaveProperty('id');
+	});
+	it('list of errors has `aria-live` attribute', () => {
```
mmcgahan: tested by snapshot

src/utils/components/withErrorList.jsx
```diff
@@ -0,0 +1,76 @@
...
+				...other
+			} = this.props;
+
+			const errorId = id && getErrorId(id);
+
+			// always pass `aria-invalid`, `id`,
+			// and `error` props to wrapped component
+			other.id = id;
+			other.error = error;
+			other['aria-invalid'] = Boolean(error);
```
mmcgahan: I strongly recommend passing these 3 props explicitly into `<WrappedComponent ..>` rather than mutating the opaque `other` object - same effect, but it makes the JSX much easier to scan because the reader doesn't have to think about `other` being modified

src/utils/components/withErrorList.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+		const actual = getErrorId(MOCK_FIELD_ID);
+		const expected = 'name-error';
+		expect(actual).toBe(expected);
+	});
+});
+
+describe('withErrorList wrapped component', () => {
+
+	it('passes `error` to the `ErrorList`', () => {
+		const component = mount(
```
mmcgahan: do you need `mount` here? `shallow` should cover anything that the HOC is responsible for - e.g. the HOC doesn't specify that the `<ErrorList>` needs to use an `li` element to display the error, so this test doesn't need to look for `li`. Instead, just validate that the props for `ErrorList` are the expected values (or take a snapshot)

src/utils/components/withErrorList.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+			<TestComponentWithErrorList
+				id={MOCK_FIELD_ID}
+				error={MOCK_ERROR_MSG}
+			/>
+		);
+		expect(component.find('li').text()).toBe(MOCK_ERROR_MSG);
+	});
+
+	it('provides TRUE `aria-invalid` prop when wrapped component has `error`', () => {
+		const component = mount(
```
mmcgahan: I think all of these `mount`s could be `shallow`, and you might be able to consolidate a few tests with a snapshot or two

akdetrick:
@mmcgahan - how are we feeling about this one? I gave this another test in `mup-web` and it's looking good there

PR #389, opened 2017-12-12
Render dropdown content in document body

mperrotti:
Re Prelim:
- ~Still needs tests to be updated~
- ~Need to test in mup-web~

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-152

#### Description
Dropdown content needs to be rendered in the document body to avoid issues with `z-index` and `overflow`

### DIFF
--- a/assets/scss/components/_dropdown.scss
+++ b/assets/scss/components/_dropdown.scss
@@ -3,7 +3,6 @@
 //
 $dropdownShadow: $interactiveShadow;
 $content-minWidth: $block-6;
-// $content-maxWidth: 512px; // match chapstick
 $triangleSize: 10px;
 $spaceFromTrigger: $space-3 + $triangleSize;
 
--- a/assets/scss/components/_dropdown.scss
+++ b/assets/scss/components/_dropdown.scss
@@ -3,7 +3,7 @@
 //
 $dropdownShadow: $interactiveShadow;
 $content-minWidth: $block-6;
-$content-maxWidth: 512px; // match chapstick
+// $content-maxWidth: 512px; // match chapstick
 $triangleSize: 10px;
 $spaceFromTrigger: $space-3 + $triangleSize;
 
@@ -29,10 +29,11 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	background-color: $C_contentBG;
 	box-shadow: $dropdownShadow;
 	box-sizing: border-box;
+	margin-top: $triangleSize;
 	position: absolute;
 	top: $spaceFromTrigger;
-	min-width: $content-minWidth;
-	max-width: $content-minWidth;
+	// min-width: $content-minWidth;
+	// max-width: $content-minWidth;
 	z-index: map-get($zindex-map, 'popup');
 
 	// outer triangle for border
--- a/assets/scss/components/_dropdown.scss
+++ b/assets/scss/components/_dropdown.scss
@@ -32,9 +32,6 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	margin-top: $triangleSize;
 	position: absolute;
 	top: $spaceFromTrigger;
-	// min-width: $content-minWidth;
-	// max-width: $content-minWidth;
-	z-index: map-get($zindex-map, 'popup');
 
 	// outer triangle for border
 	&:before {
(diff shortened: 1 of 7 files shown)


mperrotti:
Doing a more thorough check in `mup-web`, but feel free to review

assets/scss/components/_dropdown.scss
```diff
@@ -3,7 +3,7 @@
 //
 $dropdownShadow: $interactiveShadow;
 $content-minWidth: $block-6;
-$content-maxWidth: 512px; // match chapstick
+// $content-maxWidth: 512px; // match chapstick
```
burrata: should we remove this?

assets/scss/components/_dropdown.scss
```diff
@@ -29,11 +29,9 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	background-color: $C_contentBG;
 	box-shadow: $dropdownShadow;
 	box-sizing: border-box;
+	margin-top: $triangleSize;
 	position: absolute;
 	top: $spaceFromTrigger;
-	min-width: $content-minWidth;
```
burrata: 🙌  there's some `min-width: auto`s in mup-web that we can take out now.

src/interactive/Dropdown.jsx
```diff
@@ -72,11 +106,15 @@ class Dropdown extends React.PureComponent {
 	componentDidMount() {
 		document.body.addEventListener('click', this.onBodyClick);
 		document.body.addEventListener('keydown', this.onBodyKeyDown);
+		window.addEventListener('resize', this.getContentPosition);
+		document.addEventListener('scroll', this.getContentPosition, true);
```
mperrotti: Might be good to figure out how to do this only when the dropdown is opened

src/interactive/Dropdown.jsx
```diff
@@ -11,14 +13,44 @@ class Dropdown extends React.PureComponent {
...
+				case 'center':
+					return left + (width/2);
+				default:
+					return left - (contentWidth - width);
+			}
+		};
+
+		const ddPosition = {
+			x: getCoordX(this.props.align),
+			y: window.scrollY + top + height
```
mperrotti: Note for myself: `window.scrollY` may not work in all browsers

src/interactive/Dropdown.jsx
```diff
@@ -30,6 +62,8 @@ class Dropdown extends React.PureComponent {
 	}
 
 	toggleContent(e) {
+		this.getContentPosition(e);
```
mperrotti: Note for myself: `getContentPosition` doesn't take an arg

src/interactive/Dropdown.jsx
```diff
@@ -133,25 +172,38 @@ class Dropdown extends React.PureComponent {
...
-				</div>
+				<Portal isOpened={isActive}>
+					<div
+						ref={(el) => this.contentRef = el}
+						className={classNames.content}
+						aria-hidden={!isActive}
+						style={{
+							left: `${this.state.posX}px`,
+							top: `${this.state.posY}px`,
+							minWidth: `${width}px`,
```
burrata: Should these both be props and not include `px`? I can imagine wanting to have a separate `minWidth` and `maxWidth`, and for the values possibly to be `auto`.

src/interactive/Dropdown.jsx
akdetrick: The listener is fine to leave in place while the component is mounted, but I'd recommend throttling, debouncing, or `requestAnimationFrame` on the `resize` and `scroll` handler functions

src/interactive/Dropdown.jsx
```diff
@@ -133,25 +173,46 @@ class Dropdown extends React.PureComponent {
...
-				</div>
+				<Portal isOpened={isActive}>
+					<div
+						ref={(el) => this.contentRef = el}
+						className={classNames.content}
+						aria-hidden={!isActive}
+						style={{
+							left: `${this.state.posX}px`,
+							top: `${this.state.posY}px`,
+							minWidth: `${minWidth}`,
```
burrata: very minor, but this can now just be `minWidth` and the below prop `maxWidth`.

PR #400, opened 2017-12-15
Optional portalling for Dropdown

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-491

#### Description
Sometimes we don't want a dropdown to portal. E.g.: if a <Dropdown> contains components that also portal

### DIFF
--- a/assets/scss/components/_dropdown.scss
+++ b/assets/scss/components/_dropdown.scss
@@ -75,8 +75,18 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	}
 }
 
+.dropdown--noPortal {
+	.dropdown-content--left {
+		left: 0;
+	}
+	.dropdown-content--center {
+		left: 50%;
+	}
+}
+
 .dropdown-content--right {
 	@include alignDropdown("right");
+	right: 0px;
 	transform: translateX(-100%);
 	-webkit-transform: translateX(-100%);
 }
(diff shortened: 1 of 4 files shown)


assets/scss/components/_dropdown.scss
```diff
@@ -88,9 +98,11 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	transform: translateX(-50%);
 
 	&:before {
-		transform: translateX(50%);
+		left: 50%;
```
akdetrick: Should this `left` property be conditionally applied for the `--noPortal` variant only?

akdetrick: Ok, so it looks like JS overrides this when the dropdown is in a Portal... Just confirming this is right

mperrotti: yes, it does

src/interactive/Dropdown.jsx
```diff
@@ -230,7 +243,8 @@ Dropdown.propTypes = {
 	isActive: PropTypes.bool,
 	manualToggle: PropTypes.func,
 	maxWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-	minWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+	minWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+	shouldPortal: PropTypes.bool,
```
mmcgahan: i know this is already merged, but `noPortal` seems like it would be a better prop name because that's what the CSS classname variant uses, and since by default it _will_ portal, it allows callers to opt-in to 'no portalling' with `<Dropdown noPortal />`

PR #402, opened 2017-12-18
Vertical tab support

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-159

#### Description
Adds support for vertical tabs

screen shot 2017-12-18 at 5 17 26 pm

### DIFF
--- a/assets/scss/components/_tabs.scss
+++ b/assets/scss/components/_tabs.scss
@@ -87,10 +87,11 @@ Class                 | Description
 	}
 
 	.tabs--isVertical & {
+		border-right: 3px solid transparent;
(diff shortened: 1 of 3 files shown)


assets/scss/components/_tabs.scss
```diff
@@ -80,10 +85,25 @@ Class                 | Description
 			border-top-color: $C_textPrimaryInverted;
 		}
 	}
+
+	.tabs--isVertical & {
+		border-right: 3px solid transparent;
```
akdetrick: `3px` would be good to define as a local var in this file.

PR #403, opened 2017-12-18
Wc 158/ie11 text inputs

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-158

#### Description

- `TextInput` `value` prop should only be added to the `input` element if the prop is defined. This works around the wonky IE11 event model.
- `Textarea` should **not** have a `maxLength` prop that defaults to `-1`. IE11 will treat a value of `-1` as `0` for the `maxlength` attribute on a text area, which prevents users from interacting with the input. As a workaround, this branch conditionally sets the `maxLength` prop rather than falling back to `-1` as a default.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -68,6 +68,13 @@ const TextInput = (props) => {
 		paddingLeft: `${paddingSize}px`
 	};
 
+	// WC-158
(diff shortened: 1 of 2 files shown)


akdetrick:
Marking `Prelim` while I verify this fix works well in `mup-web` and `pro-web` with and without `redux-form`.

PR #410, opened 2017-12-28
Added requestAnimationFrame utility

mperrotti:
Re Prelim tag:
* Not sure if third-party util is the right way to go
* `this.scheduleUpdate` is defined in the constructor instead of just being bound. The only other place we do this is `AccordionPanel`, but it wouldn't work any other way - is that ok?'

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-154

#### Description
I decided to go with a third-party package instead of rolling our own. React references it in their documentation, so I think it's safe.
I also found and fixed a defect where `dropdown-content` would be incorrectly when scrolling horizontally and portalling

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -2,18 +2,9 @@ import PropTypes from "prop-types";
 import React from "react";
 import cx from "classnames";
 import Portal from "react-portal";
+import rafSchedule from 'raf-schd';
 
 import bindAll from "../utils/bindAll";
-
-const throttle = (fn, wait) => {
-	let time = Date.now();
-	return () => {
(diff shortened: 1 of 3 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -77,6 +78,12 @@ class Dropdown extends React.PureComponent {
 		}));
 	}
 
+	scheduleContentPosition(triggerClientRect) {
```
akdetrick: why not call `scheduleUpdate` directly?

mperrotti: Because then I can't clean it up on unmount with `this.scheduleUpdate.cancel();`

PR #411, opened 2017-12-29
withMatchMedia: use initialMedia prop if provided in constructor

jaclynj:
#### Related issues
https://meetup.atlassian.net/browse/MW-2907

#### Description
Related to / supports this PR https://github.com/meetup/meetup-web-platform/pull/423/files
If initialMedia prop is present, set that in the constructor

### DIFF
--- a/src/utils/components/withMatchMedia.jsx
+++ b/src/utils/components/withMatchMedia.jsx
@@ -44,7 +44,7 @@ const withMatchMedia = WrappedComponent => {
 			super(props);
 
 			this.state = {
-				media: {},
+				media: props.initialMedia || {},
 			};
 
 			this.handleMediaChange = this.handleMediaChange.bind(this);


swozniak:
lgtm! 👍

src/utils/components/withMatchMedia.jsx
```diff
@@ -44,7 +44,7 @@ const withMatchMedia = WrappedComponent => {
 			super(props);
 
 			this.state = {
-				media: {},
+				media: props.initialMedia || {},
```
swozniak: when would this be instantiated with/without props.initialMedia?

jaclynj: I created a wrapper component that passes initial media here: https://github.com/meetup/meetup-web-platform/pull/423/files#diff-e8b9f22f4184c5ccc6b38a7e5e3e7799R24
But this is being used in a lot of places without the new wrapper component, so if it it's used on its own, it will not have initial media.

mmcgahan:
I think this component should get proptypes at some point - or get flow-typed 😉

PR #412, opened 2017-12-29
remove initialMedia prop from dom elements

jaclynj:
#### Description
Do not pass initialMedia as a prop to the wrapped element. Causes errors in the browser dev console.

### DIFF
--- a/src/utils/components/withMatchMedia.jsx
+++ b/src/utils/components/withMatchMedia.jsx
@@ -104,7 +104,9 @@ const withMatchMedia = WrappedComponent => {
 		 * @returns {React.element}
 		 */
 		render() {
-			return <WrappedComponent {...this.props} media={this.state.media} />;
+			const props = {...this.props};
+			delete props.initialMedia;
+			return <WrappedComponent {...props} media={this.state.media} />;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/utils/components/withMatchMedia.jsx
```diff
@@ -104,7 +104,9 @@ const withMatchMedia = WrappedComponent => {
 		 * @returns {React.element}
 		 */
 		render() {
-			return <WrappedComponent {...this.props} media={this.state.media} />;
+			const props = {...this.props};
+			delete props.initialMedia;
```
mmcgahan: The `delete` keyword should generally be avoided because it prevents some V8 compiler optimizations, so this should probably be written as

```jsx
render() {
  const { initialMedia, ...props } = this.props;  // eslint-disable-line no-unused
  return <WrappedComponent {...props} media={this.state.media} />;
}
```

jaclynj: Ha I was going to do it that way until eslint complained

mmcgahan: yeah there's an old ticket in the ESLint repo about this behavior and how it shouldn't be warning in this case, but because object spread still isn't an official syntax spec (still Stage 3 (of 4)), the ESLint maintainers don't want to change the `no-unused` behavior yet :/

PR #414, opened 2018-01-03
Reduced zindex of dropdown, fixed noPortal bugs

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-3142

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -227,7 +227,6 @@ class Dropdown extends React.PureComponent {
 						aria-hidden={!isActive}
 						style={{
 							left: this.state.left,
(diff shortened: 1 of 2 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -227,7 +227,6 @@ class Dropdown extends React.PureComponent {
 						aria-hidden={!isActive}
 						style={{
 							left: this.state.left,
-							right: align === 'right' && noPortal && `-${maxWidth}`,
```
akdetrick: Looks like maybe this was branched from something not up to date with `master`?

mperrotti: Nope, that line was intentionally removed. It was causing weirdness with right-aligned modals that weren't portalled

PR #416, opened 2018-01-05
Allow TextInput value prop pass even if empty string

mperrotti:
#### Related issues

#### Description
This issue came up with Mike A trying to clear a form input in `mup-web`. Passing an empty string to `value` prevents it from getting passed through because it `value` will return false

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -71,7 +71,7 @@ const TextInput = (props) => {
 	// WC-158
 	// Only add a `value` prop if it is defined.
 	// Workaround for IE11 support (see ticket)
-	if (value || value === '') {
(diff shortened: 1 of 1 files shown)


src/forms/TextInput.jsx
```diff
@@ -71,7 +71,7 @@ const TextInput = (props) => {
 	// WC-158
 	// Only add a `value` prop if it is defined.
 	// Workaround for IE11 support (see ticket)
-	if (value) {
+	if (value || /^ *$/.test(value)) {
```
akdetrick: `/^ *$/.test('asdf')` will return `false` with this regex.

We could instead just check that a string of any kind has been passed for `value`:

```
if (value || typeof value === 'string') {...}
```

PR #417, opened 2018-01-07
Wc 147/date picker a11y

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC0147

#### Description
Replaces `flatpickr` standalone usage in `CalendarComponent` with the `react-flatpickr` React component. This enables keyboard navigation and updates underlying `flatpickr` version.


akdetrick:
`Prelim` until usage is verified working in `mup-web`

akdetrick:
✅  verified this works in `mup-web`

src/forms/CalendarComponent.jsx
```diff
@@ -3,151 +3,90 @@ import React from 'react';
...
+					'field--error': Boolean(error)
+				},
 				className
 			)
 		};
 
+		const options = {
+			altInput: true,
+			altFormat: 'D M d, Y',
+			nextArrow: `<span class="svg svg--chevron-right">
```
mperrotti: Do we have to use markup? Can't we just return it from the `<Icon/>` component?

akdetrick: I believe we do; these options get passed directly to `flatpickr` and we need to pass an html string

mperrotti: Ok, I figured it was something like that

PR #420, opened 2018-01-27
WC-169: swallow input if value takes in an empty string

browne0:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-169

#### Description
In a time input component, the time is first set as an empty string. We only care about the valid time, so return out.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Validation

Links back to https://github.com/meetup/mup-web/pull/2540

bug

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -20,16 +20,16 @@ const formatDigits = (number) => `0${number}`.slice(-2);
  * @returns {String|Object} the only parsed value or all the parsed parts if no part defined
  */
 export const getTimeParts = (time) => {
-	const [ hours, minutes ] = time.split(':');
-	return {
-		hours,
-		minutes,
-		meridian: hours < 12 ? 'AM': 'PM',
-	};
+  const [hours, minutes] = time.split(':');
+  return {
+    hours,
+    minutes,
+    meridian: hours < 12 ? 'AM' : 'PM',
+  };
 };
 const formatHours = (hours, meridian) => {
-	const newHours = meridian ? (parseInt(hours % 12) + (12 * (meridian === 'PM'))) : hours % 24 ;
-	return formatDigits(newHours);
+  const newHours = meridian ? (parseInt(hours % 12) + (12 * (meridian === 'PM'))) : hours % 24;
+  return formatDigits(newHours);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


src/forms/TimeInput.jsx
```diff
@@ -20,56 +20,56 @@ const formatDigits = (number) => `0${number}`.slice(-2);
  * @returns {String|Object} the only parsed value or all the parsed parts if no part defined
  */
 export const getTimeParts = (time) => {
-	const [ hours, minutes ] = time.split(':');
-	return {
-		hours,
-		minutes,
-		meridian: hours < 12 ? 'AM': 'PM',
-	};
+  const [hours, minutes] = time.split(':');
```
mmcgahan: looks like some bad Prettier reformatting - everything should remain tabs, not spaces - make sure you check your PRs when you first set them up to catch little things like this

src/forms/TimeInput.jsx
```diff
@@ -78,321 +78,326 @@ export class TimeInput extends React.Component {
...
 	* @param e Event Object
 	* @description called when the browser-native time input changes, in turn calls the onChange
 	* 	handler prop, if there is one provided (eg supplied by redux-form or DateTimePicker)
 	*/
-	onTimeInputChange(e) {
-		const { value } = e.target;
-		this.setState(() => ({ value }));
+  onTimeInputChange(e) {
+    const { value } = e.target;
+    // In a time input component, the time is first set as an empty string. We only care about the valid time, so return out.
```
mmcgahan: What does 'the time is first set as an empty string' mean? In the problem we were seeing, the `value` attribute of the element was already set (e.g. `<input type="time" value="18:00" />`, so it definitely wasn't empty at first. The problem we were seeing was `value` getting set to the empty string _when a user manually entered `0` in the hours field_ - something like `'time inputs are allowed to return an empty string for invalid input - ignore all invalid input'`

sort of nitpick but 'return out' is sort of an ambiguous turn of phrase. 'return immediately' or 'exit early' are a bit more clear.

src/forms/TimeInput.jsx
browne0: Fixed it, that was an editor setting. I assumed it had the pre commit hooks to reformat like mup-web, my bad.

src/forms/TimeInput.jsx
browne0: done

src/forms/TimeInput.jsx
```diff
@@ -20,56 +20,56 @@ const formatDigits = (number) => `0${number}`.slice(-2);
  * @returns {String|Object} the only parsed value or all the parsed parts if no part defined
  */
 export const getTimeParts = (time) => {
-	const [ hours, minutes ] = time.split(':');
-	return {
-		hours,
-		minutes,
-		meridian: hours < 12 ? 'AM': 'PM',
-	};
+    const [hours, minutes] = time.split(':');
```
mmcgahan: still using spaces, not tabs

src/forms/TimeInput.jsx
```diff
@@ -100,6 +100,11 @@ export class TimeInput extends React.Component {
 	*/
 	onTimeInputChange(e) {
 		const { value } = e.target;
+		// Time inputs are allowed to return an empty string for invalid input.
+		// We should ignore all invalid input and return immediately
+		// Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Validation
+		if (value === '') return;
```
aquillano: 👍

PR #421, opened 2018-02-01
Deprecate `<Popover>`

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MW-3243
Fixes https://meetup.atlassian.net/browse/MW-3242

#### Description
The original reason for this work was to improve a11y on dropdown menus and begin to deprecate `<Popover>`, who's use-cases can be handled by adding props to `<Dropdown>`. As a side-effect, it helps Yanyi with a ticket that involves upgrading `react-portal` in MWC.

screen shot 2018-02-01 at 1 30 32 pm


mperrotti:
There's some work that needs to be done in `mup-web` to support the use of `react-portal: ^4.x.x`, but everything I've checked so far appears to be working 👍

src/interactive/dropdown.test.jsx
```diff
@@ -62,16 +57,9 @@ describe('Dropdown', () => {
 		/>
 	);
 	const wrapper = mount(dropdownJSX);
-	const portalWrapper = wrapper.find(Portal);
-
-	it('renders into DOM', () => {
-		expect(wrapper).toMatchSnapshot();
```
mperrotti: Checking snapshots gave me an error after these updates:
```enzyme RangeError: Invalid string length```

I coudn't find a good way to fix it, so I'm testing everything else instead. I'd love some recommendations on how we can make sure we're getting good coverage

File: src/interactive/dropdown.test.jsx [hunk 2]
```diff
@@ -1,12 +1,11 @@
...
-import Portal from 'react-portal';
-
+import { mount } from 'enzyme';
+import Downshift from 'downshift';
 import Section from '../layout/Section';
 import Chunk from '../layout/Chunk';
 import Button from '../forms/Button';
 
-import Dropdown from './Dropdown';
+import Dropdown, {Item} from './Dropdown';
```
burrata: super nitpick, prefer `{ Item }` for readability :)

src/interactive/Dropdown.jsx
```diff
@@ -1,12 +1,30 @@
...
+import {
+	C_COOLGRAYLIGHTTRANSP,
+	C_WHITE
+} from 'swarm-constants/dist/js/colorConstants.js';
 
 import bindAll from "../utils/bindAll";
 
-const ConditionalWrap = ({condition, wrap, children}) => condition ? wrap(children) : children;
+const ConditionalWrap = ({condition, wrap, children}) => {
+	return(condition ? wrap(children) : children);
```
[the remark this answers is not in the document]
mperrotti: No, I accidentally left it in there while debugging

akdetrick:
With the launch of `MW-3242_rsvpBug`, the dependency on the `react-portal` upgrade does not need to be treated as a breaking change.

PR #422, opened 2018-02-02
remove inline base64 img uri from avatar styles

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-162

#### Description
- removed unused CSS for `avatar--icon`
- removed base64 uri from avatar styles (replaces org badge img with badge svg icon)
- refactored `avatarCharm` placeholder into a mixin that positions an actual `Icon` component
- updated stories for `AvatarMember` to show available sizes

screen shot 2018-02-02 at 4 08 58 pm
screen shot 2018-02-02 at 4 09 06 pm

### DIFF
--- a/assets/scss/components/_avatar.scss
+++ b/assets/scss/components/_avatar.scss
@@ -49,6 +49,7 @@ $avatarSizeProps: (height, width);
 		width: 44%;
 		height: 44%;
 
+		// circle behind the icon
 		&:before {
 			position: absolute;
 			top: 0;
@@ -60,6 +61,7 @@ $avatarSizeProps: (height, width);
 			border-radius: 100%;
 		}
 
+		// the icon itself
 		svg {
 			@include color-svg(getPrimaryTextColor($color));
 			width: 80%;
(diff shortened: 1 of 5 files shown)


assets/scss/components/_avatar.scss
```diff
@@ -148,7 +154,7 @@ $avatarIconSizeProps: (height, line-height, width);
 	text-indent: initial;
 	background-color: darken($avatarBg, 6%);
 
-	.svg {
+	.svg--profile {
```
mmcgahan: does this still need to be nested under `.avatar--noPhoto`?

akdetrick: Unfortunately, yes, until we move to CSS Modules. There's a chance we may use an `Icon shape="profile"` outside of an avatar, and disqualifying these rules from `.avatar--noPhoto` could cause side effects.

assets/scss/components/_avatar.scss
```diff
@@ -40,37 +33,50 @@ Class                     | Description
...
 
-	&:after {
-		background-position: center;
-		background-repeat: no-repeat;
-		background-size: 10px;
-		border-radius: 999px;
+	// Use proportionally smaller badge for larger avatars
+	&.avatar--large .svg--avatarBadge,
+	&.avatar--xxlarge .svg--avatarBadge {
+		width: 30%;
```
mperrotti: Is there a reason this is `30%` instead of a more specific value (e.g.: `{n}px`)

akdetrick: Yep - using a `%` relative to the parent makes this work with media-scaling avatars

PR #423, opened 2018-02-08
Shared <Footer> component

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-171

#### Description
`pro-web` wants to share a Footer with `mup-web`, so I refactored the `<Footer>` from `mup-web` to be reused across our apps

screen shot 2018-02-08 at 6 20 50 pm


.storybook/webpack.config.js
```diff
@@ -3,6 +3,7 @@ const path = require('path');
 const SCSS_PATH = path.resolve(__dirname, '../assets', 'scss');
 const CSS_PATH = path.resolve(__dirname, '../assets', 'css');
 const SRC_PATH = path.resolve(__dirname, '../src');
+const IMG_PATH = path.resolve(__dirname, '../assets', 'images');
```
akdetrick: Are the images here just for the sake of the story? I'd recommend using placeholder images to avoid confusion about where to get app download images

src/Footer.jsx
```diff
@@ -0,0 +1,239 @@
...
+			}}
+			className="text--secondary text--small"
+			itemClassNames="footer-item"
+			items={props.items}
+		/>
+	</Chunk>
+);
+
+
+export const SocialIconsList = (language) => {
```
akdetrick: I'd change the arg `language` to `locale` to make it clear that the function is expecting a locale code

src/Footer.jsx
```diff
@@ -0,0 +1,239 @@
...
+		<a href={socialLinks.medium}>
+			<Icon className="margin--left" shape="external-medium" size="s" />
+		</a>,
+	];
+
+	return <InlineBlockList items={socialIcons} />;
+};
+
+// this should probably be folded into mwp-cli, where lang config is kept
+const languageMap = {
```
akdetrick: I'd recommend putting this in `src/utils/` until we get around to moving it to `mwp-cli`

.storybook/webpack.config.js
mperrotti: Images are only for the story. I can get them somewhere else

src/Footer.jsx
```diff
@@ -0,0 +1,239 @@
...
+	'pt-BR': 'Português',
+	'tr-TR': 'Türkçe',
+	'th-TH': 'ไทย',
+	'ja-JP': '日本語',
+	'ko-KR': '한국어',
+	'ru-RU': 'Русский',
+};
+
+/*
+ * Turn list of languages into an array of options for select dropdown
```
akdetrick: very minor style nitpick; the comment here would be unnecessary if the map was happening closer to where it's being used:

```
options={locales
     .map(lang => ({
          label: languageMap[lang],
	      value: lang,
     })
)}
```

src/Footer.jsx
mperrotti: It actually takes a language abbreviation (`en`), not a locale code (`en-US`). It _could_ be changed to take a locale code though

src/Footer.jsx
mperrotti: This makes the test `it('should render the correct language options')` more difficult to test for. I still have 100% coverage after removing that test, but I'm open to alternatives for testing if you think it's important to pass data to `options` this way **and** test to make sure the options match

src/Footer.jsx
akdetrick: my concern here is that `getSocialLinks` expects a locale

mperrotti: You're right. I was mixing this up for the way `AppBadges` works in mup-web 😳

PR #425, opened 2018-02-09
Hide left gradient on hscroll until scrolled

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-168

Before:
screen shot 2018-02-09 at 2 59 12 pm

After:
screen shot 2018-02-09 at 2 59 03 pm

### DIFF
--- a/src/Hscroll.jsx
+++ b/src/Hscroll.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import cx from 'classnames';
+import rafSchedule from 'raf-schd';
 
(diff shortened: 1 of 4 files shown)


src/Hscroll.jsx
```diff
@@ -12,6 +14,21 @@ export const VALID_BREAKPOINTS = {
...
+	constructor(props){
+		super(props);
+		this.onScroll = this.onScroll.bind(this);
+
+		this.state = {
+			isScrolled: false
+		};
+	}
+
+	onScroll(e) {
```
akdetrick: This looks like it would benefit from `raf` (requestAnimationFrame util)

PR #427, opened 2018-02-12
Nav Items (MP-351)

kcode20:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-351

#### Description
Since we want to unify the navigation bar across all of web, create the navigation bar in Meetup-Web Components for easy implementations in any web repo.
Part 1: Create a component to create Navigation Items

DoD:
* Create component and story displaying use
* Nav Item can fulfill pro-web and mup-web's current design

#### Comments
I just copied over mup-web's navItem's component, tests, and styling into meetup-web-components: 
* NavItems: 
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/HeaderNavItem.jsx
* NavItem Tests:
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/headerNavItem.test.jsx
* SCSS for NavItems: 
https://github.com/meetup/mup-web/blob/master/src/assets/scss/shared/chrome/_header.scss

The only thing that is left out of these files was Internationalization. This will be addressed in a later PR. The NavBar story and component file was created to test if the NavItem was working as expected and will be built out further in following tasks.

### DIFF
--- a/assets/scss/components/_navbar.scss
+++ b/assets/scss/components/_navbar.scss
@@ -1,0 +1,141 @@
+@mixin _labelLarge() {
+	font-size: $font-size-small;
+	max-width: 20vw; // allow more space for labels @ medium and up
+}
+
+#globalNav {
+	@include standardBorder('bottom');
+}
+
+.header {
+	justify-content: space-between;
+
+	@include atMediaUp(medium) {
+		justify-content: space-around;
+	}
+}
+
+$zindex--aboveFloatingContent: map-get($zindex-map, 'floating-content') + 1;
+.header--aboveFloatingContent {
+	position: relative;
+	z-index: $zindex--aboveFloatingContent+2;
+}
+
+.header-navItem {
+	@include color-svg($C_textSecondary);
+	color: $C_textPrimary;
+	text-align: center;
+	position: relative;
+	padding-left: 0;
+}
+
+// use negative margin to pull in some
+// of the header height and center the logos
+.header-logo {
+	margin-bottom: -6px;
+}
+.header-logo--swarm {
+	margin-top: -8px;
+}
+
+@include atMediaUp(medium) {
+	.header-navItem {
+		@include customPropertyValue(padding-left, calc(var(--responsiveSpace)), $space);
+	}
+}
+
+.header-navItem--unauthenticated {
+
+	// because login/signup links do not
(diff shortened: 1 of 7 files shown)


assets/scss/components/_navbar.scss
```diff
@@ -0,0 +1,141 @@
...
+}
+
+@include atMediaUp(medium) {
+	.header-navItem {
+		@include customPropertyValue(padding-left, calc(var(--responsiveSpace)), $space);
+	}
+}
+
+.header-navItem--unauthenticated {
+
```
acmei: Nit pick: extra line

assets/scss/components/_navbar.scss
```diff
@@ -0,0 +1,141 @@
...
+		@include _labelLarge();
+
+		&:hover,
+		&:active {
+			color: $C_accent;
+		}
+	}
+}
+
+
```
acmei: Another extra line

src/navigation/NavBar.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Icon from '../media/Icon';
+import NavItem from './NavItem';
+
+export const NavBar = () => (
+	<NavItem
+		key={0}
```
acmei: Confused about why this prop is needed here...

src/navigation/NavItem.jsx
```diff
@@ -0,0 +1,130 @@
+import PropTypes from 'prop-types';
```
acmei: Shouldn't this component have a story too??

src/navigation/NavBar.jsx
kcode20: I don't know. I'll investigate and remove it from the next PR since this is widely used in the next PR. This came with the copy and paste stuff.

src/navigation/NavItem.jsx
kcode20: No, according to how they want it built, NavItem is not going to be something that is imported into the project, strictly NavBar.

src/navigation/NavBar.jsx
kcode20: Investigated and decided to leave the key, it makes sense in the following PR. The key is because the items are part of an array.

PR #428, opened 2018-02-12
Document guidelines for adding new components

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-505

#### Description
A list of criteria that we check for anyway when reviewing PRs. This should also be added somewhere [swarm-design-system](https://meetup.github.io/swarm-design-system/)  - probably somewhere in the Components tab?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ This checklist can be used as a guide for adding components that are easy to reu
 * [ ] Doesn't rely on a parent component to behave or look correctly
 * [ ] Doesn't have logic that controls functionality outside of itself e.g.: if there is a button in the component, it should take an event handler via props instead of assuming what the button is going to do
 * [ ] Works in supported browsers and devices
-* [ ] Design is [accessible](link to a11y guides) for all users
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,30 @@ Table of Contents
...
+* [ ] Has unit tests
+* [ ] Provides logical defaults
+* [ ] If there can be errors, error states are designed and documented
+* [ ] If the component has become complex, it has been broken into smaller components
+
+**Context-agnostic**
+* [ ] Doesn't rely on a parent component to behave or look correctly
+* [ ] Doesn't have logic that controls functionality outside of itself e.g.: if there is a button in the component, it should take an event handler via props instead of assuming what the button is going to do
+* [ ] Works in supported browsers and devices
+* [ ] Design is [accessible](link to a11y guides) for all users
```
akdetrick: let's take out the broken links until we have a place to link to

akdetrick:
I agree with maybe adding this to SDS main docs instead and just linking to that page from the MWC README

README.md
mperrotti: We do have a place, I'm just a fool and forgot to add the _real_ URLs. Adding now!

PR #429, opened 2018-02-13
Wc 179/accordion height tollerance

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-179

#### Description
Enables `AccordionPanel` content area to expand to accommodate content of any height when open (as opposed to setting a hard height based on initial content height).

### DIFF
--- a/src/interactive/AccordionPanel.jsx
+++ b/src/interactive/AccordionPanel.jsx
@@ -22,14 +22,17 @@ class AccordionPanel extends React.Component {
 	}
 
 	/**
-	 * @parm {Boolean} isOpen (is panel open)
-	 * @returns {Number} panel height
+	 * @parm {Boolean} isOpen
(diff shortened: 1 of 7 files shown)


src/interactive/AccordionPanel.jsx
```diff
@@ -22,14 +22,17 @@ class AccordionPanel extends React.Component {
...
+	 * @parm {Boolean} isOpen
+	 * @parm {Ref} contentEl - element ref
+	 * @returns {Object} inline styles
 	 */
-	getHeight(isOpen) {
-		if (!this.contentEl) {
-			return '0px';
-		}
-		return `${isOpen * this.contentEl.getBoundingClientRect().height}px`;
+	getPanelStyle(isOpen, contentEl) {
```
mperrotti: This is much better than what we had before 👏 👏 👏

PR #432, opened 2018-02-14
NavBar (MP-356)

kcode20:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-356

#### Description
Since we want to unify the navigation bar across all of web, create the navigation bar in Meetup-Web Components for easy implementations in any web repo.

Part 2: Create a Navigation Bar

DoD:
* Created component and story displaying use
* Navigation can fulfill pro-web and mup-web's current design
* Navigation is customizable for future iterations and use cases

#### Comments
I added a loader for parsing SVG's so that the Meetup Logos can be used.

This majority of the code in this ticket consists of more copy and paste from mup-web, these components and their corresponding tests with small changes:
* NavBar Component:
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/HeaderContent.jsx
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/profile/profileDropdown.test.jsx
* Profile Dropdown:
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/profile/ProfileDropdown.jsx
* Notifications Dropdown:
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/notifications/NotificationsDropdown.jsx
* Notification:
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/notifications/Notification.jsx
https://github.com/meetup/mup-web/blob/master/src/components/chrome/header/notifications/notification.test.jsx

I stripped away all the link functionality and localization. That will be handled in a subsequent ticket.


package.json
```diff
@@ -49,8 +49,9 @@
   "dependencies": {
     "autosize": "3.0.21",
     "downshift": "1.26.1",
-    "react-flatpickr": "3.6.3",
+    "file-loader": "false1.1.6",
```
komlev: Not sure if this is correct version

acmei:
The arrow next to the profile photo avatar looks too close. 
screen shot 2018-02-16 at 5 20 34 pm

package.json
kcode20: I don't know why the version looked like this. Even when I tried to fix it by re-running `yarn add file-loader` , the updated versioning was `false1.1.8`. Super weird, but I just deleted the `false` and it worked fine. So we'll just go with that.

acmei:
There's a console error for Nav related to needing unique keys when rendering multiple components.

mperrotti:
This has been sitting here for a while - should we close it?

kcode20:
No. It is still pending. I am awaiting approval.

acmei:
Padding in the notifications dropdown seems off.

screen shot 2018-03-08 at 11 39 33 am

acmei:
The unautheticated version seems to be stretching beyond the width of the screen for some reason.
screen shot 2018-03-08 at 11 40 45 am

acmei:
Should the avatar be flush against the edge of the screen?
screen shot 2018-03-08 at 11 41 25 am

acmei:
There's a console error 
```
Warning: React does not recognize the `htmlElement` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `htmlelement` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in button
    in Button (created by ActionItem)
    in ActionItem (created by NavItem)
```

kcode20:
The console error is coming from the Button component. I will tell System-Design about the error, but I don't think I should fix that within this ticket. It might have side effects that will affect buttons across the site.

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+				className: CLASS_UNAUTH_ITEM,
+			},
+		];
+
+		let authItems = [
+			self.isProMember && {
+				shrink: true,
+				linkTo: proDashboard.link,
+				label: proDashboard.label,
+				className: cx(
```
emilyeng: you don't need to use `cx` for this case since you don't have any conditional logic. you can just pass the string

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+				linkTo: groups.link,
+				label: groups.label,
+				className: `atMedium_display--none ${CLASS_AUTH_ITEM}`,
+				icon: <Icon shape="groups" size="s" className="display--block" />,
+			},
+			{
+				shrink: true,
+				linkTo: messages.link,
+				label: messages.label,
+				className: cx('navItem--messages', CLASS_AUTH_ITEM),
```
emilyeng: also here - you dont _need_ to use `cx`, could just pass a string

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+						shape="messages"
+						size="s"
+						className="display--block atMedium_display--none"
+					/>
+				),
+				hasUpdates: messages.unreadMessages > 0,
+			},
+			{
+				shrink: true,
+				linkTo: media.isAtMediumUp ? false : notifications.link,
```
emilyeng: here and below: should this technically be an empty string (or null) instead of false? to keep the type the same for this prop

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+				className: cx('navItem--notifications', CLASS_AUTH_ITEM),
+				icon: (
+					<Icon
+						shape="notifications"
+						size="s"
+						className="display--block atMedium_display--none"
+					/>
+				),
+				onClickAction:
+					media.isAtMediumUp && notifications.list.length == 0
```
emilyeng: `===`
also do you need to null check for `list` existing on `notifications`? or is it guaranteed in this context?

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+							<Icon
+								shape="chevron-down"
+								size="xxs"
+								className="padding--left-half"
+							/>
+						</FlexItem>
+					</Flex>
+				),
+				onClickAction:
+					media.isAtMediumUp && groups.list.length == 0
```
emilyeng: `===`

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,284 @@
...
+				dropdownContent: media.isAtMediumUp && profileContent,
+			},
+		];
+
+		unauthItems = unauthItems.map((item, i) => {
+			return (
+				<NavItem
+					key={i}
+					shrink={item.shrink}
+					linkTo={item.linkTo && item.linkTo}
```
emilyeng: here and below: i think this can jsut be `linkTo={item.linkTo}`

src/navigation/Nav.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Nav', () => {
+	const navSwarm = renderComponent({
+		media: { isAtMediumUp: true },
+		self: { status: 'active', name: 'Jeff Cleft' },
+	});
+	const navScript = renderComponent({
+		media: { isAtLargeUp: true },
+		self: { status: 'prereg' },
+	});
```
emilyeng: i feel like there are other cases besides the 2 above that should be unit tested?

src/navigation/components/NavItem.jsx
```diff
@@ -71,9 +70,7 @@ export const NavItem = props => {
 			{label && <span className={classNames.label}>{label}</span>}
 			{hasUpdates && (
 				<span className="counterBadge">
-					<span className="visibility--a11yHide">
-						Updates
-					</span>
+					<span className="visibility--a11yHide">Updates</span>
```
emilyeng: do we want this english hard coded here?

src/navigation/components/notifications/Notifications.jsx
```diff
@@ -0,0 +1,152 @@
...
+					<FlexItem className="valignChildren--center">
+						<h4
+							className="text--small"
+							dangerouslySetInnerHTML={dangerouslySetInnerHTML}
+						/>
+						<p className="text--hint text--small">
+							<Icon shape={getIconShape(kind)} size="xs" />
+							<time className="notification-time text--small">
+								{timeSince.toDateString()}
+								{/* TODO: Format Time*/}
```
emilyeng: is this todo still relevant?

src/navigation/components/notifications/NotificationsDropdown.jsx
```diff
@@ -0,0 +1,65 @@
...
+	notificationTitleLabel,
+	generateClassicUrl,
+}) => {
+	const isEmptyState = notifications.length < 1;
+
+	const NotificationsList = () => (
+		<div className="notifications--populated">
+			<ul className="list">
+				{notifications.map((notif, key) => (
+					// TODO: refactor
```
emilyeng: also this todo?

src/navigation/components/notifications/NotificationsDropdown.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+import {
+	MOCK_NOTIFICATION_COMMENT,
+	MOCK_NOTIFICATIONS_LIST,
+} from 'meetup-web-mocks/lib/notifications/api';
+
+import { NotificationsDropdownComponent } from './NotificationsDropdown';
+
+// add a member comment notifs without a photo
+// to the mock notifications list
```
emilyeng: this comment can prob be removed?

src/navigation/components/notifications/Notifications.jsx
kcode20: Yes, I removed the time formatting because mwc does not have translations, this has to be addressed. What are your thoughts on how this should be done?

src/navigation/components/notifications/NotificationsDropdown.jsx
kcode20: No. Will remove.

acmei:
The pro authenticated mobile view doesn't seem to have a link to the pro dashboard.

src/navigation/Nav.jsx
```diff
@@ -0,0 +1,292 @@
...
+						key={i}
+						shrink={item.shrink && item.shrink}
+						linkTo={item.linkTo}
+						label={item.label && item.label}
+						labelClassName={item.labelClassName && item.labelClassName}
+						className={item.className && item.className}
+						icon={item.icon && item.icon}
+						onClickAction={item.onClickAction && item.onClickAction}
+						dropdownContent={item.dropdownContent && item.dropdownContent}
+						hasUpdates={item.hasUpdates && item.hasUpdates}
```
emilyeng: i think i commented this before, but do you really need to check for `item.x && item.x`? i think if you just pass `item.x` as the prop and it happens to be null, it's the same thing?

src/navigation/components/notifications/Notifications.jsx
emilyeng: hmmm i wonder if instead of `updated` being a prop passed in that's a timestamp, maybe you jsut have a prop like `formattedTimeSince` for the formatted date string?

src/navigation/nav.story.jsx
```diff
@@ -0,0 +1,91 @@
...
+		label: 'Signup',
+	},
+	proDashboard: {
+		link: 'meetup.com/pro',
+		label: 'Pro Dashboard',
+		mobileLabel: 'Dashboard',
+		proLogo: MOCK_MEMBER.photo.photo_link,
+	},
+	explore: {
+		link: 'meetup.com/explore',
```
emilyeng: nitpick - i know this is for the storybook story, but is `/explore` the correct url for the Explore button? on the current nav, we direct to `/find/events`

src/navigation/components/notifications/Notifications.jsx
kcode20: I was thinking about that, but updated comes from every individual notification and the mapping of these notifications come from notificationsDropdown. I guess I can map over these notifications before I pass them to the meetup-web-component and add the `formattedTimeSince` prop that I need. Let me know if you think that's messy.

src/navigation/components/notifications/Notifications.jsx
```diff
@@ -76,7 +76,7 @@ class Notification extends React.PureComponent {
 		const {
 			memberId,
 			kind,
-			updated,
+			formattedTimeSince,
```
emilyeng: jsut remember to update the `propTypes` with the new ones!

PR #433, opened 2018-02-15
remove 0 height for open accordion panels

akdetrick:
#### Related issues
Follows up on: https://meetup.atlassian.net/browse/WC-179

#### Description
Allows `AccordionPanel` content to expand in height beyond the `min-height` originally set in `render()`.


mperrotti:
Merge it, and we'll follow up with a fix for the animation

PR #434, opened 2018-02-16
Added defaultProps to Footer component

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-184

#### Description
Added default props to `<Footer>` to make sure it can be rendered without providing props

### DIFF
--- a/src/Footer.jsx
+++ b/src/Footer.jsx
@@ -205,7 +205,7 @@ Footer.defaultProps = {
 		{
 			header: 'Your Account',
 			items: [
-				<a href='/register'>Settings</a>,
+				<a href="/register">Settings</a>,
 				<a href="/logout">Log out</a>,
 				<a href="/help">Help</a>
 			]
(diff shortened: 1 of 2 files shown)


src/Footer.jsx
```diff
@@ -196,6 +196,49 @@ export const Footer = ({
...
+	createMeetup: {
+		text: 'Create a Meetup',
+		link: '/create/'
+	},
+	linkSets: [
+		{
+			header: 'Your Account',
+			items: [
+				<a href='/register'>Settings</a>,
+				<a href="/logout">Log out</a>,
```
emilyeng: nitpick - some of these `href`s here and below are using `""` and some are using `''`

mperrotti: Woops. I'll clean up...

mperrotti:
No, @emilyeng, we don't want to handle translations in MWC. Sadly, we'll have to manually copy over the translations from `mup-web`

PR #435, opened 2018-02-21
Add AppBadges component from mup-web

mperrotti:
Does anybody know what's going on there? Seems like a Webpack issue

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-186

#### Description
Add `AppBadges` component

screen shot 2018-02-20 at 7 48 20 pm


src/media/AppBadges.jsx
```diff
@@ -0,0 +1,52 @@
...
+export const ANDROID_DOWNLOAD_LINK = 'http://meetu.ps/2ZhSyb';
+const getAppStoreDownloadPhoto = (platform, language) =>
+	// eslint-disable-next-line global-require, import/no-dynamic-require
+	require(`../../assets/images/app_download/${platform}/download_${language}.png`);
+
+/**
+ * Renders iOS and Android app badges
+ * @class AppBadges
+ */
+const AppBadges = ({ className, locale, getAppStorePhoto, ...other }) => {
```
emilyeng: now that i think of it, it might be worth changing the prop name from `locale` to `language` to be extra clear that it's looking for somehting like `fr` instead of `fr-FR`

PR #443, opened 2018-02-23
Reduce flatpickr CSS bloat

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-180

#### Description
Instead of importing all the Flatpickr CSS, I refactored the styles and kept them in `_calendar.scss`


akdetrick:
LGTM, but found on issue: when selecting from a "valid date range", the select input with the year date is looking a little weird...
screen shot 2018-02-26 at 1 39 22 pm

PR #444, opened 2018-02-23
Optionally render button with anchor or Link tag

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-174

#### Description
Let buttons render as `<a>` or `<Link>`, and tell them where to go with a `link` prop

### DIFF
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -28,10 +28,7 @@ class Button extends React.PureComponent {
 			small,
 			bordered,
 			hasHoverShadow,
-			// htmlElement,
-			// linkComponent,
-			// linkTo,
-			wrapper,
+			wrapperEl,
 			...other
 		} = this.props;
 
@@ -75,7 +72,7 @@ class Button extends React.PureComponent {
 
 		return (
 			React.cloneElement(
-				wrapper,
+				wrapperEl,
 				{
 					className: classNames.button,
 					onClick: onClick,
@@ -89,7 +86,7 @@ class Button extends React.PureComponent {
 
 Button.defaultProps = {
 	htmlElement: 'button',
-	wrapper: <button />
+	wrapperEl: <button />
 };
 
 Button.propTypes = {
@@ -99,9 +96,6 @@ Button.propTypes = {
 	small: PropTypes.bool,
 	icon: PropTypes.any,
 	right: PropTypes.bool,
-	htmlElement: PropTypes.oneOf(['button', 'a']),
-	linkComponent: PropTypes.element,
-	wrapper: PropTypes.element,
-	linkTo: PropTypes.string
+	wrapperEl: PropTypes.element,
 };
 export default Button;
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -28,8 +28,10 @@ class Button extends React.PureComponent {
 			small,
 			bordered,
 			hasHoverShadow,
-			buttonTag,
-			linkTo,
+			// htmlElement,
+			// linkComponent,
+			// linkTo,
+			wrapper,
 			...other
 		} = this.props;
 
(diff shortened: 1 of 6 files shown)


src/interactive/__snapshots__/accordionPanel.test.jsx.snap
```diff
@@ -314,6 +314,7 @@ exports[`AccordionPanel Panel with toggle switch exists and renders a toggle swi
                   <Button
                     aria-checked={false}
                     aria-labelledby="firstsection-label"
+                    buttonTag="button"
```
akdetrick: I'd prefer we use something that we can standardize across components that accept html element customization...

`tag`?
`DOMElement`?
`htmlElement`?

src/forms/button.test.jsx
```diff
@@ -121,4 +122,24 @@ describe('Button', () => {
...
+		);
+
+		it('should render element from buttonTag prop', () => {
+			expect(buttonTagComponent.find('a').length).toBe(1);
+		});
+
+		it('should render `link` prop', () => {
+			expect(buttonTagComponent.find('a').prop('link')).toBe(link);
+		});
+
```
akdetrick: might be good to add a test checking that `href` and `to` props are set correctly given the tag type

src/interactive/__snapshots__/accordionPanel.test.jsx.snap
akdetrick: I'd vote for `htmlElement`, just because `htmlFor` is a thing

mperrotti: But `htmlFor` is just a way to get around `for` being a reserved string in JavaScript, it's not describing the `for` in any way. If you don't think this makes it confusing to name this `htmlElement`, I'll continue with that.
Otherwise, how about `domElement`?

src/forms/Button.jsx
```diff
@@ -69,25 +71,38 @@ class Button extends React.PureComponent {
...
+
 Button.propTypes = {
 	reset: PropTypes.bool,
 	fullWidth: PropTypes.bool,
 	primary: PropTypes.bool,
 	small: PropTypes.bool,
 	icon: PropTypes.any,
 	right: PropTypes.bool,
+	buttonTag: PropTypes.oneOf(['button', 'a', 'Link']),
+	linkTo: PropTypes.string
```
akdetrick: we should probably do a custom prop validator here to make `linkTo` required if `buttonTag` is 'Link'

src/forms/Button.jsx
```diff
@@ -69,25 +71,38 @@ class Button extends React.PureComponent {
...
+};
+
 Button.propTypes = {
 	reset: PropTypes.bool,
 	fullWidth: PropTypes.bool,
 	primary: PropTypes.bool,
 	small: PropTypes.bool,
 	icon: PropTypes.any,
 	right: PropTypes.bool,
+	buttonTag: PropTypes.oneOf(['button', 'a', 'Link']),
```
akdetrick: `Link` is not like the other two options here, `button`, and `a`. The button and anchor strings are lowercase and will be treated by React as the DOM elements they are.

`Link` refers to an actual component, which means you'd either have to:

1. always import `Link` from `react-router` and keep the version of `react-router` in this repo in sync with consumer apps
2. split out `Link` as a separate prop, that accepts `PropTypes.element`.

I'm recommending option (2) here because it removes a dependency on `react-router`, and it would also enable you to remove the `linkTo` prop, as the `Link` component would be passed in whole as a prop:

```
<Button linkComponent={<Link linkTo="lol">My cool link</Link>} />
```

In the end, you would have a prop to accommodate a link component, and a prop that allows users to specify an html element to render. In the event a user passes both, we could favor rendering `Link`.

```
Button.propTypes = {
	...
	htmlElement: PropTypes.oneOf(['button', 'a']),
	linkComponent: PropTypes.element,
	...
};
```

mperrotti: You're 100% right, and I'm going with option 2

src/forms/button.test.jsx
```diff
@@ -121,4 +118,33 @@ describe('Button', () => {
 			});
 		});
 	});
+
+	describe('buttonTag', () =>{
```
akdetrick: nitpick - should be `describe('wrapperEl`

src/forms/Button.jsx
```diff
@@ -72,24 +73,31 @@ class Button extends React.PureComponent {
...
+				},
+				icon ? iconChildren : children
+			)
 		);
 	}
 }
 
+Button.defaultProps = {
+	htmlElement: 'button',
+	wrapperEl: <button />
```
mmcgahan: Might be worth a followup ticket to clean this up, but `cloneElement` should only be used as a last-ditch option - it's pretty expensive and can oftentimes be avoided.

In this case, the prop should be a ReactComponent, not a ReactElement, and the `defaultProp` should be `wrapperEl: 'button'`

Then, in `render`, you would do something like

```jsx
const Component = this.props.wrapperEl;

return (
  <Component className={classNames.button} onClick={onClick} {...other}>
    {icon ? iconChildren : children }
  </Component>
);
```

akdetrick: Filed a follow up as WC-199

PR #446, opened 2018-02-26
Required form field design update

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SST-10

#### Description
Replace asterisk with text that says "required". Will follow-up in `swarm-sasstools` to remove old styles with this PR: https://github.com/meetup/swarm-sasstools/pull/80

screen shot 2018-02-26 at 10 46 13 am

### DIFF
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -40,10 +40,11 @@ export class CalendarComponent extends React.Component {
 			label,
 			helperText,
 			error,
-			required,
 			datepickerOptions,
 			className,
 			onChange, // eslint-disable-line no-unused-vars
+			required,
+			requiredText,
 			...other
 		} = this.props;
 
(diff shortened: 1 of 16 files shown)


src/forms/CalendarComponent.jsx
```diff
@@ -86,7 +87,7 @@ export class CalendarComponent extends React.Component {
 		return (
 			<span>
 				{label && (
-					<label htmlFor={id || name} className={classNames.label}>
+					<label htmlFor={id || name} className={classNames.label} data-requiredtext={requiredText && `(${requiredText})`}>
```
akdetrick: I'd remove the parens here; requiring the parens in the translation string will provide context to the translators and avoid any weird situations where parens may not make sense in a certain language.

src/forms/CalendarComponent.jsx
```diff
@@ -123,6 +123,11 @@ CalendarComponent.propTypes = {
 		PropTypes.element
 	]),
 	onChange: PropTypes.func, // provided by `redux-form`
+	required: PropTypes.bool,
+	requiredText: (props) => (
```
akdetrick: This update will cause existing usage of these form components to fail propTypes in consumer apps. This may actually be a good thing if it only causes problems in dev, but a greater concern is that there is no fallback to using a `*`. 

I'd recommend either building in a fallback for the case of required fields without a `requiredText` prop, or filing a follow up bug to make usage updates in `mup-web` and `pro-web` with the upgrade of `meetup-web-components`.

PR #447, opened 2018-02-27
Fix non-multiselectable AccordionPanel bug

mperrotti:
Re Prelim - thinking on whether this is the best solution

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-193

#### Description
Open `<AccordionPanel>`s collapse their height when AccordionPanelGroup is `multiselectable`. Moved setState stuff from `this._handleToggle` to `componentWillReceiveProps`

### DIFF
--- a/src/interactive/AccordionPanel.jsx
+++ b/src/interactive/AccordionPanel.jsx
@@ -63,20 +63,6 @@ class AccordionPanel extends React.Component {
 			onClickCallback
 		} = this.props;
 
-		// this.setState(()=>
-		// 	this.getPanelStyle(isOpen, this.contentEl),
-		// 	() => {
-		// 		// console.log('first setState being hit');
-		// 		setTimeout(() => {
-		// 			this.setState(()=>
-		// 				this.getPanelStyle(!isOpen, this.contentEl),
-		// 				() => {
-		// 					// console.log('second setState being hit');
-		// 				}
-		// 			);
-		// 		}, 1);
-		// 	}
-		// );
 		setClickedPanel && setClickedPanel(this.props.clickId, !isOpen, this.getPanelStyle(!isOpen, this.contentEl)); // this.getPanelStyle(!isOpen, this.contentEl)
 		onClickCallback && onClickCallback(e, !isOpen);
 	}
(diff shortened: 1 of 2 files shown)


src/interactive/AccordionPanel.jsx
```diff
@@ -34,6 +34,9 @@ class AccordionPanel extends React.Component {
 	getPanelStyle(isOpen, contentEl) {
 		const style = { height: '0px' };
 
+		// console.log(`isOpen: ${isOpen}`);
```
akdetrick: i recommend deleting these instead of commenting them out

mperrotti: woops, missed that one

src/interactive/AccordionPanel.jsx
```diff
@@ -60,17 +63,7 @@ class AccordionPanel extends React.Component {
...
-			() => {
-				setTimeout(() => {
-					this.setState(()=>
-						this.getPanelStyle(!isOpen, this.contentEl)
-					);
-				}, 1);
-			}
-		);
-		setClickedPanel && setClickedPanel(this.props.clickId, !isOpen);
+		setClickedPanel && setClickedPanel(this.props.clickId, !isOpen, this.getPanelStyle(!isOpen, this.contentEl)); // this.getPanelStyle(!isOpen, this.contentEl)
```
akdetrick: let's remove this comment too

src/interactive/AccordionPanel.jsx
```diff
@@ -88,6 +81,25 @@ class AccordionPanel extends React.Component {
...
 
+	/**
+	 * @description set height state when the panel receives a change
+	 * @returns {undefined}
+	 */
+	componentWillReceiveProps(nextProps) {
+		if (nextProps.isOpen !== this.props.isOpen) {
+			this.setState(()=>
+				this.getPanelStyle(!nextProps.isOpen, this.contentEl),
+				() => {
```
akdetrick: this extra anonymous function may not be necessary; I believe you should be able to directly call the `setTimeout` in the `setState` callback

mperrotti: I removed it and it broke. Keeping it in

PR #449, opened 2018-02-28
DO NOT MERGE

akdetrick:
Don't do it.

Placeholder for 5.0.X release with "future css" configuration.


akdetrick:
This will not merge to master; we're going to do this in a non-breaking way.

PR #450, opened 2018-02-28
Wc 166/configure postcss

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-166

#### Description
- adds webpack configuration for PostCSS and CSS modules
- moves styling for `bounds` into a css module as example
- fix issues with custom properties & color functions
- fix issues with custom properties & sass calculation
- removes `customPropertyValue` usage
- removes `--responsiveSpace` custom property usage in favor of `--responsive-space` from constants


assets/scss/components/_forms.scss
```diff
@@ -30,6 +32,21 @@ input:focus + .icon--field {
 	font-weight: normal;
 }
 
+.label--required:after {
+	content: attr(data-requiredtext);
+	color: $C_textSecondary;
+	// font-size: $font-size-tiny;
```
mperrotti: probably not part of this PR, but let's remove this

assets/scss/components/_forms.scss
```diff
@@ -30,6 +32,21 @@ input:focus + .icon--field {
 	font-weight: normal;
 }
 
+.label--required:after {
```
mperrotti: Why aren't we replacing SASS vars with custom properties in here?

akdetrick: rather than replacing all Sass vars, I remapped them in sasstools v4:
https://github.com/meetup/swarm-sasstools/compare/master...release/4.0.0

All existing Sass vars now map to custom properties, which allows us to launch faster and clean up in steps afterward.

mperrotti: I completely forgot about that 🤦‍♂️

akdetrick:
Adding "prelim" until we investigate a plan to launch to `mup-web` without major breaking changes. We may need to undo some things in this branch before merging.

akdetrick:
Not merging to master; this should happen in a non-breaking way.

PR #454, opened 2018-03-05
Allow for setting custom validation error message on TextInput

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-548

#### Description
Allow engineers to customize the text in the browser's native validation error message

Before (screenshot taken in IE11):
screen shot 2018-02-26 at 2 10 23 pm

After (screenshot taken in Chrome):
screen shot 2018-03-05 at 5 26 46 pm

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -31,6 +31,8 @@ export const TextInput = (props) => {
 		helperText,
 		required,
 		requiredText,
+		isValid,
+		validityMessage,
 		...other
 	} = props;
 
@@ -69,6 +71,17 @@ export const TextInput = (props) => {
 	const inputStyles = iconShape && {
 		paddingLeft: `${paddingSize}px`
 	};
+	const customValidityMessage = isValid ? '' : validityMessage;
+
+	let textInput = null;
+
+	const handleOnChange = (e) => {
+		if (onChange) {
+			onChange(e);
+		}
+
+		textInput && textInput.setCustomValidity(customValidityMessage);
+	};
 
 	// WC-158
 	// Only add a `value` prop if it is defined.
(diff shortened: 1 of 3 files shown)


src/forms/TextInput.jsx
```diff
@@ -69,6 +71,17 @@ export const TextInput = (props) => {
 	const inputStyles = iconShape && {
 		paddingLeft: `${paddingSize}px`
 	};
+	const customValidityMessage = isValid ? '' : validityMessage;
+
+	let textInput = null;
```
akdetrick: why are we setting this to `null` here?

src/forms/TextInput.jsx
```diff
@@ -69,6 +71,17 @@ export const TextInput = (props) => {
...
+
+	const handleOnChange = (e) => {
+		if (onChange) {
+			onChange(e);
+		}
+
+		textInput && textInput.setCustomValidity(customValidityMessage);
```
akdetrick: why are we checking for `textInput` before setting it's validity message?

src/forms/TextInput.jsx
mperrotti: Because I was getting a lint message when I just tried to set it to `let textInput`

src/forms/TextInput.jsx
mperrotti: Tests were failing because of setting the textInput to null

src/forms/TextInput.jsx
akdetrick: You _should_ be able to just declare the variable, `let textInput;`. I'm not seeing any lint errors when I try it locally. This is just a style nitpick though; not a blocker.

mperrotti: Updated :)

PR #462, opened 2018-03-08
enable soft limit for char counter fields

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-196

#### Description
Enables soft character limit for char counter fields by avoiding passing the `maxLength` prop as an HTML attribute, which triggers most browsers to give the field a hard limit.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -77,6 +77,10 @@ export const TextInput = (props) => {
 		other.value = value;
 	}
 
+	// Character limits should be a "soft" limit.
+	// Avoid passing maxLength as an HTML attribute
+	if (maxLength) delete other.maxLength;
+
 	return (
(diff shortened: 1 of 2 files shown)


src/forms/TextInput.jsx
```diff
@@ -77,6 +77,10 @@ export const TextInput = (props) => {
 		other.value = value;
 	}
 
+	// Character limits should be a "soft" limit.
+	// Avoid passing maxLength as an HTML attribute
+	if (maxLength) delete other.maxLength;
```
eilinora: do we want this to be a universal thing? i would assume making it a soft limit would be an option

akdetrick: We do in fact want to make this universal. There are better reasons to use a soft limit than a hard limit, and it's nice to codify this product decision in our components.

PR #464, opened 2018-03-09
fall back to '*' for required field requiredText

akdetrick:
Avoids breaking change with `requiredText` by falling back on "*" as a default required message.


akdetrick:
`MG-441` should take care of the cleanup in `mup-web`. Once we do this cleanup in both `mup-web` and `pro-web`, we can remove this fallback and make `requiredText` a required prop

PR #468, opened 2018-03-14
Don't let disabled buttons call onClick

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-565


src/forms/Button.jsx
```diff
@@ -78,7 +78,7 @@ class Button extends React.PureComponent {
 		return (
 			<Component
 				className={classNames.button}
-				onClick={onClick}
+				onClick={!disabled && onClick}
```
seanlipkin: 👍

PR #469, opened 2018-03-14
Add Typeahead component

mperrotti:
#### Related issues
https://meetup.atlassian.net/browse/MG-495
Fixes https://meetup.atlassian.net/browse/WC-157

#### Description
Adding a typeahead component for engineers to use on features like venue searching. This component leans hard on PayPal's [Downshift](https://github.com/paypal/downshift) component. This component is also used for our `<Dropdown>`

screen shot 2018-03-14 at 1 05 46 pm


assets/scss/components/_typeahead.scss
```diff
@@ -0,0 +1,26 @@
+.typeahead {
+	position: relative;
+
+	.typeahead-input {
```
akdetrick: does this need to be nested to override some other input class?

src/forms/Typeahead.jsx
```diff
@@ -0,0 +1,84 @@
...
+				(<div className="typeahead">
+					<TextInput
+						className="typeahead-input"
+						{...getInputProps({
+							...inputProps
+						})}
+					/>
+					{isOpen
+						?
+							<div
```
akdetrick: Are there any props we need to pass to the dropdown element in order for `Downshift` to apply the correct aria attributes?

src/forms/Typeahead.jsx
```diff
@@ -0,0 +1,84 @@
...
+												}
+											)}
+										>
+											{item.props.children}
+										</div>
+									))
+								}
+							</div>
+						:
+							null
```
akdetrick: Is there a reason to use the tenary here when the second expression is always `null`?
(as opposed to `{isOpen && <dropdownStuff>}`

assets/scss/components/_typeahead.scss
mperrotti: Sadly it needs to be nested to override the bottom margin. `flush--bottom` would also kill the bottom padding, so that was a no-go

src/forms/Typeahead.jsx
mperrotti: `Downshift` adds its own aria attributes, but if you see any missing in the rendered HTML, let me know and I'll add them

src/forms/Typeahead.jsx
mperrotti: When I first tried that I was getting issues, but I can give it another try

PR #470, opened 2018-03-14
Signup functionality (MP-587)

kcode20:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-587

#### Description
Currently, the MWC navigation component doesn't handle sign-up functionality.

DOD:

- Add Functionality to Sign-up

Do tiny fixes:
- Fix See All Groups Link
- Fix Logo Link
- Fix Profile Link
- Flow Error for the Notification Dropdown

screen shot 2018-03-14 at 5 40 42 pm

### DIFF
--- a/assets/scss/components/_signup.scss
+++ b/assets/scss/components/_signup.scss
@@ -1,4 +1,7 @@
 .meetup-signupModal {
+	.padding--left-half {
+		padding-left: $space-half;
+	}
 	.meetup-signupModal-wrapper {
 		margin: $space $space-2;
 	}
--- a/assets/scss/components/_signup.scss
+++ b/assets/scss/components/_signup.scss
@@ -8,7 +8,4 @@
 			background-color: $C_facebook;
 		}
 	}
-	.meetup-signUpModal-or {
-		text-transform: uppercase;
-	}
 }
--- a/assets/scss/components/_signup.scss
+++ b/assets/scss/components/_signup.scss
@@ -1,0 +1,14 @@
+.meetup-signUpModal {
+	.meetup-signUpModal-wrapper {
+		margin: $space $space-2;
+	}
+	.meetup-signUpModal-facebook {
+		background-color: $C_facebook;
+		&:hover {
+			background-color: $C_facebook;
+		}
+	}
+	.meetup-signUpModal-or {
+		text-transform: uppercase;
+	}
+}
--- a/assets/scss/components/_signup.scss
+++ b/assets/scss/components/_signup.scss
@@ -1,8 +1,8 @@
-.meetup-signUpModal {
-	.meetup-signUpModal-wrapper {
+.meetup-signupModal {
+	.meetup-signupModal-wrapper {
 		margin: $space $space-2;
 	}
-	.meetup-signUpModal-facebook {
+	.meetup-signupModal-facebook {
 		background-color: $C_facebook;
 		&:hover {
 			background-color: $C_facebook;
--- a/assets/scss/components/_signup.scss
+++ b/assets/scss/components/_signup.scss
@@ -8,4 +8,7 @@
 			background-color: $C_facebook;
 		}
 	}
+	.meetup-signupModal-or {
+		text-transform: uppercase;
+	}
 }
--- a/src/SignupModal.jsx
+++ b/src/SignupModal.jsx
@@ -15,11 +15,11 @@ import { C_MEDIUMGRAY } from 'swarm-constants/dist/js/colorConstants';
 
 /* consts */
 export const SIGNUP_MODAL_CLASS = 'meetup-signUpModal';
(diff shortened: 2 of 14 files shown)


assets/scss/components/_signup.scss
```diff
@@ -0,0 +1,14 @@
...
+		margin: $space $space-2;
+	}
+	.meetup-signUpModal-facebook {
+		background-color: $C_facebook;
+		&:hover {
+			background-color: $C_facebook;
+		}
+	}
+	.meetup-signUpModal-or {
+		text-transform: uppercase;
```
emilyeng: can you use the class `.text--uppercase` instead?

src/SignupModal.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+import { C_MEDIUMGRAY } from 'swarm-constants/dist/js/colorConstants';
+
+/* consts */
+export const SIGNUP_MODAL_CLASS = 'meetup-signUpModal';
+export const SIGNUP_MODAL_WRAPPER_CLASS = 'meetup-signUpModal-wrapper';
+export const SIGNUP_MODAL_FACEBOOK_CLASS = 'meetup-signUpModal-facebook';
+export const SIGNUP_MODAL_GOOGLE_CLASS = 'meetup-signUpModal-google';
+export const SIGNUP_MODAL_OR_CLASS = 'meetup-signUpModal-or';
+export const SIGNUP_MODAL_EMAIL_CLASS = 'meetup-signUpModal-email';
```
emilyeng: nitpick - you can use template strings for the child class names like
```
export const SIGNUP_MODAL_WRAPPER_CLASS = `${SIGNUP_MODAL_CLASS}-wrapper`;
export const SIGNUP_MODAL_FACEBOOK_CLASS = `${SIGNUP_MODAL_CLASS}-facebook`;
...
```

src/SignupModal.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @param {Object} props component properties
+ * @returns {React.element} SignupModal
+ */
+export const SignupModal = function SignupModal({ onDismiss, signupOptions }) {
```
emilyeng: could be simplified as
```
export const SignupModal = ({ onDismiss, signupOptions }) => {
```

src/SignupModal.jsx
```diff
@@ -0,0 +1,119 @@
...
+						)}
+					>
+						<Flex>
+							<FlexItem shrink className="inverted">
+								<Icon
+									shape="external-facebook"
+									size="xs"
+									className="align--left"
+								/>
+							</FlexItem>{' '}
```
emilyeng: hmm what is this `{' '}` used for? can this be achieved with padding or margin?

src/SignupModal.test.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { SignupModal } from './SignupModal';
+import { signupOptions } from './signupModal.story';
+
+describe('SignupModal', () => {
+	const MOCK_PROPS = {
+		signupOptions: signupOptions,
```
emilyeng: can just pass `signupOptions`, since the key and value match
```
const MOCK_PROPS = {
    signupOptions,
    ...
};
```

src/navigation/Nav.jsx
```diff
@@ -29,20 +30,37 @@ export const DropdownLoader = ({ label }) => (
...
 export class Nav extends React.Component {
+	/**
+	 * @constructor
+	 * @param {Object} props properties passed into component
+	 */
+	constructor(props) {
+		super(props);
+		this.onDismissSignUpModal = this.onDismissSignUpModal.bind(this);
+		this.onClickSignUpAction = this.onClickSignUpAction.bind(this);
+		this.state = { isSignUpModalOpen: false };
```
emilyeng: nitpick - i'm seeing `SignUpModal` used in a lot of these functions or props, but we call the component itself `SignupModal`. maybe worth it to choose one and stick with it

assets/scss/components/_signup.scss
kcode20: For some reason the "text--uppercase" class isn't working...

emilyeng:
just a couple minor things
- looks like there could be a little space in-between "Already a member?" and "Login"
- the "OR" looks misaligned with the text of the buttons but maybe it's fine

screen_shot_2018-03-14_at_6_48_22_pm

otherwise LGTM

kcode20:
The `OR` is in the center of the buttons and the text in the buttons is centered in it's flex within the button. Trying to align those is a little difficult. I tried to make it a little less noticeable by flushing all padding on the left for the flex that contains the button text. But it didn't make much of a difference.
screen shot 2018-03-14 at 7 18 03 pm

emilyeng:
actually one comment... i think the spelling of "Signup" in both the modal and the nav should be two words, right? 

"Signup" --> "Sign up"

PR #472, opened 2018-03-14
pass error in textarea redux form wrapper only when field has been touched.

akdetrick:
Fixes an issue on the schedule form where some textareas were showing validation error states before touched.


six5532one:
lgtm once checks pass

PR #473, opened 2018-03-16
Visual improvements to :focus for AccordionPanel

mperrotti:
Re Prelim tag:
* [x] Review design with Todd
* [x] On AccordionPanels with ToggleSwitch, should we _only_ keyboard navigate and show `:focus` on the ToggleSwitch? I think so, but I want a second opinion

#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-230

#### Description
A less severe focus state for `AccordionPanel` that only shows for users navigating with a keyboard

Interacting using a keyboard:
accordionfocus_keyboard

Interacting using a cursor:
accordionfocus_mouse

### DIFF
--- a/src/interactive/AccordionPanel.jsx
+++ b/src/interactive/AccordionPanel.jsx
@@ -23,6 +23,7 @@ class AccordionPanel extends React.Component {
 
 		this.state = {
 			height: props.isOpen ? 'auto' : '0px'
+			// hasFocus: false
 		};
 	}
 
@@ -111,6 +112,9 @@ class AccordionPanel extends React.Component {
 			indicatorIcon;
 	}
 
+	/**
+	 * @returns undefined
+	 */
 	onTransitionEnd() {
 		if (this.props.isOpen) {
 			this.setState(()=>({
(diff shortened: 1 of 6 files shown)


src/interactive/AccordionPanel.jsx
```diff
@@ -23,6 +23,7 @@ class AccordionPanel extends React.Component {
 
 		this.state = {
 			height: props.isOpen ? 'auto' : '0px'
+			// hasFocus: false
```
akdetrick: let's kill this state property instead of commenting it out

src/interactive/AccordionPanel.jsx
```diff
@@ -119,6 +123,20 @@ class AccordionPanel extends React.Component {
 		}
 	}
 
+	// /**
```
akdetrick: Favor removing this over commenting it out

src/interactive/AccordionPanel.jsx
mperrotti: lolwoops

PR #474, opened 2018-03-16
TogglePill updates

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-200

#### Description
Nicole came up with a new design for our `TogglePill` component when her team was working on the post-event feedback flow. I did a minor version bump because the styles Yanyi wrote to support this component in `mup-web` need to be removed in order for us to have the expected result in the post-event feedback flow.

screen shot 2018-03-16 at 3 43 38 pm

screen shot 2018-03-16 at 4 08 23 pm


Makefile
```diff
@@ -1,5 +1,5 @@
 CI_BUILD_NUMBER ?= $(USER)-snapshot
-VERSION ?= 4.6.$(CI_BUILD_NUMBER)
```
akdetrick: We should add a release note for `4.7` briefly explaining the change

PR #475, opened 2018-03-16
use meta.touched in all redux-form wrappers that handle errors

akdetrick:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-204

#### Description
For all `redux-form` wrappers that handle errors, check `meta.touched`. This ensures validation on the field will not show an error until a user has interacted with it unless the `validateBeforeTouched` prop is passed.

### DIFF
--- a/src/forms/redux-form/CalendarComponent.jsx
+++ b/src/forms/redux-form/CalendarComponent.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import CalendarComponent from '../CalendarComponent';
 
 /**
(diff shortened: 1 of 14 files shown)


src/forms/redux-form/CalendarComponent.jsx
```diff
@@ -8,8 +8,10 @@ import CalendarComponent from '../CalendarComponent';
  * @return {Component} TextInput
  */
 const ReduxFormCalendarComponent = props => {
-	const { meta, input, ...other } = props;
-	return <CalendarComponent error={meta.error} {...input} {...other} />;
+	const { meta, input, validateBeforeTouched, ...other } = props;
+	const error = (validateBeforeTouched || meta.touched) ? meta.error : null;
```
mmcgahan: looks like `meta` is now a required prop? should get a proptype if so, and I suspect it will break some existing implementations?

akdetrick:
This addresses the issue of showing a field in an error state before users have interacted with it. The expected behavior being that a user doesn't see a form with errors before they have interacted with it.

The `meta` prop is always passed via `redux-form`, but it wouldn't hurt to add the appropriate proptypes...

akdetrick:
Looks like `validateBeforeTouched` is no longer being used in the consumer apps. The expected behavior is:

Errors should not display unless the user has interacted with the field, or form submission fails (due to invalid fields). 

The changes in this branch are to prevent errors from displaying when forms first load. This is an issue that can happen if validation happens too early in app code. While that could be considered a defect in the app feature, it's nice handle `redux-form` errors in the standard components in a way that ensures the user will only see the errors when they are most relevant (editing fields and failed submits).

akdetrick:
Tested in `mup-web`. Seems to be working as expected; errors only appear on submit or user interaction.

akdetrick:
Nothing in consumer apps appear to be using the `validateBeforeTouched` prop. We can add this back in later if need be.

PR #479, opened 2018-03-21
MWC Nav Update (MP-642)

kcode20:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-642

#### Description
Catch-all for problems identified in the meetup-web-component from design updates and PR comments:

DoD:
fix loading state condition for dropdown
fix overflow of text for profile dropdown
add default logo for navigation

#### Comments
Prelim until tested in pro-web


kcode20:
Is there a way to render the component to snapshot it. I don't think the tests that I created are descriptive enough. It creates a snapshot of the props passed to the component instead of what the whole component actually entails.

src/navigation/Nav.jsx
```diff
@@ -76,40 +77,49 @@ export class Nav extends React.Component {
 		} = navItems;
 		const isLoggedOut = self.status === 'prereg' || !self.name;
 		const classNames = cx('padding--all', className);
+		const proLogo = ((proDashboard.mainAccount || {}).group_photo || {}).thumb_link;
```
komlev: I know we try not to use lodash that
But I guess this case could be way easier if we can just use lodash/get
It'll be like

get(proDashboard, 'mainAccount.group_photo.thumb_link', '')
Rather than using this || {} syntax

komlev:
You can try using Enzyme's not "shallow" but "mount" to have rendered snapshot

src/navigation/Nav.jsx
mmcgahan: We've talked through this before - `_.get` is cool, but it's designed to handle a lot of edge cases that don't apply here https://github.com/lodash/lodash/blob/4.4.2-npm-packages/lodash.get/index.js. It uses a couple dozen functions and a handful of regexes to be highly-performant and hide the complexity of those operations behind a clever interface - it's easy to use but the implementation is much harder to reason through than the `|| {}` syntax here, so I recommend avoiding it. If you would like to make this easier to read, I recommend writing a selected that takes an 'account' object and returns the `thumb_link` and name it `getLogo`, but even that might be unnecessary indirection.

PR #482, opened 2018-03-22
Console error :skull: so close

eilinora:
added bool to 3 more components

### DIFF
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -5,7 +5,6 @@ import withErrorList from '../utils/components/withErrorList';
 
 import Flatpickr from 'react-flatpickr';
 
-
 /**
  * @module CalendarComponent
  * @description Single date picker component.
@@ -13,7 +12,7 @@ import Flatpickr from 'react-flatpickr';
  *
  * For full documentation of available `datePickerOptions`, see:
  * https://chmln.github.io/flatpickr/options/
-*/
+ */
 export class CalendarComponent extends React.Component {
 	constructor(props) {
 		super(props);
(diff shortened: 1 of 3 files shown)


src/forms/CalendarComponent.jsx
```diff
@@ -114,20 +108,11 @@ CalendarComponent.propTypes = {
...
 	datepickerOptions: PropTypes.object,
-	error: PropTypes.oneOfType([
-		PropTypes.string,
-		PropTypes.element
-	]),
-	helperText: PropTypes.oneOfType([
-		PropTypes.string,
-		PropTypes.element
-	]),
+	error: PropTypes.oneOfType([PropTypes.string, PropTypes.element, PropTypes.bool]),
```
eilinora: only real change

src/forms/ErrorList.jsx
```diff
@@ -12,29 +12,22 @@ const ERROR_CLASSNAME = 'text--error text--small';
...
-			{error && <li className={cx(ERROR_CLASSNAME, 'text--small')}>{error}</li> }
+		<ul aria-live="assertive" {...other}>
+			{error && <li className={cx(ERROR_CLASSNAME, 'text--small')}>{error}</li>}
 		</ul>
 	);
 };
 ErrorList.propTypes = {
 	errorId: PropTypes.string,
-	error: PropTypes.string,
+	error: PropTypes.oneOfType([PropTypes.string, PropTypes.element, PropTypes.bool]),
```
eilinora: only real change

src/forms/TimeInput.jsx
```diff
@@ -402,7 +402,7 @@ TimeInput.defaultProps = {
 
 TimeInput.propTypes = {
 	name: PropTypes.string.isRequired,
-	error: PropTypes.string,
+	error: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
eilinora: only real change

src/forms/CalendarComponent.jsx
mperrotti: What's the use case for passing a boolean to `error`? I thought we used this prop for rendering an error message

eilinora: The schedule form is passing false when there is no error to show

PR #483, opened 2018-03-22
adding controlled prop to be able to have controlled/uncontrolled com…

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MUP-19251

#### Description
Add ability to have uncontrolled checkbox, so it does not stores checked value in state

### DIFF
--- a/src/forms/Checkbox.jsx
+++ b/src/forms/Checkbox.jsx
@@ -15,17 +15,25 @@ export const DISABLED_CHECKBOX_CLASS = 'disabled';
 class Checkbox extends React.PureComponent {
 	constructor(props) {
 		super(props);
-		this.state = {
-			checked: props.checked || false,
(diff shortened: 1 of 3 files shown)


src/forms/Checkbox.jsx
```diff
@@ -15,17 +15,25 @@ export const DISABLED_CHECKBOX_CLASS = 'disabled';
...
 	constructor(props) {
 		super(props);
-		this.state = {
-			checked: props.checked || false,
+		const state = {
+			focused: false,
 		};
+
+		if (props.controlled) {
+			state.checked = props.checked || false;
```
DA-Denis: How about to reduce some code:
const state = {
   focused: false,
   checked: props.controlled ? props.checked : false
};

PR #485, opened 2018-03-26
Make entire dropdown menu item area clickable

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-568

#### Description
The padding on `dropdownMenu-item` was causing it's children not to be clickable unless you clicked directly on the child element. It's even harder to click on the child element when it's an inline element.

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -290,7 +290,7 @@ Dropdown.defaultProps = {
 Dropdown.propTypes = {
 	trigger: PropTypes.element.isRequired,
 	content: PropTypes.element,
-	menuItems: PropTypes.array,
+	menuItems: PropTypes.arrayOf(PropTypes.element),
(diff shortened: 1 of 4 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -10,18 +10,9 @@ import {
...
-			backgroundColor: isActive && C_COOLGRAYLIGHTTRANSP
-		}}
-	>
-		{children}
-	</div>
-);
+const ConditionalWrap = ({condition, wrap, children}) => condition ? wrap(children) : children;
 
 /**
  * @module Dropdown
```
akdetrick: Do we need to change the propType for `items` to expect an array of elements?

mperrotti: Aha, yes.

akdetrick:
It would be good to double check usage to make sure nobody is passing a string to `items`

PR #488, opened 2018-03-29
Manually update Typeahead's highlighted index

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-208

#### Description
For some reason, the Typeahead isn't reliably changing it's internal state when a user hovers a new item in the dropdown. I think this might be an issue with [Downshift](https://github.com/paypal/downshift), but I can't identify what it is in order to submit a PR to them.
This is a bit "brute force", but I'd love to hear

### DIFF
--- a/src/forms/Typeahead.jsx
+++ b/src/forms/Typeahead.jsx
@@ -15,39 +15,41 @@ class Typeahead extends React.PureComponent {
 	constructor(props) {
 		super(props);
 		this.onItemMouseEnter = this.onItemMouseEnter.bind(this);
-		this.onItemKeyUp = this.onItemKeyUp.bind(this);
+		this.stateReducer = this.stateReducer.bind(this);
 
 		this.state = {
-			actualHighlightedIndex: 0
+			highlightedIndex: -1
 		};
 	}
 
 	onItemMouseEnter (i) {
-		this.setState({actualHighlightedIndex: i});
+		this.setState({highlightedIndex: i});
 	}
 
-	onItemKeyUp(e) {
-		switch (e.key) {
-			case 'ArrowDown':
+	stateReducer(state, changes) {
+		switch (changes.type) {
+			case Downshift.stateChangeTypes.keyDownArrowDown:
 				this.setState({
-					actualHighlightedIndex:
-						parseInt(this.state.actualHighlightedIndex + 1) >= this.props.items.length
+					highlightedIndex:
+						parseInt(this.state.highlightedIndex + 1) >= this.props.items.length
 							?
 								0
 							:
-								parseInt(this.state.actualHighlightedIndex+1)
+								parseInt(this.state.highlightedIndex+1)
 				});
-				break;
-			case 'ArrowUp':
+				return changes;
+			case Downshift.stateChangeTypes.keyDownArrowUp:
 				this.setState({
-					actualHighlightedIndex:
-						this.state.actualHighlightedIndex < 1
+					highlightedIndex:
+						this.state.highlightedIndex < 1
 							?
-								0
+								parseInt(this.props.items.length-1)
 							:
-								parseInt(this.state.actualHighlightedIndex-1)
+								parseInt(this.state.highlightedIndex-1)
 				});
-				break;
+				return changes;
+			default:
+				return changes;
 		}
 	}
 
@@ -60,7 +62,10 @@ class Typeahead extends React.PureComponent {
 		} = this.props;
 
 		return (
-			<Downshift {...other}>
+			<Downshift
+				stateReducer={this.stateReducer}
+				{...other}
+			>
 				{({
 					getInputProps,
 					getItemProps,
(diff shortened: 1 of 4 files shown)


src/forms/Typeahead.jsx
```diff
@@ -12,6 +12,44 @@ export const TA_ITEM_CLASSNAME = 'typeahead-item';
...
+	onItemMouseEnter (i) {
+		this.setState({actualHighlightedIndex: i});
+	}
+
+	onItemKeyUp(e) {
+		switch (e.key) {
+			case 'ArrowDown':
+				this.setState({
+					actualHighlightedIndex:
+						parseInt(this.state.actualHighlightedIndex + 1) >= this.props.items.length
```
akdetrick: will `length` be off by one here , or is `actualHighlightedIndex` zero-indexed?

src/forms/Typeahead.jsx
```diff
@@ -12,6 +12,44 @@ export const TA_ITEM_CLASSNAME = 'typeahead-item';
  * @module Typeahead
  */
 class Typeahead extends React.PureComponent {
+	constructor(props) {
+		super(props);
+		this.onItemMouseEnter = this.onItemMouseEnter.bind(this);
+		this.onItemKeyUp = this.onItemKeyUp.bind(this);
+
+		this.state = {
+			actualHighlightedIndex: 0
```
akdetrick: no harm in just calling this `highlightedIndex`, since this internal state is now the point of truth for which item is highlighted

src/forms/Typeahead.jsx
```diff
@@ -15,39 +15,41 @@ class Typeahead extends React.PureComponent {
...
+		this.setState({highlightedIndex: i});
 	}
 
-	onItemKeyUp(e) {
-		switch (e.key) {
-			case 'ArrowDown':
+	stateReducer(state, changes) {
+		switch (changes.type) {
+			case Downshift.stateChangeTypes.keyDownArrowDown:
 				this.setState({
```
akdetrick: The `changes` argument here refers to the parts of state that will be updated in `Downshift` state change. For example....

Let's say this is the initial `state` of `Downshift`:
```
state = {
	selectedIndex: 0,
	highlightedIndex: 0,
	foo: 'bar',
};
```

Then let's say that `Downshfit` is changing state for `highlightedIndex` only, based on some user action. This is what `stateReducer` would be called with:

```
stateReducer = (state, changes) => {
	console.dir(state); // full state object as described above
	console.dir(changes); // only the parts of state that change, { highlightedIndex: <updated value> }

	// stateReducer returns the parts of state to update
	// to use as an argument for a setState in Downshift
	return changes;
}
```

Downshift is essentially doing this in calls to its `setState`:

`setState(this.props.stateReducer(this.state, { <state changes> }))`

----

Rather than fixing this issue only in `Typeahead` internal state, `stateReducer` provides an opportunity to intercept `Downshift` state changes to amend the values as needed:

```
return { ...changes, { highlightedIndex: <our calculatedIndex> } }
```

This will ensure that any management of state in `Typeahead` is in sync with `Downshift` state.

akdetrick: This may also solve any weirdness between `highlightedIndex` and `selectedIndex` in Downshift

mperrotti: I originally did it this way, but then I _always_ need to use `highlightedIndex` to check if we should be adding the className `typeahead-item--isActive`.
This is a problem because we can't rely on `highlightedIndex` to be correct when navigating with a cursor. In the venue search, Typeahead doesn't even do an internal state change when using the cursor on some of the dropdown items, so I can't even bring that interaction into `stateReducer` as well

akdetrick: So `Typeahead` can keep the internal state you're using for `highlightedIndex`, but we should also try passing that state property to `Downshift` in the return value to correct its internal state of highlighted item

mperrotti: Ahhh, I understand. Ok, I'll do that 👍

src/forms/Typeahead.jsx
```diff
@@ -12,6 +12,56 @@ export const TA_ITEM_CLASSNAME = 'typeahead-item';
...
+				highlightedIndexState =
+					parseInt(this.state.highlightedIndex + 1) >= this.props.items.length
+						?
+							0
+						:
+							parseInt(this.state.highlightedIndex+1);
+				this.setState({
+					highlightedIndex: highlightedIndexState
+				});
+				return {
```
akdetrick: Because we're returning basically the same thing in each `case`, I'd recommend doing the return outside of the switch statement. Each `case` should end with `break;` instead. The `default` case can be removed for this statement.

src/forms/Typeahead.jsx
```diff
@@ -12,6 +12,54 @@ export const TA_ITEM_CLASSNAME = 'typeahead-item';
...
+					});
+				break;
+			case Downshift.stateChangeTypes.keyDownArrowUp:
+				highlightedIndexState =
+					this.state.highlightedIndex < 1
+						?
+							parseInt(this.props.items.length-1)
+						:
+							parseInt(this.state.highlightedIndex-1);
+					this.setState({
```
akdetrick: The internal `setState` can also happen outside the switch statement. The `switch` would then only have the responsibility of updating the value of `highlightedIndexState`

mperrotti: I tried that, it doesn't work for some reason

PR #490, opened 2018-03-29
Make the browser-native clear button work on TimeInput

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-209

#### Description
Firefox has no good way of hiding the native clear button in `input[type="time"]`, so I decided to just keep it around and make it clear the input.
However, Malik made a change to `return` if the input value was empty on change. Instead of throwing out the change entirely, I set it to `00:00`.

### DIFF
--- a/src/forms/TimeInput.jsx
+++ b/src/forms/TimeInput.jsx
@@ -107,13 +107,14 @@ export class TimeInput extends React.PureComponent {
 	 * 	handler prop, if there is one provided (eg supplied by redux-form or DateTimePicker)
 	 */
 	onTimeInputChange(e) {
-		const { value } = e.target;
-		const isValid = /^([0-1]?[1-9]|2[0-4]):([0-5][0-9])(:[0-5][0-9])?$/.test(value);
+		let { value } = e.target;
 
 		// Time inputs are allowed to return an empty string for invalid input.
 		// We should ignore all invalid input and return immediately
 		// Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Validation
-		if (!isValid) return;
+		if (value === '') {
+			value = '00:00';
+		}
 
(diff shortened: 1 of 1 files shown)


src/forms/TimeInput.jsx
```diff
@@ -107,11 +107,14 @@ export class TimeInput extends React.PureComponent {
...
 	onTimeInputChange(e) {
-		const { value } = e.target;
+		let { value } = e.target;
+
 		// Time inputs are allowed to return an empty string for invalid input.
 		// We should ignore all invalid input and return immediately
 		// Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Validation
-		if (value === '') return;
+		if (value === '') {
+			value = '00:00';
```
mmcgahan: this is a valid value that will basically never be what the user intended the value to be, which doesn't seem like a good user experience. It _sort of_ aligns with what the 'clear' button would intuitively do in firefox, but not really, and it definitely doesn't align with what you would expect to have saved when you accidentally enter `27:00`, so I think it would be best to try a different solution

mperrotti: I can't figure out a way around this. We either:
* Let the browser-native clear button do nothing (only affects Firefox)
* Make the clear button reset the time to `12:00am`

I think making the clear button reset to `12:00am` feels way less strange than having it do nothing at all.

FWIW: a user can't type `27:00` into a time input, it will just end up being `02:07 [AM|PM]` because of the way the browser handles time inputs

mmcgahan: ok - I don't remember the steps to reproduce the original issue, but there was definitely something that could be typed that made the `value` empty. I guess just go with this solution for now and hopefully it doesn't put people into a confusing state.

PR #492, opened 2018-04-03
Pass localeCode to getSocialLinks and update snaps

jmhammons:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-215

#### Description

Noticing that the social icons in the site footer don't link to language-specific social sites anymore.

Steps to reproduce:

1. log out
2. go to the footer on meetup.com
3. pick french language
4. click on Twitter icon

It should go to https://twitter.com/MeetupFR/ but is instead going to https://twitter.com/Meetup/. Some other languages also have their own language-specific fb and twitter pages.

### DIFF
--- a/src/footer.test.jsx
+++ b/src/footer.test.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
 
+import config from 'mwp-config';
+
 import {
 	Footer,
 	FooterCategory,
@@ -9,7 +11,6 @@ import {
 } from './Footer';
 
 const DEFAULT_LANG = 'en-US';
-const FRENCH_LANG = 'fr-FR';
 const onLanguageSelect = jest.fn();
 
 const footerLinkSets = [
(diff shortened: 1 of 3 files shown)


src/footer.test.jsx
```diff
@@ -116,9 +115,28 @@ describe('LanguageSelectInput', () => {
...
+		'pl-PL',
+		'pt-BR',
+		'sv-SE',
+		'tr-TR',
+		'th-TH',
+		'ja-JP',
+		'ko-KR',
+	];
+	allLocales.forEach(localeCode => {
+		it(`renders for ${localeCode} locale`, () => {
```
jmhammons: Not sure if this is a good way to do this or not. Open to suggestions.

src/footer.test.jsx
```diff
@@ -116,9 +115,28 @@ describe('LanguageSelectInput', () => {
...
+		'it-IT',
+		'nl-NL',
+		'pl-PL',
+		'pt-BR',
+		'sv-SE',
+		'tr-TR',
+		'th-TH',
+		'ja-JP',
+		'ko-KR',
+	];
```
mmcgahan: you can import this like

```js
import config from 'mwp-config';
const allLocales = config.locales;
```

src/footer.test.jsx
mmcgahan: LGTM

src/footer.test.jsx
jmhammons: Oh snap. Yay.

PR #493, opened 2018-04-04
Footer: adding legalCopy prop for GDPR

jmhammons:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-216

#### Description

Adds a `legalCopy` prop to the `Footer` comp for GDPR compliance.

See ticket, but here is the idea...

gdpr-footer-copy

### DIFF
--- a/src/Footer.jsx
+++ b/src/Footer.jsx
@@ -20,8 +20,6 @@ import withMatchMedia from './utils/components/withMatchMedia';
 
 import { getSocialLinks } from './utils/getSocialLinks';
 
-export const defaultLegalCopy = 'Meetup is a wholly owned subsidiary of WeWork Companies, Inc.';
-
 export const FooterCategory = (props) => (
 	<Chunk>
 		<h4 className="text--bold margin--bottom">{props.header}</h4>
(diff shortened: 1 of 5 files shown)


src/Footer.jsx
```diff
@@ -187,6 +190,13 @@ export const Footer = ({
 							separator="·"
```
jmhammons: Would be nice if I can add the separator at medium, probably.

src/Footer.jsx
```diff
@@ -20,6 +20,8 @@ import withMatchMedia from './utils/components/withMatchMedia';
 
 import { getSocialLinks } from './utils/getSocialLinks';
 
+export const defaultLegalCopy = 'Meetup is a wholly owned subsidiary of WeWork Companies, Inc.';
```
jmhammons: Is this the right way to go in this repo -- english defaults here and 'apps' pass in translated strings?

mperrotti: This is fine, but I'd recommend just defining this in the story and passing it in from there.

jmhammons: Hey @mperrotti -- just getting back today. So is the recommendation to define the english string in the story and the component separately? Is it okay that the story and component might get out of sync copy-wise. I'm fine to change it if that's how it's done here, just trying to understand the reasoning.

mperrotti: I'm suggesting not to define anything in the component. Just put the `legalCopy` prop in place, and then pass `legalCopy="Meetup is a wholly...`

src/Footer.jsx
```diff
@@ -183,10 +184,17 @@ export const Footer = ({
 				<Bounds>
 					<div className="padding--all">
 						<InlineBlockList
-							className="text--small align--center atMedium_align--left"
+							className="text--small align--center atMedium_align--left footerList-legal"
 							separator="·"
 							items={[
 								<span>{`© Meetup ${new Date().getFullYear()}`}</span>,
+								(legalCopy && <span>{legalCopy}</span>),
+							].filter(item => item)}
```
jmhammons: I _think_ I need this, so we don't have an unwanted `<li/>` if a consumer app doesn't pass `legalCopy` in, but will double-check.

PR #494, opened 2018-04-05
Improve focus handling in <Modal />

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-217

#### Description
* Ensure that focus when tabbing while a modal is open is restricted to focusable content in the modal
* Tab order wraps within the modal (the first focusable element will be focused after the last one has been reached)
* You can specify what the first focusable element should be. By default, it will be the "close" button

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "autosize": "3.0.21",
     "downshift": "1.31.2",
     "file-loader": "1.1.8",
+    "focus-trap-react": "false3.1.2",
     "lint-staged": "7.0.0",
     "raf-schd": "2.1.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "autosize": "3.0.21",
     "downshift": "1.31.2",
     "file-loader": "1.1.8",
+    "focus-trap-react": "false3.1.2",
```
akdetrick: not sure how this happens; we should remove `false` from this version number.

mperrotti: This happens when I do `yarn add package-name`, and it's very annoying. Removing..

PR #498, opened 2018-04-09
Better handling of text wrapping on Dropdown menuItems

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-214

#### Description
Instead of always wrapping on word breaks or using `white-space: nowrap` to make sure text in Dropdown menuItems doesn't ever wrap, have the text wrap only when the text has made use of all available space decided by `max-width`

Before (wrapping too soon):
screen shot 2018-04-09 at 3 00 18 pm

After:
screen shot 2018-04-09 at 2 56 19 pm

### DIFF
--- a/assets/scss/components/_dropdown.scss
+++ b/assets/scss/components/_dropdown.scss
@@ -115,6 +115,10 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	-webkit-transform: translateX(-50%);
 }
 
+.dropdownMenu {
+	width: max-content;
+}
+
 .dropdownMenu-item {
 	box-sizing: border-box;
 	border-bottom: 1px solid $C_border;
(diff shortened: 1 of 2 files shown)


assets/scss/components/_dropdown.scss
```diff
@@ -115,6 +115,10 @@ $spaceFromTrigger: $space-3 + $triangleSize;
 	-webkit-transform: translateX(-50%);
 }
 
+.dropdownMenu {
+	width: max-content;
```
akdetrick: Do we need to set a fallback for Edge here?

mperrotti: I can't think of a good fallback to set. Something like
```
@supports not (width: max-content) {
  .dropdownMenu {
    white-space: nowrap;
  }
}
```
Could lead to text overflowing out of the container in some cases. It's probably less weird to not have a fallback and just let the items wrap "too soon".

PR #500, opened 2018-04-11
Mp 790

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-790
#### Description
Navbar should not be in loading state for empty groups.
Currently is stays in loading state even if the groups are loaded but there are no groups in the array.
Hence the problem in production


src/navigation/Nav.jsx
```diff
@@ -86,6 +86,7 @@ export class Nav extends React.Component {
 		const showSwarmLogo = Boolean(
 			media.isAtMediumUp && !media.isAtLargeUp && !isLoggedOut
 		);
+		const isGroupsLoaded = !!groups.list;
```
acmei: For clarity would you mind using `Boolean(groups.list)`?

komlev:
@acmei done

src/navigation/nav.story.jsx
```diff
@@ -79,6 +79,11 @@ storiesOf('Nav', module)
 	.add('authenticated', () => (
 		<Nav self={MOCK_MEMBER} navItems={navItems} style={{ width: '100%' }} />
 	))
+	.add('authenticated not loaded groups', () => {
```
eilinora: nitpick
`authenticated but members groups has not been loaded`

PR #511, opened 2018-04-23
Tooltip pop up above

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-593

#### Description
Tooltips appear below their triggers by default, but sometimes we need them to appear above. This PR also contains changes Emily will need to control when the tooltip appears.

screen shot 2018-04-23 at 5 51 58 pm

### DIFF
--- a/src/interactive/Tooltip.jsx
+++ b/src/interactive/Tooltip.jsx
@@ -66,7 +66,7 @@ class Tooltip extends React.PureComponent {
 		} = this.props;
 
 		// Do not pass along to children
-		delete other.manualToggle;
+	// delete other.manualToggle;
 		delete other.isActive;
 
 		const classNames = {
@@ -179,7 +179,7 @@ Tooltip.propTypes = {
 	popAbove: PropTypes.bool,
 	className: PropTypes.string,
 	isActive: PropTypes.bool,
-	manualToggle: PropTypes.func,
+	manualToggle: PropTypes.bool,
 	maxWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 	minWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 	noPortal: PropTypes.bool,
--- a/src/interactive/Tooltip.jsx
+++ b/src/interactive/Tooltip.jsx
@@ -61,6 +61,7 @@ class Tooltip extends React.PureComponent {
 			minWidth,
 			noPortal,
 			id,
+			popAbove,
 			...other
 		} = this.props;
 
@@ -85,6 +86,10 @@ class Tooltip extends React.PureComponent {
 			return this.triggerRef;
 		};
 
+		const getContent = () => {
+			return this.contentRef;
+		};
+
 		return(
 			<div
 				className={classNames.dropdown}
@@ -104,8 +109,10 @@ class Tooltip extends React.PureComponent {
 				{ isActive &&
 					<FloatingPosition
 						getTrigger={getTrigger}
+						getContent={getContent}
 						noPortal={noPortal}
 						align={align}
+						popAbove={popAbove}
 					>
 						{({
 							top,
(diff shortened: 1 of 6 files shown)


src/interactive/Tooltip.jsx
```diff
@@ -167,9 +176,10 @@ Tooltip.propTypes = {
 	trigger: PropTypes.element.isRequired,
 	content: PropTypes.element,
 	align: PropTypes.oneOf(["left", "right", "center"]).isRequired,
+	popAbove: PropTypes.bool,
```
mperrotti: Is this prop name intuitive? Should we do something more flexible like `popupFrom=[left|right|top|bottom]`?

akdetrick: My comment above is related, but starting a new prop and treating `align` as deprecated could also be an option. Maybe `position=[left|right|top|bottom]`?

src/interactive/Tooltip.jsx
```diff
@@ -167,9 +175,10 @@ Tooltip.propTypes = {
...
 	className: PropTypes.string,
 	isActive: PropTypes.bool,
-	manualToggle: PropTypes.func,
+	manualToggle: PropTypes.bool,
```
akdetrick: This looks like this could be a breaking change. This component doesn't have much, if any usage right now, so it's probably ok as long as:
- `pro-web` isn't using this prop
- `mup-web` isn't using this prop
- other `manualToggle` props used in MWC accept `bool` (helps keep the interface predictable)

src/interactive/Tooltip.jsx
akdetrick: A separate prop allows users to control horizontal and vertical alignment separately. `popAbove` allows someone position the tooltip using `dropdown-content--above dropdown-content--right`. It depends on if we think we need that level of control, or if `[left|right|top|bottom]` is enough

src/interactive/tooltip.test.jsx
```diff
@@ -40,14 +100,33 @@ describe('Tooltip', () => {
...
+				popAbove
+				trigger={dropdownTrigger}
+				content={dropdownContent}
+			/>
+		);
+		const aboveDropdownWrapper = mount(aboveDropdown);
+		const trigger = aboveDropdownWrapper.find('.dropdown-trigger');
+
+		it('applies correct alignment className to tooltip content', () => {
+			trigger.simulate('focus');
```
akdetrick: For this, it doesn't look like you need to trigger `focus` - the markup appears to be rendered either way. If that's the case, you could just cover this (and more) with a snapshot test. The test above for `dropdown-content` would also be redundant

src/interactive/Tooltip.jsx
mperrotti: `manualToggle` is accepts `func` for `Dropdown`, and nobody was using this prop. It was accidentally copied over from `Dropdown` proptypes

akdetrick: would it be safe to change the `Dropdown` `manualToggle` propType to `bool` so that our prop signatures are consistent?

src/interactive/Tooltip.jsx
```diff
@@ -104,8 +108,10 @@ class Tooltip extends React.PureComponent {
 				{ isActive &&
 					<FloatingPosition
 						getTrigger={getTrigger}
+						getContent={getContent}
 						noPortal={noPortal}
 						align={align}
+						popAbove={popAbove}
```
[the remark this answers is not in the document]
mperrotti: They do separate things though. One is about horizontal alignment to the trigger, the other is about what direction it pops out from.

Maybe `direction=[top|bottom|left|right]`?

src/interactive/tooltip.test.jsx
mperrotti: Removing second test and adding a snapshot test 👍

PR #516, opened 2018-04-25
Replace withToggleControl usage with ToggleWrapper

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-589


src/forms/__snapshots__/toggleSwitch.test.jsx.snap
```diff
@@ -1,32 +1,13 @@
...
-  tabIndex="0"
->
-  <ToggleSwitch
-    checked={true}
-    isActive={false}
-    name="testToggle"
-    toggleActive={[Function]}
-  />
-</span>
+<ToggleWrapper
```
akdetrick: We lost some coverage here - could using `mount` instead of `shallow` help?

mperrotti: Using `mount` instead of `shallow` didn't bring coverage up

akdetrick: Let's to a follow up ticket to see if we can do anything to improve the snapshot. This otherwise looks good to go

PR #517, opened 2018-04-25
Fix TogglePill topic variant - icon shape toggle

mperrotti:
#### Related issues

#### Description
Heart icon wasn't updating because it was trying to use `isActive` prop, not the dynamic `isActive` render prop.

### DIFF
--- a/src/forms/TogglePill.jsx
+++ b/src/forms/TogglePill.jsx
@@ -91,7 +91,7 @@ class TogglePill extends React.PureComponent {
 							className={labelClassNames}
 							htmlFor={id}>
 							{children}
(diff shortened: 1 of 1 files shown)


src/forms/TogglePill.jsx
```diff
@@ -97,7 +91,7 @@ class TogglePill extends React.PureComponent {
 							className={labelClassNames}
 							htmlFor={id}>
 							{children}
-							{(topic) ? topicChildren : null}
+							{(topic) ? topicChildren(isActive) : (isActive) => null}
```
akdetrick: I don't think the function in the right hand return here is doing anything. You could just do
`{(topic) ? topicChildren(isActive) : null}`

PR #524, opened 2018-05-03
Dependency updates

sadafie:
#### Description

`mwp-cli` has been upgraded to webpack 4. In the process of updating `mup-web` to use the webpack 4 build in `mwp-cli`, certain dependencies in `meetup-web-components` would conflict with the versions being installed by `mwp-cli` and `mup-web`. These updates simply unblock mup-web.

However, I notice that this repo is using wepback 2! We should upgrade it to webpack 4 as a separate ticket.


sadafie:
SV - going to merge.

PR #528, opened 2018-05-07
Ensure Dropdowns and Tooltips are not clipped by the viewport

mperrotti:
Re Prelim label:
~Still need to add tests~

#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-225

#### Description
Added functionality to `FloatingPosition` to reposition popup content if it overflows the viewport

screen shot 2018-05-07 at 1 41 04 pm

### DIFF
--- a/src/interactive/tooltip.story.jsx
+++ b/src/interactive/tooltip.story.jsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { storiesOf } from '@storybook/react';
-import { decorateWithBasics } from "../utils/decorators";
+// import { decorateWithBasics } from "../utils/decorators";
 import Tooltip from "./Tooltip";
 import Button from "../forms/Button";
 
@@ -53,7 +53,7 @@ class ManualToggleDropdown extends React.PureComponent {
 }
 
 storiesOf("Tooltip", module)
-	.addDecorator(decorateWithBasics)
+	// .addDecorator(decorateWithBasics)
 	.addWithInfo(
 		"Basic Tooltip component",
 		"Aligned right by default",
@@ -81,14 +81,7 @@ storiesOf("Tooltip", module)
 		"Tooltip above trigger",
 		"Aligned right and appearing above the trigger",
 		() => (
-			<div
-				style={{
-					marginTop: "800px",
-					width: "500px",
-					height: "1000px",
-					marginLeft: "600px"
-				}}
-			>
+			<div style={{textAlign: 'center'}}>
 				<Tooltip
 					direction="top"
 					minWidth="0"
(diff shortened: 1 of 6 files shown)


src/utils/components/FloatingPosition.jsx
```diff
@@ -36,24 +37,46 @@ class FloatingPosition extends React.PureComponent {
...
 		const {
 			left,
 			top,
 			width,
 			height
 		} = positionData;
-
-		const scrollTop = window.scrollY || window.pageYOffset;
-		const scrollLeft = window.scrollX || window.pageXOffset;
+		const getAdjustedAlignment = alignment => {
```
akdetrick: You could define this as an exported pure function outside of the render method by passing in what the function needs instead of using closure scope to access those values.

`export const getAdjustedAlignment = (preferredAlignment, positionData, viewportWidth) => {...}`

and call it like so in the component...
`align: getAdjustedAlignment(alignment, positionData, window.innerWidth)`

and in tests...
````
const expected = 'right';
const actual = getAdjustedAlignment('left', MOCK_POSITION_DATA, MOCK_VIEWPORT_WIDTH);

expect(actual).toBe(expected);
````

File: src/utils/components/FloatingPosition.jsx [hunk 2]
```diff
@@ -36,24 +37,46 @@ class FloatingPosition extends React.PureComponent {
...
+			const overflowLeft = left + contentWidth > window.innerWidth;
+			const overflowRight = (left + width) - contentWidth < 0;
+
+			// if of viewport on the left side, go right
```
akdetrick: i think the comments here may mismatch what's being returned

src/utils/components/floatingPosition.test.jsx
```diff
@@ -39,4 +39,12 @@ describe('FloatingPosition', function() {
 	it('matches snapshot', () => {
 		expect(floatingPosition).toMatchSnapshot();
 	});
+
+	it('adjusts popup content position if it overflows the viewport', () => {
```
akdetrick: I'd recommend also adding a test to verify `preferredAlignment` is honored if the viewport does **not** clip the content width

src/interactive/tooltip.story.jsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { storiesOf } from '@storybook/react';
-import { decorateWithBasics } from "../utils/decorators";
+// import { decorateWithBasics } from "../utils/decorators";
```
akdetrick: recommend uncommenting or removing this if it's not needed

PR #531, opened 2018-05-09
Fix errors in FloatingPosition

mperrotti:
#### Related issues

#### Description
- Positioning functions actually unbind on unmount, prevent the `forceUpdate` error from being thrown
- Return if `getContent()` like we do with `getTrigger()` to prevent the error about `getClientRect` being null

### DIFF
--- a/src/utils/components/floatingPosition.test.jsx
+++ b/src/utils/components/floatingPosition.test.jsx
@@ -12,23 +12,30 @@ class TestComponent extends React.PureComponent {
 			return this.triggerRef;
 		};
 
+		const getContent = () => {
+			return this.contentRef;
+		};
+
 		return(
-			<FloatingPosition
-				getTrigger={getTrigger}
-				noPortal={false}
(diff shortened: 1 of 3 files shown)


src/utils/components/floatingPosition.test.jsx
```diff
@@ -12,23 +12,30 @@ class TestComponent extends React.PureComponent {
...
-				}) => (
-					<div>
-						<div ref={el => (this.triggerRef = el)}>Trigger</div>
-						<p>top: {top}</p>
-						<p>left: {left}</p>
-					</div>
-				)}
-			</FloatingPosition>
+			<div>
+				<div ref={el => (this.triggerRef = el)}>Trigger</div>
```
akdetrick: was this committed by mistake?

mperrotti: Nope, I needed to do this so the tests wouldn't error

PR #532, opened 2018-05-10
Typeahead multiselect

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-226

#### Description
Allows a user to select multiple items and provides a prop to open the Typeahead menu on focus

screen shot 2018-05-10 at 11 58 22 am

### DIFF
--- a/assets/scss/components/_typeahead.scss
+++ b/assets/scss/components/_typeahead.scss
@@ -31,7 +31,3 @@
 .typeahead-item--isActive {
 	background-color: $C_coolGrayLightTransp;
 }
-
-.typeahead-item--isSelected {
-	background-color: blue !important;
-}
--- a/assets/scss/components/_typeahead.scss
+++ b/assets/scss/components/_typeahead.scss
@@ -21,7 +21,17 @@
 	cursor: pointer;
 	padding: $space-half $space;
 
-	&--isActive {
-		background-color: $C_coolGrayLightTransp;
+	input[type="checkbox"],
+	input[type="radio"],
+	label {
+		pointer-events: none;
 	}
+}
+
+.typeahead-item--isActive {
+	background-color: $C_coolGrayLightTransp;
+}
+
+.typeahead-item--isSelected {
+	background-color: blue !important;
 }
--- a/src/forms/Typeahead.jsx
+++ b/src/forms/Typeahead.jsx
@@ -18,18 +18,8 @@ class Typeahead extends React.PureComponent {
 		super(props);
 
 		this.handleSelection = this.handleSelection.bind(this);
-		// this.handleSingleSelection = this.handleSingleSelection.bind(this);
-		// this.handleMultiselectSelect = this.handleMultiselectSelect.bind(this);
 		this.stateReducer = this.stateReducer.bind(this);
 	}
-
-	// handleMultiselectSelect(item) {
-	// 	this.props.onSelect(item, [...this.props.multiSelectValues, item]);
-	// }
-
-	// handleSingleSelection(selectedItem, stateAndHelpers) {
-	// 	!this.props.multiSelect && this.props.onSelect && this.props.onSelect(selectedItem, stateAndHelpers);
-	// }
 
 	handleSelection(selectedItem, stateAndHelpers) {
 		if(this.props.multiSelect) {
@@ -72,10 +62,9 @@ class Typeahead extends React.PureComponent {
 			openOnFocus,
 			openOnSelect,
 			itemToString,
+			onSelect, // eslint-disable-line no-unused-vars
 			...other
 		} = this.props;
-
-		delete other.onSelect;
 
 		return (
 			<Downshift
--- a/src/forms/Typeahead.jsx
+++ b/src/forms/Typeahead.jsx
@@ -17,28 +17,37 @@ class Typeahead extends React.PureComponent {
 	constructor(props) {
 		super(props);
 
-		this.handleSingleSelection = this.handleSingleSelection.bind(this);
-		this.handleMultiselectSelect = this.handleMultiselectSelect.bind(this);
+		this.handleSelection = this.handleSelection.bind(this);
+		// this.handleSingleSelection = this.handleSingleSelection.bind(this);
+		// this.handleMultiselectSelect = this.handleMultiselectSelect.bind(this);
 		this.stateReducer = this.stateReducer.bind(this);
 	}
 
-	handleMultiselectSelect(item) {
-		this.props.onSelect(item, [...this.props.multiSelectValues, item]);
-	}
+	// handleMultiselectSelect(item) {
+	// 	this.props.onSelect(item, [...this.props.multiSelectValues, item]);
+	// }
 
-	handleSingleSelection(selectedItem, stateAndHelpers) {
-		!this.props.multiSelect && this.props.onSelect && this.props.onSelect(selectedItem, stateAndHelpers);
+	// handleSingleSelection(selectedItem, stateAndHelpers) {
+	// 	!this.props.multiSelect && this.props.onSelect && this.props.onSelect(selectedItem, stateAndHelpers);
+	// }
+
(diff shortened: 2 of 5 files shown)


assets/scss/components/_typeahead.scss
```diff
@@ -21,7 +21,17 @@
...
+	label {
+		pointer-events: none;
 	}
 }
+
+.typeahead-item--isActive {
+	background-color: $C_coolGrayLightTransp;
+}
+
+.typeahead-item--isSelected {
```
mperrotti: Didn't actually mean to commit this

assets/scss/components/_typeahead.scss
```diff
@@ -21,7 +21,17 @@
 	cursor: pointer;
 	padding: $space-half $space;
 
-	&--isActive {
-		background-color: $C_coolGrayLightTransp;
+	input[type="checkbox"],
+	input[type="radio"],
+	label {
+		pointer-events: none;
```
mperrotti: This feels like a sloppy way to prevent default click behavior on inputs

src/forms/Typeahead.jsx
```diff
@@ -52,25 +102,34 @@ class Typeahead extends React.PureComponent {
...
+														item.props.className,
+														{
+															'typeahead-item--isActive': highlightedIndex == i,
+														}
+													),
+													...item.props
+												})}
+											>
+												{
+													typeof item.props.children === 'function'
```
mperrotti: Is this ternary weird?

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,82 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
 			...other
 		} = this.props;
 
+		delete other.onSelect;
+
 		return (
 			<Downshift
+				onSelect={this.handleSingleSelection}
+				selectedItem={multiSelectValues}
+				stateReducer={multiSelect && this.stateReducer}
```
mperrotti: This is incorrect - I only need to use `stateReducer` when we want to keep the Typeahead menu open after selection

assets/scss/components/_typeahead.scss
akdetrick: it's at least scoped to typeahead

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,81 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
+				return changes;
+		}
+	}
 
 	render() {
 		const {
 			inputProps,
 			height,
 			items,
+			multiSelect,
```
akdetrick: needs proptypes

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,81 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
+		this.handleMultiselectSelect = this.handleMultiselectSelect.bind(this);
+		this.stateReducer = this.stateReducer.bind(this);
+	}
+
+	handleMultiselectSelect(item) {
+		this.props.onSelect(item, [...this.props.multiSelectValues, item]);
+	}
+
+	handleSingleSelection(selectedItem, stateAndHelpers) {
+		!this.props.multiSelect && this.props.onSelect && this.props.onSelect(selectedItem, stateAndHelpers);
```
akdetrick: `onSelect` needs a propType

src/forms/typeahead.story.jsx
```diff
@@ -32,6 +33,74 @@ const typeaheadItems = [
...
+		super(props);
+
+		this.selectHandler = this.selectHandler.bind(this);
+
+		this.state = {
+			selectedItems: []
+		};
+	}
+
+	selectHandler(prevSelection, selectedItems) {
```
akdetrick: some jsdoc comments here would be useful

akdetrick: actually, this is in a story; not as important.

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,91 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
+		// this.handleSingleSelection = this.handleSingleSelection.bind(this);
+		// this.handleMultiselectSelect = this.handleMultiselectSelect.bind(this);
+		this.stateReducer = this.stateReducer.bind(this);
+	}
+
+	// handleMultiselectSelect(item) {
+	// 	this.props.onSelect(item, [...this.props.multiSelectValues, item]);
+	// }
+
+	// handleSingleSelection(selectedItem, stateAndHelpers) {
```
acmei: Do these lines of code still need to be here if commented out?

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,91 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
+
+	handleSelection(selectedItem, stateAndHelpers) {
+		if(this.props.multiSelect) {
+			this.props.onSelect(selectedItem, [...this.props.multiSelectValues, selectedItem]);
+		} else {
+			this.props.onSelect && this.props.onSelect(selectedItem, stateAndHelpers);
+		}
+	}
+
+	stateReducer(state, changes) {
```
acmei: This is super interesting. Is it so that a user can select using the mouse or keyboard?

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,91 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
 			items,
+			multiSelect,
+			multiSelectValues,
+			openOnFocus,
+			openOnSelect,
 			itemToString,
 			...other
 		} = this.props;
 
+		delete other.onSelect;
```
acmei: Since you're already destructuring the props above, you could do the following...
```js
const { 
   // ... all the other props before this
   itemToString,
   onSelect,
   ...other
} = this.props;
```
to remove onSelect from other.

src/forms/Typeahead.jsx
acmei: Seems ok to me

src/forms/TypeaheadItem.jsx
```diff
@@ -13,6 +13,9 @@ TypeaheadItem.propTypes = {
 		PropTypes.array,
 		PropTypes.object
 	]).isRequired,
-	className: PropTypes.string,
-	children: PropTypes.node.isRequired
+	className: PropTypes.string
+	// children: PropTypes.oneOfType([
+	// 	PropTypes.node,
+	// 	PropTypes.function
+	// ]).isRequired
```
acmei: Not sure why this is commented out. Seems like it shouldn't be?

src/forms/Typeahead.jsx
mperrotti: wooops. forgot to delete

src/forms/Typeahead.jsx
mperrotti: Users can select by mouse or keyboard by default, but this gives us more control over what gets passed to Downshift's internal state so we can do things like keep the menu open by default

src/forms/TypeaheadItem.jsx
mperrotti: Yes, I need to add it back in :x

src/forms/Typeahead.jsx
```diff
@@ -14,32 +14,80 @@ const defaultItemToString = itemValue => (typeof itemValue === 'string' ? itemVa
...
  */
 class Typeahead extends React.PureComponent {
+	constructor(props) {
+		super(props);
+
+		this.handleSelection = this.handleSelection.bind(this);
+		this.stateReducer = this.stateReducer.bind(this);
+	}
+
+	handleSelection(selectedItem, stateAndHelpers) {
```
akdetrick: still could use some jsdoc here

PR #536, opened 2018-05-14
add refCallback prop to TextInput

sadafie:
Needed for pro-web > pro-web-components > SearchBar > react-autosuggest.

SearchBar swallows up the ref.

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -12,7 +12,6 @@ export const FIELD_WITH_ICON_CLASS = 'field--withIcon';
  * @module TextInput
  */
 export const TextInput = (props) => {
-
 	const {
 		name,
 		value,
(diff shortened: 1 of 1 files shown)


src/forms/TextInput.jsx
```diff
@@ -33,6 +32,7 @@ export const TextInput = (props) => {
 		required,
 		requiredText,
 		validityMessage,
+		refCallback,
```
akdetrick: we haven't added flow typing to this repo yet, so I'd recommend adding a proptype for this prop

sadafie:
> If you think refCallback would be used often, I'd also recommend adding a story to show usage.

This underlying issue is when you nest too many components and one of the components swallows up the input ref. For example in pro-web we have:

`Redux-Form <Field>`  >  `<LocationAutoComplete>`  > `SearchBar` > `TextInput` > `<input>`

Not sure how common of a use case this would be? I'll see about adding a story.

PR #538, opened 2018-05-15
Added offset prop to FloatingPosition

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-229

#### Description
Added an `offset` prop to `FloatingPosition` to allow for fine-tuning the placement of the popup

screen shot 2018-05-15 at 4 37 28 pm

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -100,7 +100,8 @@ class Dropdown extends React.PureComponent {
 			className,
 			trigger,
 			content,
-			align, // eslint-disable-line no-unused-vars
+			align,
+			offset,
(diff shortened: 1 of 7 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -244,6 +246,10 @@ Dropdown.propTypes = {
 	content: PropTypes.element,
 	menuItems: PropTypes.arrayOf(PropTypes.element),
 	align: PropTypes.oneOf(["left", "right", "center"]).isRequired,
```
mperrotti: Maybe this should be in `FloatingPosition` instead?

PR #539, opened 2018-05-16
Dismiss Modal when clicking overlay shim

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-637

#### Description
We introduced a bug with `focus-trap` where clicks were not being handled on elements outside the `FocusTrap` component


mperrotti:
Sadaf took a quick look and said it's good to go. Merging...

PR #542, opened 2018-05-19
pass along date string and instance

mmcgahan:
The Flatpickr onChange callback is called with some additional args that are currently being swallowed - this PR makes sure they are available to any passed-in `onChange`

This is non-breaking for all current uses of CalendarComponent that only care about the first arg

### DIFF
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -28,8 +28,8 @@ export class CalendarComponent extends React.Component {
 	 * `redux-form` however, expects a single value. This function ensures that any `onChange`
 	 * prop passed to this component invokes with a single date object.
 	 */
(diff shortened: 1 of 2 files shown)


src/forms/CalendarComponent.jsx
```diff
@@ -49,13 +49,10 @@ export class CalendarComponent extends React.Component {
...
 		const classNames = {
-			label: cx(
-				{
-					'label--required': required,
-					'flush--bottom': helperText,
-				},
-				labelClassName
-			),
+			label: cx({
+				required,
```
akdetrick: The class name should be `label--required` here. There may be a `required` class in app-specific code, but it's not provided by this package.

mmcgahan: yeah, sorry, bad merge.

PR #545, opened 2018-05-23
navbar flickering on pro pages fix

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-1203
#### Description
SSR does not support withmatchMedia HOC, so first render of navbar has the wrong size state rendered.
This fix is to provide parent media object from connectWithMatchMedia which is supported on the server side.

There is a link to a thread on slack around that problem
https://meetuphq.slack.com/archives/C2GLWSMAA/p1527075866000363

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -81,7 +81,15 @@ export class Nav extends React.Component {
 	 * @return {React.element} the navbar component
 	 */
 	render() {
-		const { media, self, navItems, localeCode, className, ...other } = this.props;
+		const {
+			parentMedia,
+			media: _media,
+			self,
+			navItems,
+			localeCode,
+			className,
+			...other
+		} = this.props;
 		const {
 			login,
 			signup,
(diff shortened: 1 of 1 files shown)


acmei:
I'm a little confused about this implementation. Why not just pass a single media prop instead of both parentMedia and _media?

src/navigation/Nav.jsx
```diff
@@ -378,6 +387,7 @@ Nav.defaultProps = {
 Nav.propTypes = {
 	self: PropTypes.object,
 	media: PropTypes.object,
```
acmei: This needs to be updated to `_media`

mmcgahan:
The lack of code comments about usage is a little concerning, and the implementation adds complexity that could be avoided by using `connectWithMatchMedia` on the base _unwrapped_ `Nav` component, e.g.

```jsx
import { Nav } from 'meetup-web-components/lib/Nav';
import connectWithMatchMedia from 'mwp-app-render/lib/components/connectWithMatchMedia';

const EnhancedNav = connectWithMatchMedia(Nav);

...
<EnhancedNav ... />
```

the `renameProp` usage makes this particularly tricky, because the inconsistency between what is passed in and what the component actually has in its props makes it difficult to follow - we need to be much more consistent with leaving code comments about things that make usage a little 'weird'/clever.

My recommendation is to revert this change and wrap `Nav` with `connectMatchMedia` inside pro-web directly

PR #547, opened 2018-05-23
Accept Media (MP-1229)

kcode20:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-1229
#### Description
SSR does not support withmatchMedia HOC, so first render of navbar has the wrong size state rendered.
The initial fix was to provide parent media object from connectWithMatchMedia which is supported on the server side, but this was reverted. There is a link to a thread on slack around that problem:
https://meetuphq.slack.com/archives/C2GLWSMAA/p1527075866000363

The pro-web solution for this fix:
```jsx
import { Nav } from 'meetup-web-components/lib/Nav';
import connectWithMatchMedia from 'mwp-app-render/lib/components/connectWithMatchMedia';

const EnhancedNav = connectWithMatchMedia(Nav);

...
<EnhancedNav ... />
```
Accept the media passed by connectWithMatchMedia from pro-web in meetup-web-components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [5.0]
+
+*   **BREAKING CHANGE** `Nav` component now requires media as a prop.
+
 # [3.3]
 
-- `ReduxFormNumberInput` for use with Redux-Form
+*   `ReduxFormNumberInput` for use with Redux-Form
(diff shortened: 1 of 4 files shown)


kcode20:
@mmcgahan based on your comment on slack I am going to merge it. I made the changes you requested.

PR #548, opened 2018-05-24
Added InfoToggle component

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-639

#### Description
Added a component to support the info tooltip pattern we've been seeing a lot of lately. This will replace [InfoTooltip in mup-web](https://github.com/meetup/mup-web/blob/c3219b603abd77e7635baf521a260da793553499/src/components/InfoTooltip.jsx)

screen shot 2018-05-24 at 2 53 37 pm

### DIFF
--- a/src/interactive/__snapshots__/infoToggle.test.jsx.snap
+++ b/src/interactive/__snapshots__/infoToggle.test.jsx.snap
@@ -2,7 +2,9 @@
 
 exports[`InfoToggle should render 1`] = `
 <div>
-  <span />
+  <span>
+    Get info
+  </span>
   <Tooltip
     align="right"
     direction="bottom"
--- a/src/interactive/__snapshots__/infoToggle.test.jsx.snap
+++ b/src/interactive/__snapshots__/infoToggle.test.jsx.snap
@@ -1,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`InfoToggle should render 1`] = `
+<div>
+  <span />
+  <Tooltip
+    align="right"
+    direction="bottom"
+    id="tooltipTestId"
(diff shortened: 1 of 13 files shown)


src/interactive/__snapshots__/infoToggle.test.jsx.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`InfoToggle should render 1`] = `
+<div>
+  <span />
```
akdetrick: Looks like a label prop needs to be mocked

src/utils/components/FloatingPosition.jsx
```diff
@@ -128,6 +129,8 @@ class FloatingPosition extends React.PureComponent {
 	}
 
 	render() {
+		const throttledMaxWidth = this.props.media.isAtSmallUp ? '350px' : '200px';
```
akdetrick: "throttle" implies handling something like a stream of events... `boundedMaxWidth` may be more descriptive of what this render prop provides.

akdetrick: just `maxWidth` would also work here; it's probably not necessary to expose _how_ at works as much as what the prop value should be used for.

mperrotti: I'm going to use `boundedMaxWidth` so I can still do `maxWidth={maxWidth || boundedMaxWidth}` in `Dropdown` and `Toggle` and let people override what gets passed through `FloatingPosition`

mperrotti:
@amandakievet - fixed it in this PR: https://github.com/meetup/meetup-web-components/pull/549

PR #550, opened 2018-05-25
Mp 1249 create meetup link

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-1249

#### Description
There no "Create a Meetup" link in the navbar component, this PR adds one

virtualbox_2018-05-25_17-30-47

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -83,6 +83,7 @@ export class Nav extends React.Component {
 		const { media, self, navItems, localeCode, className, ...other } = this.props;
 		const {
 			login,
+			create,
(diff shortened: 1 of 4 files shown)


src/navigation/Nav.jsx
```diff
@@ -140,6 +141,12 @@ export class Nav extends React.Component {
 		);
 
 		let unauthItems = [
+			media.isAtMediumUp && create && {
```
acmei: Can a storybook story get added to show this in the nav?

komlev: Also I've added case where we do not have create

PR #552, opened 2018-05-30
Resolve stylelint errors

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-651

#### Description
Ran stylelint and fixed errors. I also made some small changes to our linting rules. I plan to open another PR in `meetup-web-platform` to update `stylelint-config-base` with these changes.


assets/scss/components/_icon.scss
```diff
@@ -39,7 +33,9 @@
 	// (ensures the entire element is clickable, not just the drawn bits
 	position: relative;
 
-	use { pointer-events: none; }
+	use {
```
akdetrick: I thought a single property was ok on one line?
Maybe I misunderstood the lint rule for this. Enforcing line breaking always is fine, too.

mperrotti: We configured the rule to allow this, but it looks like stylelint autofixed this anyway. I'm fine with either

PR #553, opened 2018-05-30
Removed `large` variant of TogglePill

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-661

#### Description
The design team no longer wants or needs a `large` variant of the `TogglePill`

### DIFF
--- a/src/forms/TogglePill.jsx
+++ b/src/forms/TogglePill.jsx
@@ -22,7 +22,6 @@ class TogglePill extends React.PureComponent {
 			topic,
 			id,
 			small,
-			large,
 			name,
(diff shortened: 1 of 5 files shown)


src/forms/TogglePill.jsx
```diff
@@ -22,7 +22,6 @@ class TogglePill extends React.PureComponent {
 			topic,
 			id,
 			small,
-			large,
```
akdetrick: You may want to leave this here, or `delete other.large` to pluck the prop from the spread in case any consumers still pass the `large` prop.

PR #558, opened 2018-06-01
Stylelint cleanup

mperrotti:
#### Related issues

#### Description
- Upgraded mwp-config with latest rules so we could kill stylelint overrides
- Ran stylelint with `stylelint-order` plugin
- Ran Prettier


src/forms/calendarComponent.story.jsx
```diff
@@ -28,7 +25,16 @@ storiesOf('CalendarComponent', module)
 				/>
 			</Chunk>
 			<Chunk>
-				<p className="text--caption">For a full list of available <code>datepickerOptions</code>, see <a className="link" target="_blank" href="https://chmln.github.io/flatpickr/options/">flatpickr documentation</a></p>
+				<p className="text--caption">
+					For a full list of available <code>datepickerOptions</code>, see{' '}
```
akdetrick: what on earth did prettier do here?

`see{' '}` <- looks like it's enforcing a space in a weird way to avoid a trailing space. 

Not a blocker, but would be nice to clean up.

src/interactive/AccordionPanel.jsx
```diff
@@ -71,18 +71,19 @@ class AccordionPanel extends React.Component {
...
 	/**
 	 * @description allows the AccordionPanel content to be toggled with the "Enter" and "space" keys
 	 * @returns {undefined}
 	 */
 	onKeyUp(e) {
-		const isActivatingButton = [
-			' ', /* space bar */
-			'Enter'
-		].some(key => e.key === key);
+		const isActivatingButton = [' ' /* space bar */, 'Enter'].some(
```
akdetrick: nitpick; let's just move the key comment above this const (or remove it)

```
/* {' '} = space bar */
const isActivatingButton ...
```

src/interactive/accordion.story.jsx
```diff
@@ -5,328 +5,496 @@ import { storiesOf } from '@storybook/react';
...
 						<AccordionPanel
 							isOpen
-							label='First (open by default)'
+							label="First (open by default)"
 							panelContent={
-								<div className='runningText'>
-									<p>Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard McClintock, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the undoubtable source.</p>
+								<div className="runningText">
+									<p>
+										Contrary to popular belief, Lorem Ipsum is not simply random
```
akdetrick: another nitpick - could we store this text content in a variable above the tests to make the test code easier to read/diff? This can be a follow up ticket.

src/interactive/accordionPanel.test.jsx
```diff
@@ -1,45 +1,64 @@
...
-				label='First Section'
+				label="First Section"
 				setClickedPanel={panelStateCallback}
 				onClickCallback={onClickCallback}
 				panelContent={
-					<div className='runningText'>
-						<p>Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard McClintock, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the undoubtable source.</p>
+					<div className="runningText">
+						<p>
+							Contrary to popular belief, Lorem Ipsum is not simply random text.
```
akdetrick: Same as above; it might be nice to have a separate file, like `textContentMocks`, to use for populating content in tests

src/interactive/modal.story.jsx
```diff
@@ -19,284 +19,416 @@ const onDismiss = e => {
...
-			<div style={iconSpriteStyle} dangerouslySetInnerHTML={{ __html: iconSprite }} />
-		</div>
-	)
+		'fixed',
+		'This is the basic usage with fixed position content.',
+		() => (
+			<div style={wrapperStyle}>
+				<div className="padding--all">
+					<p>
+						Throw down all the stuff in the kitchen i cry and cry and cry unless
```
akdetrick: another candidate for a separate text content mocks file

src/utils/templates/component.jsx
```diff
@@ -1,5 +1,7 @@
-export default (name) => {
-	const nameLcase = name.replace(/^[A-Z]/, (firstLetter) => firstLetter.toLowerCase());
+export default name => {
```
akdetrick: Let's file a follow up ticket to remove or update these templates

mperrotti:
I see something in the Changelog about version `5.0`, but the Makefile is still at `4.x`...should we bump a major version, or change the version number for that entry in the Changelog?

mmcgahan:
> I see something in the Changelog about version 5.0, but the Makefile is still at 4.x...should we bump a major version, or change the version number for that entry in the Changelog?

That's my bad - there were some packages in MWP that weren't explicitly supporting MWC > 4.x, so I had to bump it back down to 4.9. If you could rename `5.0` to `4.9` in the CHANGELOG, I think that's fine

src/utils/templates/component.jsx
mperrotti: Filed

PR #564, opened 2018-06-12
AccordionPanelGroup for React 17

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-241

#### Description
Replaced deprecated methods `componentWillReceiveProps` and `componentWillMount` with static `getDerivedPropsFromState` method

### DIFF
--- a/src/interactive/AccordionPanelGroup.jsx
+++ b/src/interactive/AccordionPanelGroup.jsx
@@ -5,12 +5,25 @@ import cx from 'classnames';
 export const ACCORDIONPANELGROUP_CLASS = 'accordionPanelGroup';
 
 /**
- * Determines if the passed in value is a primitive type
- * @param {any} value
- * @return {boolean} True if value is a primitive type
+ * @description takes AccordionPanelGroup and AccordionPanel data and returns an array of new panel `isOpen` states
+ * @param {Boolean} isMultiselect
+ * @param {Array} statesList
+ * @param {Object} panelData
+ * @returns {Array} newPanelStatesList - an array of each panel's `isOpen` prop value
  */
-export const isPrimitive = value =>
-	['object', 'function'].indexOf(typeof value) === -1;
+export const getNewPanelState = (isMultiselect, statesList, panelData) => {
+	const newPanelStatesList = statesList.map((panelState, i) => {
+		const defaultState = isMultiselect ? panelState : false;
+
+		// not the panel that was clicked
+		if (panelData.panelIndex !== i) {
+			return defaultState;
+		}
+
+		return panelData.isOpen;
+	});
+	return newPanelStatesList;
+};
 
 /**
  * @module AccordionPanelGroup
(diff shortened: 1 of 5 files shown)


src/interactive/AccordionPanelGroup.jsx
```diff
@@ -5,12 +5,25 @@ import cx from 'classnames';
...
- * @return {boolean} True if value is a primitive type
+ * @description takes AccordionPanelGroup and AccordionPanel data and returns an array of new panel `isOpen` states
+ * @param {Boolean} isMultiselect
+ * @param {Array} statesList
+ * @param {Object} panelData
+ * @returns {Array} newPanelStatesList - an array of each panel's `isOpen` prop value
  */
-export const isPrimitive = value =>
-	['object', 'function'].indexOf(typeof value) === -1;
+export const getNewPanelState = (isMultiselect, statesList, panelData) => {
```
akdetrick: optional code style thing: this could return immediately instead of using a function statement block...

```
export const getNewPanelState = (isMultiselect, statesList, panelData) => statesList
	.map((panelState, i) => {
		...
	});
```

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -5,12 +5,25 @@ import cx from 'classnames';
 export const ACCORDIONPANELGROUP_CLASS = 'accordionPanelGroup';
 
 /**
- * Determines if the passed in value is a primitive type
- * @param {any} value
- * @return {boolean} True if value is a primitive type
+ * @description takes AccordionPanelGroup and AccordionPanel data and returns an array of new panel `isOpen` states
+ * @param {Boolean} isMultiselect
+ * @param {Array} statesList
+ * @param {Object} panelData
```
akdetrick: what do you think of naming this `clickedPanelData` for clarity?

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -19,155 +32,70 @@ class AccordionPanelGroup extends React.Component {
...
-		const panelProps = {
-			indicatorAlign: this.props.indicatorAlign,
-			indicatorIcon: this.props.indicatorIcon,
-			indicatorIconActive: this.props.indicatorIconActive,
-			indicatorSwitch: this.props.indicatorSwitch,
-			isOpen,
-			clickId,
-			setClickedPanel: this.setPanelStates.bind(this),
+		return {
+			panelStatesList: panelPropsHaveChanged
```
akdetrick: `derivedPanelListState` is initialized as `undefined`, so you could remove the extra `panelPropsHaveChanged` flag and just check to see if the derived panel state has been set:

```
return derivedPanelListState && derivedPanelListState || prevState.panelStatesList;
```

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -205,3 +146,6 @@ AccordionPanelGroup.propTypes = {
 };
 
 export default AccordionPanelGroup;
+
+// TODO: impliment non-multiselectable
```
akdetrick: 🔪

src/interactive/accordionPanelGroup.test.jsx
```diff
@@ -58,11 +58,14 @@ describe('AccordionPanelGroup', () => {
 		/>,
 	];
 
-	describe('isPrimitve', () => {
-		it('returns true when passed a primitive value', () => {
-			const values = [1, 'string', true, () => {}, [], {}];
-			const valueTypes = values.map(x => isPrimitive(x));
-			expect(valueTypes).toEqual([true, true, true, false, false, false]);
+	describe('getNewPanelState', () => {
+		it('returns expected new panel state list when passed accordionPanelGroup data', () => {
```
akdetrick: would be nice to also have a test that asserts only the clicked panel index changes when multiselect is `false`

src/interactive/accordionPanelGroup.test.jsx
```diff
@@ -113,30 +116,28 @@ describe('AccordionPanelGroup', () => {
...
 		});
 
-		it('calls componentWillReceiveProps on AccordionPanel primitive prop changes', () => {
-			const spy = jest.spyOn(
-				AccordionPanelGroup.prototype,
-				'componentWillReceiveProps'
-			);
-
-			const component = shallow(
+		it('calls getDerivedStateFromProps on AccordionPanel prop changes', () => {
```
akdetrick: React itself has tests to ensure this lifecycle method is called as expected, and this test is indirectly testing `getNewPanelState` anyway. We could simplify this if...

The `getNewPanelState` tests have a few more assertions to check output based on different arguments, since this is what we're calling in `getDerivedStateFromProps` anyway. Those tests would cover expected state updates to the panel state list.

The tests around `getDerivedStateFromProps` here could then be simplified to just use a spy on `getNewPanelState` to verify it gets called after a `component.setProps()`. No need to assert expected state of a panel in here, as that would be covered by `getNewPanelState` tests.

src/interactive/AccordionPanelGroup.jsx
mperrotti: Into it

akdetrick:
This refactor simplifies things in AccordionPanelGroup quite a bit by removing a lot of code - seems like this should be a minor version bump with a note in the changelog.

aquillano:
@akdetrick @mperrotti I will look at this today.

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -19,155 +30,73 @@ class AccordionPanelGroup extends React.Component {
...
-		const panelProps = {
-			indicatorAlign: this.props.indicatorAlign,
-			indicatorIcon: this.props.indicatorIcon,
-			indicatorIconActive: this.props.indicatorIconActive,
-			indicatorSwitch: this.props.indicatorSwitch,
-			isOpen,
-			clickId,
-			setClickedPanel: this.setPanelStates.bind(this),
+		return {
+			// panelStatesList: panelPropsHaveChanged
```
aquillano: ✂️

aquillano: ✂️

mperrotti: Yes, I used `yarn link` to test with `mup-web` and saw no issues on the schedule form :)

mperrotti:
Yes, I used yarn link to test with mup-web and saw no issues on the schedule form :)

PR #565, opened 2018-06-12
Sticky modal close area

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-672

#### Description
Added a prop to `Modal` that allows us to make the close area of a modal stick to the top of the modal

modalstickyheader

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [4.11]
+
+- **Change** Added the ability to set the close area at the top of a Modal to be sticky using the `stickyCloseArea` prop
+
 # [4.10]
 
 - **Change** Reformatted code with Prettier and Stylelint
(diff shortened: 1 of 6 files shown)


akdetrick:
this is a new feature introduced as a non-breaking change; would be good to bump the minor version and changelog for this

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# [4.11]
+
+- **Change** Added the ability to set the close area at the top of a Modal to be sticky using the `stickyCloseArea` prop
```
akdetrick: we can be more specific here and call the change out as `**Feature**`

assets/scss/components/_modal.scss
```diff
@@ -73,6 +72,18 @@ $modal-width: 440px;
 	z-index: map-get($zindex-map, "popup") + 1;
 }
 
+.modal-closeButtonContainer--sticky {
+	background-color: $C_contentBG;
+	border-bottom: 1px solid $C_border;
+	position: sticky;
+	position: -webkit-sticky;
```
akdetrick: Should the vendor-prefixed value come _before_ the standard in this case?

mperrotti: I think both work, but putting it before feels better in this case

PR #571, opened 2018-06-18
expose ref to textarea component

browne0:
#### Description
Exposes ref to Textarea component so developers have access to the raw textarea element.

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -24,6 +24,10 @@ export class Textarea extends React.PureComponent {
 	 */
 	componentDidMount() {
 		if (this.props.autosize) {
+			if (this.props.textarearef) {
+				autosize(this.props.textarearef);
+				return;
+			}
 			autosize(this.textarea);
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/forms/Textarea.jsx
```diff
@@ -60,6 +64,10 @@ export class Textarea extends React.PureComponent {
 	 */
 	componentDidUpdate(prevProps) {
 		if (this.props.value !== prevProps.value) {
+			if (this.props.textarearef) {
+				autosize.update(this.props.textarearef);
+				return;
+			}
 			autosize.update(this.textarea);
```
mmcgahan: it doesn't look like `this.textarea` will be defined when `textarearef` is passed in as a prop - we should probably avoid calling `autosize.update` on `undefined`

browne0: It shouldn't reach that code, there's a return in the block above 🙂

mmcgahan:
Just for posterity: Malik is going to look into using a `ref` on the `<Textarea>` to access each instance's `<textarea>`, e.g. 

```jsx
class Foo extends React.Component {
  focusTextarea = () => {
    this.mwcTextarea.textarea.focus();
  };
  render() {
    return (
      <div>
        <Modal onDismiss={this.focusTextarea}
        <Textarea ref={ instance => this.mwcTextarea = instance } />
      </div>
    );
  }
}
```

PR #573, opened 2018-06-18
Fix GridList layout bug affecting responsive autoheight grid layouts

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-687

#### Description
A `max-width` was not being set on GridList items at larger viewports. This was causing unpredictable behavior when the GridList item's content caused the item to be wider

Before (6-col grid at large)
screen shot 2018-06-18 at 1 37 37 pm

After (6-col grid at large)
screen shot 2018-06-18 at 1 37 16 pm


Makefile
```diff
@@ -1,5 +1,5 @@
 CI_BUILD_NUMBER ?= $(USER)-snapshot
-VERSION ?= 4.11.$(CI_BUILD_NUMBER)
+VERSION ?= 4.12.$(CI_BUILD_NUMBER)
```
mperrotti: One could argue this is a "breaking" change because it will cause long strings of text without spaces (e.g.: the screenshots attached to this PR description) to overflow the `gridList-item`.
I don't think it's "breaking" enough to bump a major version, but I'm open to it if anybody thinks this we should be bumping a major version.

mperrotti:
This could also be achieved by dropping flexbox layout and switching to grid layout. Is that too much change for this PR?

akdetrick:
Refactoring gridList to use CSS grid layout would definitely be an option. I think you could still call that a minor release, especially if it avoids the overflow issue you mentioned above

mperrotti:
Using grid layout will let us delete some lines of CSS, but it will behave the same. The overflow issue needs to be decided on a case-by-case basis - e.g.: truncate the text, wrap the text, modify the layout to avoid narrow columns, etc...

We could handle the wrapping issue at the component level, but I'm not comfortable with making any assumptions

PR #581, opened 2018-07-03
fixing calendar component invalid date parse attempt

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-1555

#### Description
Flatpickr might call onChange with non-existent date. Seems like this happens when we provide date constraints (minDate, maxDate) without selecting any date yet.
Like in that case where we have two date fields and they affect each other.
virtualbox_2018-07-03_12-57-50

This PR fixes this problem. It does not parse if the date has not been selected yet.


mmcgahan:
Good find, thanks @komlev !

PR #585, opened 2018-07-09
Move AdminBar to MWC

lipenco:
#### Related issues
Move component https://meetup.atlassian.net/browse/MP-1599

#### Description

Move the existing component https://github.com/meetup/mup-web/blob/master/src/app/group/home/GroupHomeAdminLinks.jsx from mup-web into meetup-web-components. This is necessary so that this component can be shared between mup-web and pro-web. Additionally, add an indicator that the user is currently using production data through an 'isProd' prop passed.

DoD:

AdminBar in meetup-web-components
isProd pro available, with corresponding UI to indicate a user is currently using production data
Unit testing

screen shot 2018-07-09 at 3 54 14 pm


src/utils/components/groupHomeAdminLinks.story.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { storiesOf } from '@storybook/react';
+import { MOCK_GROUP } from 'meetup-web-mocks/lib/api';
+
+import GroupHomeAdminLinks from './GroupHomeAdminLinks';
+
+storiesOf('GroupHomeAdminLinks', module)
+	.add('default: isAdmin={false}', () => (
+		<div>
+			<GroupHomeAdminLinks isAdmin={false} />
+			<p>Component is not rendered if props: isAdmin=false</p>
```
mperrotti: I like these little notes 👍

src/utils/components/groupHomeAdminLinks.story.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { MOCK_GROUP } from 'meetup-web-mocks/lib/api';
+
+import GroupHomeAdminLinks from './GroupHomeAdminLinks';
+
+storiesOf('GroupHomeAdminLinks', module)
+	.add('default: isAdmin={false}', () => (
```
mperrotti: Add our standard decorator by putting `.addDecorator(decorateWithBasics)` before this line.
This will give you icons, and center the hint text you added. `decorateWithBasics` can be imported from `src/utils/decorators.jsx`

src/interactive/AdminBar.jsx
```diff
@@ -0,0 +1,239 @@
...
+					}`}
+					className="link"
+					target="_blank"
+					rel="noopener noreferrer"
+				>
+					Org
+				</a>
+			</li>
+		)}
+		{event !== undefined &&
```
emilyeng: i think this can be just `event &&` since `undefined` would evaluate as falsy

src/interactive/AdminBar.jsx
```diff
@@ -0,0 +1,239 @@
...
+		});
+	};
+
+	toggleHighlighter = () => {
+		this.setState(state => ({ showHighlighter: !state.showHighlighter }));
+	};
+
+	render() {
+		const { group, event, isQL, isAdmin, self, isProd, nodeEnv } = this.props;
+		const host = nodeEnv === 'production' ? 'meetup.com' : 'dev.meetup.com';
```
emilyeng: question: should this be using `nodeEnv` or `isProd` to determine the host? i don't know and am curious

lipenco: In case when we are working on dev env, but using production data it is : `isProd: true, nodeEnv: 'development', host: `dev.meetup.com'`

src/interactive/AdminBar.jsx
lipenco: Good point, I took from mup-web, but you are right that it doesn't need to be so verbose.

src/interactive/AdminBar.jsx
mmcgahan: I think that when working with prod data, you probably want to link to prod admin, so you could use `isProd` (although I would recommend renaming to `isProdApi` to remove ambiguity with `process.env.NODE_ENV`).

PR #586, opened 2018-07-10
Fix accordion race condition

browne0:
#### Description
There was a weird race condition issue that was happening from the original implementation of the AccordionGroup component. This resolves all the race condition issues except one issue that isn't currently being used in mup-web or pro-web.

Will follow up to fix.

### DIFF
--- a/src/interactive/AccordionPanelGroup.jsx
+++ b/src/interactive/AccordionPanelGroup.jsx
@@ -11,22 +11,17 @@ export const ACCORDIONPANELGROUP_CLASS = 'accordionPanelGroup';
  * @param {Object} clickedPanelData
  * @returns {Function} a Promise that maps statesList to an array of each panel's `isOpen` prop value
  */
-const getNewPanelState = (isMultiselect, statesList, clickedPanelData) => {
-	return new Promise((resolve, reject) => {
-		resolve(
-			statesList.map((panelState, i) => {
-				const defaultState = isMultiselect ? panelState : false;
+const getNewPanelState = (isMultiselect, statesList, clickedPanelData) =>
+	statesList.map((panelState, i) => {
+		const defaultState = isMultiselect ? panelState : false;
 
-				// not the panel that was clicked
-				if (clickedPanelData.panelIndex !== i) {
-					return defaultState;
-				}
+		// not the panel that was clicked
+		if (clickedPanelData.panelIndex !== i) {
+			return defaultState;
+		}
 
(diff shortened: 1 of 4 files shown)


src/interactive/AccordionPanelGroup.jsx
```diff
@@ -39,42 +39,54 @@ class AccordionPanelGroup extends React.Component {
 		this.handlePanelClick = this.handlePanelClick.bind(this);
 	}
 
-	static getDerivedStateFromProps(nextProps, prevState) {
-		let derivedPanelListState;
+	// commenting this out as this is not needed for most of the stories in storybook
```
mperrotti: Maybe instead of commenting code out, stash it or just save it to another file outside the repo?

src/interactive/accordion.story.jsx
```diff
@@ -51,13 +47,54 @@ const defaultPanels = [
...
+}
+
 storiesOf('Accordion', module)
 	.addDecorator(decorateWithBasics)
 	.addWithInfo('default', 'Basic Accordion group', () => (
 		<div className="span--100 padding--all">
 			<AccordionPanelGroup accordionPanels={defaultPanels} />
 		</div>
 	))
+	.addWithInfo('controlled accordion', 'Controlled Accordion group', () => (
```
mperrotti: I'd either stash this, or change the name of the story to something like "controlled accordion - broken right now"

src/interactive/AccordionPanelGroup.jsx
browne0: okay

src/interactive/accordion.story.jsx
browne0: okay

src/interactive/AccordionPanelGroup.jsx
```diff
@@ -9,7 +9,7 @@ export const ACCORDIONPANELGROUP_CLASS = 'accordionPanelGroup';
  * @param {Boolean} isMultiselect
  * @param {Array} statesList
  * @param {Object} clickedPanelData
- * @returns {Array} newPanelStatesList - an array of each panel's `isOpen` prop value
+ * @returns {Function} a Promise that maps statesList to an array of each panel's `isOpen` prop value
  */
 export const getNewPanelState = (isMultiselect, statesList, clickedPanelData) =>
```
mperrotti: Are we no longer using this function? If so, we should remove

PR #588, opened 2018-07-12
Mp 1587 add admin bar to pro web

lipenco:
#### Related issues
Fixes https://meetup.atlassian.net/browse/Mp-1587 

#### Description
Group should be an optional param for AdminBar component - pro-web doesn't not need to display the buttons or group related information.

AdminBar on pro-web

using prodApi:

screen shot 2018-07-12 at 3 55 41 pm

using QL and prodApi:

screen shot 2018-07-13 at 10 08 51 am

using QL and !prodAp:
screen shot 2018-07-13 at 10 08 56 am
prodApi:

### DIFF
--- a/assets/scss/components/_adminBar.scss
+++ b/assets/scss/components/_adminBar.scss
@@ -6,7 +6,12 @@
 	z-index: 9999;
 }
 
-.warning {
+.redbar {
 	background-color: var(--c-red);
+	width: 100%;
+}
+
+.greenbar {
+	background-color: var(--c-teal);
 	width: 100%;
 }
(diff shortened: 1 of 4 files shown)


eilinora:
Can we make it blink with a warning sign? 

But for real what if we made the QL on prod red and QL on dev green? People learn color patterns more than words. :)

package.json
```diff
@@ -46,11 +46,10 @@
     "react-transition-group": "^2.2.1",
     "swarm-animation": "^0.0.95",
     "swarm-icons": "^3.4.334",
-    "swarm-sasstools": "^4.2.404"
+    "swarm-sasstools": "^5.0.412 || ^4.2.408"
```
eilinora: curious why the or here?

lipenco: My bad, rebasing glitch

eilinora:
If you are on prod and not qling some does a color bar show at all? Hopefully not :)

PR #596, opened 2018-07-17
new Select component that does not track `value` in internal state

mmcgahan:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-251

#### Description

the control of `value` should be left to the parent in many cases - the new `<Select>` component is preferred over the less-flexible `<SelectInput>`. See Jira ticket for more info.


package.json
```diff
@@ -86,7 +86,7 @@
     "css-loader": "0.28.4",
     "enzyme": "3.3.0",
     "enzyme-adapter-react-16": "1.1.0",
-    "enzyme-to-json": "3.1.4",
+    "enzyme-to-json": "3.3.4",
```
mmcgahan: allows use of `React.Fragment` in tests

mperrotti:
Also, it might be nice to remove `selectInput.story.jsx`

src/forms/SelectInput.jsx
```diff
@@ -31,75 +31,23 @@ export class SelectInput extends React.PureComponent {
...
-			),
-			field: cx(
-				'select--reset span--100 padding--selectArrow',
-				{ 'field--error': (errors && errors.length > 0) || error },
-				className
-			),
-			helperText: cx('helperTextContainer', { required }),
-		};
+		// eslint-disable-next-line no-unused-vars
+		const { options, error, errors, requiredText, ...other } = this.props;
```
jaclynj: Looks like you're no longer using requiredText but this was something that was added for the schedule form, we ended up deprioritizing adding it, but the ticket I created for this still exists and I believe it's something Sylvan and the org success team want to eventually tackle.
Here's the ticket: https://meetup.atlassian.net/browse/MG-441
So in order to complete this ticket, this component will need to support the requiredText prop.

jaclynj: Update: I'm not sure the product requirement directly applies to the SelectInput/Select component but if we're changing required text from '*' to the word 'required' spelled out then it should be changed everywhere in mup-web which is what that ticket is for.

I don't necessarily agree with the product change, but this flexibility was here for a reason.

mmcgahan: Thanks for the context. I'll have to think about it a bit more, but yeah it looks like I'll have to add it back in. I _really_ think this is something that a design system should enforce consistency on rather than making it overridable, but given the interactions with TRNs, I don't see a clean and simple way of changing the copy to `(required)` universally... *le sigh*

src/forms/select.story.jsx
```diff
@@ -0,0 +1,77 @@
...
+	.add('with default selection using `value`', () => (
+		<Select
+			value="drhorse"
+			label="Select a name for your horse"
+			id="horsename"
+			name="horsename"
+		>
+			<option value="geoffrey">Geoffrey</option>
+			<option value="drhorse">Doctor Horse, MD Junior</option>
+			<option value="chompyhorse">Mister Chompy</option>
```
eilinora: :)

src/forms/SelectInput.jsx
mperrotti: Yea, we only made it "override-able" because we don't want to handle translation in our tools

eilinora: Good call Jackie! I do agree though that we should be consistent. Defect is marked as important :)

mmcgahan: I've deviated slightly from the original `requiredText` implementation and instead made `required: boolean | string` to allow people to pass in custom 'required' text through that property. The reason for doing it that way is that it avoids having props that are dependent on each other (i.e. `requiredText` assumes that `required` is `true`) - if you supply a string to `required`, then the input will be required.

The fallback to `'*'` is still included, but no more need for a `defaultProp`

PR #597, opened 2018-07-17
Flowing NumberInput + initial display fix

eilinora:
- flowing
- adding check when initial input is 0 showing nothing so placeholder text shows

### DIFF
--- a/src/forms/NumberInput.jsx
+++ b/src/forms/NumberInput.jsx
@@ -1,4 +1,4 @@
-import PropTypes from 'prop-types';
+// @flow
 import React from 'react';
 import cx from 'classnames';
 import Button from '../forms/Button';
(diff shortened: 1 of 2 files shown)


src/forms/NumberInput.jsx
```diff
@@ -31,73 +69,77 @@ export class NumberInput extends React.Component {
...
 
 		return {
 			value: isNewValue ? nextProps.value : prevState.value,
 		};
 	}
 
-	_updateValueByStep(isIncreasing) {
-		const currentVal = new Number(this.state.value);
-		const step = new Number(this.props.step);
+	_updateValueByStep: Function;
```
mmcgahan: we should avoid using `Function` (just like `Object`) - this is `boolean => number`

src/forms/NumberInput.jsx
```diff
@@ -31,73 +69,77 @@ export class NumberInput extends React.Component {
...
+	_updateValueByStep(isIncreasing: boolean) {
```
mmcgahan: might as well make these class properties, and make `this.state` a class property, and then remove the `constructor` entirely.

src/forms/NumberInput.jsx
```diff
@@ -31,73 +69,77 @@ export class NumberInput extends React.Component {
...
+			return this.props.min;
+		}
+
+		if (this.props.max && newValue > this.props.max) {
 			newValue = this.props.max;
-		} else if (newValue < this.props.min) {
-			newValue = this.props.min;
 		}
 
 		return newValue;
```
mmcgahan: these `if` statements are just a workaround by initial author who didn't know how to write a min-max expression - let's use this:

```js
const newValue = isIncreasing ? currentVal + step : currentVal - step;
return Math.min(Math.max(newValue, this.props.min), this.props.max || Infinity);
```

feel free to make an intermediate variable to make it a bit more legible

```js
const newValue = isIncreasing ? currentVal + step : currentVal - step;
const minConstrained = Math.max(newValue, this.props.min);
return Math.min(minConstrained, this.props.max || Infinity);
```

src/forms/NumberInput.jsx
```diff
@@ -12,11 +12,49 @@ export const FAUX_INPUT_CLASS = 'fauxInput';
...
+	min: number,
+	name: string,
+	decrementAction: Function,
+	incrementAction: Function,
+	_updateValueByStep: Function,
+	onBlur: Function,
+	onFocus: Function,
+	onKeyDown: Function,
+	onChange?: Function,
+	step: number,
```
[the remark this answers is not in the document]
eilinora: stripping `parseInt` it should really just be a number

PR #598, opened 2018-07-17
Fix prop tables

mperrotti:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-600

#### Description
Prop tables weren't rendering any information in the info panel (can be seen by clicking the blue "Show Info" button in the upper right). Now it's rendering the data, and I've added descriptions for every prop

Before:
screen shot 2018-07-17 at 3 25 09 pm

After:
screen shot 2018-07-17 at 3 25 26 pm


src/__snapshots__/SignupModal.test.jsx.snap
```diff
@@ -1,22 +1,22 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`SignupModal should match the snapshot 1`] = `
-<Modal
+<ModalComponent
```
mmcgahan: these display names are going to get real noisy real fast - I hate to make you go through this again, but it would be super helpful to manually assign the displayName, e.g.

```js
const Modal = withLoading(ModalComponent);
Modal.displayName = 'Modal';
export default Modal;

Comment (mperrotti) [reply src/__snapshots__/SignupModal.test.jsx.snap]: Ahh, I like that. Worth it :)

File: src/interactive/__snapshots__/accordionPanel.test.jsx.snap [hunk 2]
```diff
@@ -172,7 +172,7 @@ exports[`AccordionPanel Panel with right aligned icon exists and renders icon le
             <FlexItem
               id="label-firstsection"
             >
-              <FlexItem
+              <FlexItemComponent
```
mperrotti: @mmcgahan - Some snapshots are rendering components wrapped in `withLoading` with `NameComponent` wrapped in `Component`.
Do you know why this happens? Is this something we need to change or "fix"?

mmcgahan: that's expected - since `withLoading` adds a wrapper around the base component, `mount`-rendered snapshots will show the 'outer' layer with the expected component name as well as the 'inner' component with the real component's name. We should generally use `shallow` for snapshots

PR #599, opened 2018-07-18
WC 252 controlled time

mmcgahan:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WC-252

#### Description

This is a major refactor of `TimeInput` to create a new child time input component that allows a parent component to be the single source of truth for the time `value` prop. `TimeInput` is now just a simple state-supplying shell around `InputTime`, which supplies a standard `<input type="time" />` implementation as well as falling back to a completely-rewritten 3-text-field 'fieldset' input for browsers that do not suport `input[type=time]`.

Most of the diff is in the 3-text-field `<FieldsetTime>` - the corresponding behavior was previously implemented in `TimeInput` directly, and maintained quite a bit of internal component state and manually enforced min/max/formatting rules on each input. The new implementation heavily relies on `LocalTime` objects from js-joda to make it easier to ensure 'correctness' for both display and `onChange` callback arguments.


mmcgahan:
I'll request reviewers once I complete some of the TODOs

mmcgahan:
@mperrotti I think you'll be in the best position to look into whether the new `FieldsetTime` matches the behavior of the fallback time input behavior of the original `TimeInput`. @eilinora and @jaclynj probably most helpful if you have a scan through to look for things that could use more explanation/commentin/reorganization to make it clear. Also if there are any concerns with type annotations and such.

I'm eager to get this merged quickly to unblock some of my Event Schedule work, but it's also important that this is RIGHT since it's a component that will likely be used for a long time in a lot of places

src/forms/InputTime.jsx
```diff
@@ -0,0 +1,120 @@
+// @flow
```
mperrotti: We should add stories for this too instead of just having stories for `FieldsetTime`

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+			),
+		};
+		const displayId = id || name;
+
+		const { hour, minute, meridian } = getValueComponents(is24Hr, value);
+		return (
+			<fieldset className={classNames.fauxInput}>
+				<Flex noGutters>
+					<FlexItem shrink>
+						<input
```
mperrotti: We should probably provide visually hidden `<label>`s for each input. This can be accomplished by adding the class `visibility--a11yHide`

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+					disabled,
+					error,
+				}
+			),
+		};
+		const displayId = id || name;
+
+		const { hour, minute, meridian } = getValueComponents(is24Hr, value);
+		return (
+			<fieldset className={classNames.fauxInput}>
```
mperrotti: Would be nice to show a label of some kind in the stories - I recommend using a `<legend>` and rendering whatever was passed to a `label` prop (most likely will come down from the `InputTime` component)

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+		this.props.onChange(localTime.with(ChronoField.AMPM_OF_DAY, meridian).toString());
+	};
+
+	// Handle up/down arrow behavior in hour/minute fields
+	onNumberKeyDown = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
+		if (!KEY_INCREMENT[e.keyCode]) {
+			return;
+		}
+		e.preventDefault(); // don't do anything but what I tell you
+		const stepMultiplier = e.shiftKey ? 10 : 1; // hold shift for fast increment
```
mperrotti: Thank you for porting this over from `TimeInput` <3

src/forms/InputTime.jsx
```diff
@@ -0,0 +1,120 @@
...
+import * as React from 'react';
+import cx from 'classnames';
+
+import withErrorList from '../utils/components/withErrorList';
+import FieldsetTime from './FieldsetTime';
+
+type Props = React.ElementConfig<HTMLInputElement> & {
```
mperrotti: Could you add Docgen comments for these props? You can see an example in [this PR](https://github.com/meetup/meetup-web-components/pull/598/files#diff-8c4ea546d2c1ed928f5f523260d0001cR19)

src/forms/InputTime.jsx
```diff
@@ -0,0 +1,120 @@
...
+	name: string,
```
mperrotti: Could you add Docgen comments for these props? You can see an [example in this PR](https://github.com/meetup/meetup-web-components/pull/598/files#diff-8c4ea546d2c1ed928f5f523260d0001cR19)

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+import Select from './Select';
+
+export const HOURS_INPUT_CLASS = 'timeInput-hours';
+export const HOURS_INPUT_NAME = 'hour';
+export const MINUTES_INPUT_CLASS = 'timeInput-minutes';
+export const MINUTES_INPUT_NAME = 'minute';
+export const MERIDIAN_INPUT_CLASS = 'timeInput-meridian';
+
+type Props = React.ElementConfig<'input'> & {
+	value: string, // empty string or HH:mm
```
mperrotti: Could you add Docgen comments for these props? You can see an [example in this PR](https://github.com/meetup/meetup-web-components/pull/598/files#diff-8c4ea546d2c1ed928f5f523260d0001cR19)

mmcgahan:
thanks for the review @mperrotti. A couple quick responses

> Would be nice if we could make hours < 10 have have 2 digits (e.g.: 2:00 => 02:00)
The current setup shows the leading zero for 24 hour time, but not for 12 hour time, which is a relatively common formatting distinction between the two systems. I'd like to keep it

> It's a little weird to me that holding "shift" and hitting up or down in the hours input doesn't stop at 12, but keeps looping through hours
That seems intuitive to me since clocks normally loop? 8am + 10hrs = 6pm rather than 8am + 10hrs = 12pm?

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+		const meridian = parseInt(e.target.value, 10);
+		this.props.onChange(localTime.with(ChronoField.AMPM_OF_DAY, meridian).toString());
+	};
+
+	// Handle up/down arrow behavior in hour/minute fields
+	onNumberKeyDown = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
+		if (!KEY_INCREMENT[e.keyCode]) {
+			return;
+		}
+		e.preventDefault(); // don't do anything but what I tell you
```
eilinora: :)

src/forms/FieldsetTime.jsx
mmcgahan: I'll clarify through docs, but this component is really intended to be a replacement specifically for `<input type="time" />` - labelling responsibilities are handled by parent, e.g. `InputTime`, which is generally the component that will be used by consumers.

Definitely need to make that clear in the code, though - for consistently, this component should _never_ be used directly in consumer code.

mmcgahan: In which case, I should probably just port the stories over to `InputTime` entirely, and not have `FieldsetTime.story.jsx` at all

src/forms/FieldsetTime.jsx
```diff
@@ -0,0 +1,247 @@
...
+							<Select
+								id={`${displayId}-meridian`}
+								name="mwc-meridian"
+								className={classNames.meridian}
+								disabled={disabled}
+								value={meridian.toString()} // 0 or 1
+								onChange={this.onMeridianChange}
+							>
+								<option value="0">AM</option>
+								<option value="1">PM</option>
```
jaclynj: AM and PM won't ever need to be translated?

mmcgahan: it's from Latin, so I don't think so - at least, the original implementation didn't take translations for it. I want to say that maybe there are kanji characters for am/pm that we might eventually want to support, but I'm not sure - I think we generally show 24Hr time for languages other than English. Either way, out of scope for this update :)

mmcgahan:
screen shot 2018-07-20 at 12 05 56

src/forms/InputTime.jsx
mmcgahan: addressed by removing `FieldsetTime` story and adding lots of explicit variants for `InputTime` variants

src/forms/InputTime.jsx
mmcgahan: done

src/forms/FieldsetTime.jsx
mmcgahan: addressed by removing `FieldsetTime` story

mmcgahan:
@mperrotti I think comments have been addressed - thanks for the review. Lots of specific stories added for `InputTime` as shown above

PR #607, opened 2018-07-24
increase width of numberInput

M0nica:
#### Description
Make the numberInput field width longer because currently if the max is 10 it will cut off if a user types in 100. The expected behavior is that if a user is able to type in 100 that 100 should not be cut off.

screen shot 2018-07-24 at 3 27 10 pm
screen shot 2018-07-24 at 3 27 23 pm


M0nica:
Thanks @mperrotti. I just pushed a change to address your comment.

mperrotti:
Great! Feel free to merge

PR #615, opened 2018-08-07
MP-1866 update component to use is_pro_admin from self

acmei:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-1866

#### Description
Update the Nav component to use `is_pro_admin` from self

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -164,7 +164,7 @@ export class Nav extends React.Component {
 		];
 
 		let authItems = [
-			self.isProMember && {
+			self.is_pro_admin && {
 				shrink: true,
 				linkTo: media.isAtMediumUp ? proDashboard.link : '',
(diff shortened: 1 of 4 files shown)


src/navigation/Nav.jsx
```diff
@@ -164,7 +164,7 @@ export class Nav extends React.Component {
 		];
 
 		let authItems = [
-			self.isProMember && {
```
rj1061: Why are we replacing `isProMember` with `is_pro_admin`? Are we limiting the scope of something we're displaying? I am quite confused.

chenrui333: naming convention due to old endpoint in chapstick

acmei: `isProMember` is a parameter we defined because we were doing a separate API call to get the info. Now the info comes from `self` so we don't need to define our own separate variable.

PR #617, opened 2018-08-09
Use highres_link in AvatarMember

andeliv:
#### Related issues
https://meetup.atlassian.net/browse/MG-2498
mup-web pr: https://github.com/meetup/mup-web/pull/4139

#### Description
From new member_list endpoint we receive only `highres_link` and `thumb_link` for Photo. Use `highres_link` as a fallback for large size in case `photo_link` is not present.

### DIFF
--- a/src/media/AvatarMember.jsx
+++ b/src/media/AvatarMember.jsx
@@ -19,7 +19,16 @@ class AvatarMember extends React.PureComponent {
 		const { member, org, fbFriend, className, ...other } = this.props;
 		const { big, large, xxlarge } = other;
 
-		const photoLink = big || large || xxlarge ? 'photo_link' : 'thumb_link';
+		let photoLink = 'thumb_link';
+
+		if (big || large || xxlarge) {
+			photoLink = 'photo_link';
+
+			// try `highres_link` in case `photo_link` is undefined
+			if ((member.photo || {})[photoLink] == undefined) {
+				photoLink = 'highres_link';
+			}
+		}
 		const showNoPhoto = (member.photo || {})[photoLink] == undefined;
 
 		const classNames = cx(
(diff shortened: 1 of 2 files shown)


eilinora:
Just need to increment the build # and update the change log but otherwise looks good

src/media/AvatarMember.jsx
```diff
@@ -19,7 +19,16 @@ class AvatarMember extends React.PureComponent {
...
+
+		if (big || large || xxlarge) {
+			photoLink = 'photo_link';
+
+			// try `highres_link` in case `photo_link` is undefined
+			if ((member.photo || {})[photoLink] == undefined) {
+				photoLink = 'highres_link';
+			}
+		}
 		const showNoPhoto = (member.photo || {})[photoLink] == undefined;
```
mmcgahan: This seems like something that could be made more clear by breaking the logic out into a pure function - 2 value reassignments, 3 levels of nesting, and 4 `||` fallbacks in just 8 lines of code is tough to work with even with comments. It's also best to avoid checks for `undefined` if you're more specifically interested in the truthy value.

```js
const getPhoto = (photo, size) => {
  if (!photo) {
    return undefined; // well defined empty case
  }
  switch(size) {
    case 'big':
    case 'large':
    case 'xxlarge': // clear handling of these 3 overlapping size handlers
      return photo.photo_link || photo.highres_link; // one place to define fallback
    default:
      return photo.thumb_link
  }
}

...

const photoSize = big || large || xxlarge ? 'big' : 'default';
const photoLink = getPhoto(member.photo, photoSize);
const showNoPhoto = typeof photoLink !== 'string'; // _any_ non-string value should be considered invalid.
```

We'll eventually want

PR #619, opened 2018-08-14
Upgrade jest 23

mmcgahan:
#### Description

Upgrade to Jest 23 - not a lot of changes, but had to rewrite history because there was a 157MB snapshot that couldn't be updated...


mmcgahan:
SV

PR #621, opened 2018-08-16
mark all notification as read when clicked on Notification

andrew-bykovets:
#### Related issues
Fixes https://meetup.atlassian.net/browse/PYMT-157

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -66,7 +66,7 @@ export class Nav extends React.Component {
 	 */
 	onClickDropdownAction() {
 		this.setState(() => ({ showMobileDashboard: true }));
-		if (this.props.onOpenClickHandler) {
+		if (this.props.markAllAsReadOnOpen) {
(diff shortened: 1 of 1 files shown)


src/navigation/Nav.jsx
```diff
@@ -395,6 +407,7 @@ Nav.propTypes = {
 
 	/** The locale code of the current user */
 	localeCode: PropTypes.string,
+	onOpenClickHandler: PropTypes.bool,
```
komlev: should be function?

src/navigation/Nav.jsx
```diff
@@ -76,6 +79,15 @@ export class Nav extends React.Component {
 		this.setState(() => ({ showMobileDashboard: false }));
 	}
 
+	markAllNotifAsRead = () => {
+		const { notifications } = this.props.navItems;
+		const { list, notificationsDropdown } = notifications;
+		if (list) {
```
komlev: should check on length, it is going to error out on list[0]

PR #623, opened 2018-08-22
PYMT-157: Should mark notifications as read on the `mediumup` size screens 

vitzym:
#### Related issues
Fixes https://meetup.atlassian.net/browse/PYMT-157


src/navigation/Nav.jsx
```diff
@@ -264,7 +281,7 @@ export class Nav extends React.Component {
 				),
 				onClickAction:
 					media.isAtMediumUp && !isNotificationsLoaded
-						? notifications.getNotificationsQuery
+						? this.onClickDropdownAction()
```
mmcgahan: this is a bug - shouldn't call the function here, only when clicked

mmcgahan: I've pushed a fix to master

PR #629, opened 2018-09-04
allow manual toggle of tooltip

browne0:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MG-2364

#### Description
Accidentally removed the ability to click on the x in the tooltip to close it. This adds it back and allows the ability to close the tooltip by clicking outside of it.

### DIFF
--- a/src/interactive/Tooltip.jsx
+++ b/src/interactive/Tooltip.jsx
@@ -259,7 +259,16 @@ Tooltip.propTypes = {
 	/** Whether to render the dropdown content directly in the component instead of bulling it out and attaching to the document root */
 	noPortal: PropTypes.bool,
 	/** Required function to close the dropdown when working with a manually toggled tooltip */
-	onClose: PropTypes.func,
+	onClose: (props, propName, componentName) => {
+		if (
+			props.manualToggle === true &&
+			(props[propName] === undefined || typeof props[propName] !== 'function')
+		) {
+			return new Error(
+				'The onClose function is required if manualToggle is passed to this component!'
+			);
+		}
+	},
 };
 
 export default Tooltip;
--- a/src/interactive/Tooltip.jsx
+++ b/src/interactive/Tooltip.jsx
@@ -22,7 +22,47 @@ class Tooltip extends React.PureComponent {
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


src/interactive/Tooltip.jsx
```diff
@@ -94,15 +134,15 @@ class Tooltip extends React.PureComponent {
 		return (
 			<div
 				className={classNames.dropdown}
-				onMouseLeave={this.closeContent}
+				onMouseLeave={manualToggle ? undefined : this.closeContent}
```
Mikodin: Nitpick - this and the following could be `manualToggle && this.closeContent`, though the ternary may be a little more explicit

src/interactive/Tooltip.jsx
```diff
@@ -22,7 +22,47 @@ class Tooltip extends React.PureComponent {
...
+		if (this.props.manualToggle) {
+			document.removeEventListener('click', this.handleClickOutsideContent);
+		}
+	}
+
+	static getDerivedStateFromProps(nextProps, prevState) {
+		if (nextProps.manualToggle && prevState.isActive !== nextProps.isActive) {
+			return { isActive: nextProps.isActive };
+		}
+		return prevState;
```
Mikodin: Is this the same as `return null` ?

browne0: yes, but this way imo is more explicit of what it is actually doing 😄

src/interactive/Tooltip.jsx
browne0: Agreed, though I believe being explicit is beneficial for future people reading code, rather than saving some time now 🙂

src/interactive/Tooltip.jsx
Mikodin: No doubt, cool with me 🗻 .  Just wanted to be sure haha

jaclynj: Interesting. I disagree though, "Returning null is slightly more performant (since React won't have to merge the state with itself). It also more clearly signals the intent of not updating anything." See here and comments related: https://github.com/reactjs/rfcs/issues/26#issuecomment-395407230

PR #630, opened 2018-09-04
update tooltip component

browne0:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MG-2364

#### Description
Address some PR feedback, as well as passing along the event object from the original onClose.

### DIFF
--- a/src/interactive/infoToggle.story.jsx
+++ b/src/interactive/infoToggle.story.jsx
@@ -7,6 +7,45 @@ import { action } from '@storybook/addon-actions';
 
 import InfoToggle from './InfoToggle';
 
+class ManualInfoToggle extends React.PureComponent {
+	state = {
+		tooltipOpen: false,
+	};
+
+	toggleTooltip = () => {
+		this.setState(prevState => ({ tooltipOpen: !prevState.tooltipOpen }));
+	};
+
+	closeTooltip = () => {
+		this.setState(() => ({ tooltipOpen: false }));
+	};
+
+	render() {
+		return (
+			<InfoToggle
+				label="label"
+				tooltipId="info-tooltip"
+				tooltipContent={
+					<div className="runningText padding--all">
+						<p>
+							This is a tooltip is being passed props that adjust the
+							alignment of the popup bubble to point to the trigger.
+						</p>
+					</div>
+				}
+				onClick={this.toggleTooltip}
+				tooltipProps={{
+					align: 'right',
+					withClose: true,
+					manualToggle: true,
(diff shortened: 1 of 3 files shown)


src/interactive/infoToggle.story.jsx
```diff
@@ -2,11 +2,50 @@ import React from 'react';
...
+	state = {
+		tooltipOpen: false,
+	};
+
+	toggleTooltip = () => {
+		this.setState(prevState => ({ tooltipOpen: !prevState.tooltipOpen }));
+	};
+
+	closeTooltip = () => {
+		this.setState(() => ({ tooltipOpen: false }));
```
mmcgahan: sort of marginal benefit here, but consider using composition rather than defining two separate methods that each update `state.tooltipOpen`

```js
changeTooltip(tooltipOpen) {
  this.setState({ tooltipOpen });
}
toggleTooltip = () => this.changeTooltip(!this.state.tooltipOpen)
closeTooltip = () => this.changeTooltip(false)
```

src/interactive/infoToggle.story.jsx
```diff
@@ -2,11 +2,50 @@ import React from 'react';
...
 
 import InfoToggle from './InfoToggle';
 
+class ManualInfoToggle extends React.PureComponent {
+	state = {
+		tooltipOpen: false,
+	};
+
+	toggleTooltip = () => {
+		this.setState(prevState => ({ tooltipOpen: !prevState.tooltipOpen }));
```
mmcgahan: avoid calling the arg to the updater function `prevState` - it's actually the _future_ value of state when state updates are being applied, so `prevState` is misleading. Just call it `state`

browne0: https://reactjs.org/docs/react-component.html#setstate 
according to the docs it is referencing the previous state

mmcgahan: ah, I stand corrected. I still find it odd since it could be a future value relative to `this.state`, but ok.

mmcgahan: maybe i should use this as an excuse to to tweet @ dan_abramov

browne0: 💯 i support it haha

mmcgahan: if he ever answers... https://twitter.com/mikemcgahan/status/1037074327269568512

mmcgahan: React docs update PR https://github.com/reactjs/reactjs.org/pull/1155

PR #635, opened 2018-09-11
add photoSrc property for AvatarMember

andeliv:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MG-2779

#### Description
It seems like we're using the highres image to display a thumbnail next to the member. This is causing some performance issues. We want to use a smaller image.

Custom photo link removes dependency on `size` prop and gives more flexibility for the component
e.g. we could use large avatars with `thumb_link` to improve performance


eilinora:
Approved, but see comments in the mupweb PR :)

PR #636, opened 2018-09-12
Adding className option + wrapper to Select component fields

eilinora:
#### Description
- goal is to make it more flexible for styling

- no visual change

### DIFF
--- a/src/forms/Select.jsx
+++ b/src/forms/Select.jsx
@@ -78,7 +78,7 @@ export const SelectInput = (props: Props) => {
 				</label>
 			)}
 			{helperText && <div className={classNames.helperText}>{helperText}</div>}
-			<div className={cx('selectWrapper', selectClassName)}>
+			<div className={`selectWrapper ${selectClassName || ''}`}>
 				<select
 					name={name}
 					id={id || name}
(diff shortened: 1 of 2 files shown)


src/forms/Select.jsx
```diff
@@ -74,14 +78,16 @@ export const SelectInput = (props: Props) => {
...
 			{helperText && <div className={classNames.helperText}>{helperText}</div>}
-			<select
-				name={name}
-				id={id || name}
-				required={Boolean(required)}
-				className={classNames.field}
-				{...other}
-			/>
-			<Icon className="select-customArrow" shape="chevron-down" size="xs" />
+			<div className={cx('selectWrapper', selectClassName)}>
```
katzelein: Didn't Dr. Mike say we should be using cx only when we really need it, but that for something like this we could just do `selectWrapper ${selectClassName}`?

eilinora: still personally prefer the other way, but updated :)

mmcgahan: This case, where you're combining a string with a variable, is sort of borderline on which implementation is 'better', so I personally wouldn't call it out either way. String templates in JS aren't particularly elegant, but I don't think either approach is especially awkward, so unless you feel strongly about it, don't feel compelled to call it out in the future :).

PR #637, opened 2018-09-18
Allows textarea resize to be disabled

katzelein:
#### Related issues
Will be used to complete https://meetup.atlassian.net/browse/MG-2843

#### Description
Allows resizing to be disabled on textarea through optional prop.

screen shot 2018-09-18 at 10 07 38 am

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -91,6 +91,7 @@ export class Textarea extends React.PureComponent {
 			helperText,
 			required,
 			requiredText,
+			disableResize,
 			...other
 		} = this.props;
 
@@ -100,6 +101,7 @@ export class Textarea extends React.PureComponent {
 				{
 					'field--error': error,
 					'textarea--autoheight': autosize,
+					'textarea-no-resize': disableResize,
 				},
 				className
 			),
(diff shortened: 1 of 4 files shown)


src/forms/Textarea.jsx
```diff
@@ -100,6 +101,7 @@ export class Textarea extends React.PureComponent {
 				{
 					'field--error': error,
 					'textarea--autoheight': autosize,
+					'textarea-no-resize': disableResize,
 				},
```
katzelein: It might make sense to also disable resizing when `autosize` is true, but I wanted to leave the option open.

src/forms/Textarea.jsx
```diff
@@ -100,6 +101,7 @@ export class Textarea extends React.PureComponent {
 				{
 					'field--error': error,
 					'textarea--autoheight': autosize,
+					'textarea-no-resize': disableResize,
```
katzelein: It might make sense to also disable resizing when autosize is true, but I wanted to leave the option open.

src/forms/textarea.test.jsx
```diff
@@ -153,7 +153,7 @@ describe('Textarea', function() {
 
 		expect(autosizePlugin.default).toHaveBeenCalled();
 	});
-	it('should set be able to min and max height', function() {
+	it('should be able to set min and max height', function() {
```
katzelein: Just a typo correction.

PR #639, opened 2018-09-20
Allow screen readers to read alt text on the Avatar component

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/sds-676

#### Description
This pr adds `aria-label` to `<Avatar />` and sets the value to the `alt` prop. I added a default prop to `<Avatar />` instead of requiring one because I didn't want to introduce a breaking change and am unsure if any teams builds fail on lint errors.

### DIFF
--- a/src/media/Avatar.jsx
+++ b/src/media/Avatar.jsx
@@ -46,6 +46,7 @@ class Avatar extends React.PureComponent {
 
 		const aria = {
 			role: 'img',
+			'aria-label': alt,
 		};
 
 		const computedProps = {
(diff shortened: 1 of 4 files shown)


hsbacot:
@mmcgahan Yes, I've checked using voice over in storybook and it is not reading the alt twice.

There are a few things to note about this that should be attached here. I'm also not sure of the history behind the component so I wanted to alter it as little as possible.

1) I went with `aria-label` because the primary takeaway from the story was that voice over was not reading the alt text as is.
2) The props are passed to both the `span` and `a` causing the `a` to be read in voice over as an image since we set `role='img'` on the node. 

I wanted to submit this to allow for the alt text to be read.

hsbacot:
I'm merging it in order to get the alt text to voice over. I plan on revisiting this later as well.

PR #641, opened 2018-09-27
fixing navbar accessibility

komlev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-2058

#### Description
Navbar component fails couple of a11y guidelines. There was no ability to control notifications/profile dropdown with keyboard. This PR fixes it.


src/interactive/Dropdown.jsx
```diff
@@ -58,7 +58,11 @@ class Dropdown extends React.PureComponent {
 
 	onKeyDown(e) {
 		if (e.key === 'Enter' && this.state.isActive) {
-			this.closeContent();
+			// postpoining closure till next cycle
```
kcode20: nit: typo: postponing

src/interactive/Dropdown.jsx
```diff
@@ -58,7 +58,11 @@ class Dropdown extends React.PureComponent {
...
+			// otherwise closure cancels default keyboard events
+			setTimeout(() => {
+				this.closeContent();
+			}, 0);
```
kcode20: Little confused, we are calling `this.closeContent` after 0 milliseconds? Is `setTimeout` really postponing if we aren't specifying a larger wait interval?

komlev: It just calls that method async, rather than running that code right away.

browne0: no need to wrap the this.closeContent in a function

`setTimeout(this.closeContent, 0)`

hsbacot:
I think this is a good start and an improvement over what we have. This doesn't appear to allow the contents of the notifications and the profile dropdown to be accessed with a keyboard now that they are interactive.

I'm very new so I'm unclear of the immediate goals of this pr and could be missing context. Could you attach the guidelines that are failing?

komlev:
@hsbacot it is mainly
2.1.1 Keyboard from WCAG
Navbar's Notifications dropdown and Profile dropdown are not operable with the keyboard.

src/navigation/Nav.jsx
```diff
@@ -304,19 +305,21 @@ export class Nav extends React.Component {
...
-							shrink
-							className="display--none atMedium_display--block"
-						>
-							<Icon
-								shape="chevron-down"
-								size="xxs"
-								className="padding--left-half"
-							/>
-						</FlexItem>
+						<Button aria-haspopup reset aria-label={profile.label}>
```
mmcgahan: This `button` element is actually nested in a `button` created by `NavItem.jsx`, which results in an invalid DOM that prints a warning in dev and causes a pretty major UI layout issue on slow connections (https://meetup.atlassian.net/browse/MG-3133). The `button` created by `NavItem` already has `aria-haspopup`, so I've pushed an update that removes this `<Button />` wrapper and passes `aria-label` directly to the `Flex` wrapper, which should address the voiceover a11y concern

PR #642, opened 2018-09-27
polishing Dropdown component

juanbiberretta:
#### Related issues
Written to help with 
https://meetup.atlassian.net/browse/MG-2863
https://meetup.atlassian.net/browse/MG-2991

#### Description
- Added `closeOnBodyClick` prop to reduce overhead every time someones wants the Dropdown to close after they click the Dropdown (think menu items that take re-render the page). Previously you had to implement a `manualToggle` function. 
- Flow typed Dropdown. 
- Fixed error preventing from displaying `Dropdown with menuItems` on Storybook.


src/interactive/Dropdown.jsx
```diff
@@ -1,68 +1,110 @@
...
+	manualToggle?: (e: SyntheticKeyboardEvent<*> | SyntheticMouseEvent<*>) => void,
 
-	closeContent(e) {
+	/** The largest width the dropdown content can be */
+	maxWidth?: string | number,
+
+	/** The smallest width the dropdown content can be */
+	minWidth?: string | number,
+
+	/** Whether to render the dropdown content directly in the component instead of bulling it out and attaching to the document root */
```
mmcgahan: typo 'bulling'. might have meant 'building', but that's sort of an ambiguous way to describe rendering

src/interactive/Dropdown.jsx
```diff
@@ -71,27 +113,38 @@ class Dropdown extends React.PureComponent {
...
-			this.closeContent();
+			this.closeContent(e);
 		}
-	}
+	};
 
 	componentDidMount() {
-		document.body.addEventListener('click', this.onBodyClick);
-		document.body.addEventListener('keydown', this.onBodyKeyDown);
+		if (document.body) {
```
mmcgahan: curious about this guard - I heard mention there was an error related to this, but not sure how `document.body` would not exist in `componentDidMount` - any info?

src/interactive/Dropdown.jsx
```diff
@@ -1,68 +1,110 @@
...
+	maxWidth?: string | number,
+
+	/** The smallest width the dropdown content can be */
+	minWidth?: string | number,
+
+	/** Whether to render the dropdown content directly in the component instead of bulling it out and attaching to the document root */
+	noPortal?: boolean,
+
+	/** Props to pass to the `Downshift` component */
+	downshiftProps?: Object,
```
katzelein: You should flesh this out

src/interactive/Dropdown.jsx
```diff
@@ -71,27 +113,38 @@ class Dropdown extends React.PureComponent {
...
-	}
+	};
 
 	componentDidMount() {
-		document.body.addEventListener('click', this.onBodyClick);
-		document.body.addEventListener('keydown', this.onBodyKeyDown);
+		if (document.body) {
+			document.body.addEventListener('click', this.onBodyClick);
+		}
+		if (document.body) {
```
mmcgahan: just put one guard around both calls to `addEventListener` rather than two

eilinora: no need for the separate ifs :)

src/interactive/Dropdown.jsx
```diff
@@ -71,27 +113,38 @@ class Dropdown extends React.PureComponent {
 			ref => !ref.contains(e.target)
 		);
 
-		if (isNotDropdownClick) {
+		if (isNotDropdownClick || (!isNotDropdownClick && this.props.closeOnBodyClick)) {
```
mmcgahan: I'm assuming this is the fix (always good to call out the specific changes with a comment in PRs that have a lot of 'maintenance noise' like this).

No need for `!isNotDropdownClick` here - `if (isNotDropdownClick || this.props.closeOnBodyClick)` is equivalent. However, since `this.props.closeOnBodyClick` is like a universal override, I recommend giving it special treatment at the top of the handler

```js
onBodyClick = e => {
  if (this.props.closeOnBodyClick) {
    closeContent(e);
  }
  // continue with `isNotDropdownClick` logic
}
```

That avoids unnecessary calculations and makes the override behavior very clear

package.json
```diff
@@ -140,16 +140,16 @@
...
     "**/*.{js,jsx}": [
-	  "eslint",
-	  "git add"
+      "eslint",
+      "git add"
     ]
   }
-}
\ No newline at end of file
+}
```
katzelein: Considering the problems that Dr. Mike has been having with mwp packages, you should probably have him look at the package.json and swarm_constants files to confirm the changes are necessary.

src/interactive/Dropdown.jsx
juanbiberretta: Flow is actually so wierd about it:
screen shot 2018-09-27 at 2 14 29 pm

package.json
juanbiberretta: Well @browne0 guided me through this, it's basically just creating a stub for the constants file, if I didn't do that it would complain when trying to import the module into `Dropdown.jsx`

src/interactive/Dropdown.jsx
mmcgahan: ah, amazing, I see. Flow can't guarantee that calling `addEventListener` doesn't change the value of `document`, which is why it errors... hmmm I guess leave the two checks for now

src/interactive/Dropdown.jsx
juanbiberretta: `Downshift` is an npm package, not sure where to find the flow definitions :(

src/interactive/Dropdown.jsx
juanbiberretta: sounds good, thanks! and yes, should've highlighted that

src/interactive/Dropdown.jsx
mmcgahan: Looks like [there are a lot of possibilities](https://github.com/paypal/downshift#table-of-contents) - I think it's fair to leave that out of scope for this PR, but if you could include a link to the 'downshift' repo in the code comment above, that would be helpful for reference.

I think Downshift is a target for replacement long term with [Reach UI](https://ui.reach.tech/), which @hsbacot has been mulling a bit.

hsbacot:
I'm a little late to the discussion here but I'm curious around why we'd want to add this as a prop and encapsulate a lot of opt-in functionality within `<Dropdopwn />`. Have we considered using refs for something like this?

@mmcgahan yes, this component is something I'd like to rethink when considering reach-ui and the limited control of the contents it gives back to the consumer. Making a lot of decisions and encapsulating that logic could also be beneficial here too. I'm so naive to how this is used it's hard to say. Thoughts? @mmcgahan @juanbiberretta @eilinora

mmcgahan:
@hsbacot Yeah I'm not completely familiar with the original constraints of this component, but I think ideally the `closeOnBodyClick` behavior would be the default (and possibly not even opt-out) behavior of the component.

My naive suspicion is that the dropdown was initially used in interfaces where it showed a list of links to other pages (e.g. the Navigation bar at the top of mup-web), so closing-on-click wasn't really necessary. It's very possible we could just make the dropdown close on _any_ click, always, although maybe we _should_ make it opt-out for cases like having an `<input>` inside a dropdown? I think that kind of UI might exist in Pro for some list-filtering options

hsbacot:
@mmcgahan good point. I'm having a hard time finding examples that aren't links in around where it's used. I'm also curious about this being on all contents. I feel like this might make more sense on the contents or items but our current api might make that difficult to achieve. I think this is much more complex than a yes/no and would be very hard to maintain if we started that way and then had to retroactively support inputs and checkboxes etc.

src/interactive/Dropdown.jsx
mmcgahan: Just was thinking about this a little more, and it's _possible_ that you wouldn't get the Flow error if you used `window.document.body` rather than `document.body` because I think Flow treats `window` as immutable, so it _might_ not worry about `document` getting overwritten... really not sure though. Just wanted to leave the comment :)

package.json
eilinora: The problems were more with code changes in mwp, these should be fine :)

src/interactive/Dropdown.jsx
browne0: you can just use window in this case

PR #643, opened 2018-09-28
cleaning up closeOnBodyClick

juanbiberretta:
#### Description
- After testing the new prop and talking to @hsbacot  it became apparent that there's no easy way to close the Dropdown after an action has been taken on the menu item contained within. 
- There's now to guarantee that the menu item action (owned by parent and passed through `menuItems`) will fire before the Dropdown closes. As of right now, `manualToggle`, alongside `isActive` is our best bet. 
- This PR removes all `closeOnBodyClick` functionality since it's basically useless (at least the previous PR flowed the file and fixed some bugs).

### DIFF
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -50,9 +50,6 @@ type Props = {
 	/** Props to pass to the `Downshift` component */
 	downshiftProps?: Object,
 
-	/** Whether the dropdown should close when a click on the body is registered */
-	closeOnBodyClick?: boolean,
-
 	/** Optional custom function to execute on dropdown click */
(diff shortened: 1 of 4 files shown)


src/interactive/Dropdown.jsx
```diff
@@ -50,9 +50,6 @@ type Props = {
 	/** Props to pass to the `Downshift` component */
 	downshiftProps?: Object,
 
-	/** Whether the dropdown should close when a click on the body is registered */
-	closeOnBodyClick?: boolean,
```
eilinora: if we are removing this its  a breaking change as i believe this is a breaking change in mupweb which is fine, we just need to note that in the version of this by incrementing the version # accordingly and when this upgrade happens in mupweb handle the removal :)

PR #644, opened 2018-09-28
implemente new closeOnMenuItemClick prop and functionality

juanbiberretta:
#### Description
- Read changelog for description. 
- This PR builds on the idea put forth in: https://github.com/meetup/meetup-web-components/pull/642
- While that PR got merged, the functionality wasn't there, this PR renames some variables and adjusts the logic to work accordingly. 
- Big thank you to @browne0 for helping me figure out a workaround.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [6.1]
+
+-   **New feature** `Dropdown` component now supports a `closeOnMenuItemClick` boolean prop that specifies
+    whether `Dropdown` should close after a menu item is clicked on (assuming `menuItems` has been passed in).
+    Hopefully this reduces overhead and the necessity to implement `manualToggle` and manage `isActive` to achieve
+    this behavior.
+
 # [6.0]
 
 -   **BREAKING CHANGE** `NumberInput`'s `onChange` callback will now receive a
--- a/src/interactive/Dropdown.jsx
+++ b/src/interactive/Dropdown.jsx
@@ -50,8 +50,9 @@ type Props = {
 	/** Props to pass to the `Downshift` component */
 	downshiftProps?: Object,
 
-	/** Whether the dropdown should close when a click on the body is registered */
-	closeOnBodyClick?: boolean,
+	/** Whether the dropdown should close after a menuItem click is fired. Used only in case
+	 when menuItems are passed in. */
+	closeOnMenuItemClick?: boolean,
 
 	/** Optional custom function to execute on dropdown click */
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# [6.1]
+
+-   **New feature** `Dropdown` component now supports a `closeOnMenuItemClick` boolean prop that specifies
+    whether `Dropdown` should close after a menu item is clicked on (assuming `menuItems` has been passed in).
+    Hopefully this reduces overhead and the necessity to implement `manualToggle` and manage `isActive` to achieve
+    this behavior.
```
browne0: nitpick for changelogs: 

should be `**NEW FEATURE**`

You should also add a breaking change noting that `closeOnBodyClick` is now removed. This will cause all instances of Dropdown in all clients to break. We can fix this in pro-web, but other consumers need to know about it as well.

src/interactive/Dropdown.jsx
```diff
@@ -106,18 +107,34 @@ class Dropdown extends React.PureComponent<Props, State> {
 		}
 	};
 
-	onBodyClick = (e: SyntheticMouseEvent<*>) => {
-		if (this.props.closeOnBodyClick) {
-			this.closeContent(e);
-			return;
+	/** Handles execution of menuItem action and then closes the dropdown
+	 	_if_ `closeOnMenuItemClick` is set to true and there are menuItems.
+	 	If there are menuItems but `closeOnMenuItemClick` is false it'll just
```
browne0: I think the second sentence isn't needed, it's just describing what the code does

src/interactive/Dropdown.jsx
```diff
@@ -106,18 +107,37 @@ class Dropdown extends React.PureComponent<Props, State> {
...
+	 	fire the action and keep the dropdown open. */
+	onDropdownMenuItemClick = (
+		e: SyntheticMouseEvent<*>,
+		itemProps: { onClick?: (SyntheticMouseEvent<*>) => void }
```
hsbacot: We don't need the entire object of item props here since `onClick` is the only property used

src/interactive/Dropdown.jsx
```diff
@@ -106,18 +107,37 @@ class Dropdown extends React.PureComponent<Props, State> {
...
+	) => {
+		if (itemProps.onClick) {
+			if (!this.contentRef || !this.triggerRef) {
+				return;
+			}
+			itemProps.onClick(e);
+
+			console.log(itemProps);
```
hsbacot: NIT: Let's remove these logs

PR #646, opened 2018-10-03
adding webkit overflow scroll so that ios scrolls modal content

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/sds-740

#### Description
This allows scrolling on containers that aren't the body on ios
More info here: https://github.com/tachyons-css/tachyons/issues/566

### DIFF
--- a/assets/scss/components/_modal.scss
+++ b/assets/scss/components/_modal.scss
@@ -58,6 +58,8 @@ $modal-width: 440px;
 }
 
 .view--modal {
+	-webkit-overflow-scrolling: touch;
+
 	.stripe-heroContent {
 		min-height: 0;
 		text-align: inherit;


hsbacot:
I pulled out the flowtyping of the component because I was running into issues flowtyping the dynamic mediaquery properties being set without refactoring the component. I want to look into how this is being used & how its tested before refactoring for the benefits of typing.

burrata:
Curious if this issue would be related to a recent Elevate ticket that's shown up on Android: https://meetup.atlassian.net/browse/MG-3100

hsbacot:
@burrata It shouldn't, this is only compatible with mobile safari
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

PR #647, opened 2018-10-04
updated dates to use js-joda LocalDate

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-730

#### Description
Converted `Date` objects in  the calendar component story book to use `LocalDate` from `js-joda`

### DIFF
--- a/src/forms/calendarComponent.story.jsx
+++ b/src/forms/calendarComponent.story.jsx
@@ -3,9 +3,12 @@ import CalendarComponent from './CalendarComponent';
 import { storiesOf } from '@storybook/react';
 import { decorateWithBasics, decorateWithInfo } from '../utils/decorators';
 import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { LocalDate } from 'js-joda';
 
 import Section from '../layout/Section';
 import Chunk from '../layout/Chunk';
+
+const jodaDate = LocalDate.now();
 
 storiesOf('Forms/CalendarComponent', module)
 	.addDecorator(decorateWithBasics)
@@ -17,10 +20,11 @@ storiesOf('Forms/CalendarComponent', module)
 				<CalendarComponent
 					name="event_time"
 					label="Start at"
-					value={new Date()}
+					value={jodaDate}
 					datepickerOptions={{
-						maxDate: boolean('max date today', false) && new Date(),
-						minDate: boolean('min date today', false) && new Date(),
+						maxDate: boolean('max date today', false) && jodaDate,
+						minDate:
+							boolean('min date today', false) && jodaDate.plusWeeks(16),
 					}}
 				/>
 			</Chunk>
@@ -44,7 +48,7 @@ storiesOf('Forms/CalendarComponent', module)
 				<CalendarComponent
 					name="event_time"
 					label="Start at"
-					value={new Date()}
+					value={jodaDate}
 					required
 					requiredText="(required)"
 				/>
@@ -57,7 +61,7 @@ storiesOf('Forms/CalendarComponent', module)
 				<CalendarComponent
 					name="event_time"
 					label="Start at"
-					value={new Date()}
+					value={jodaDate}
 					error="this is an error"
 				/>
 			</div>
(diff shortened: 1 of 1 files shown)


src/forms/calendarComponent.story.jsx
```diff
@@ -70,65 +74,56 @@ storiesOf('Forms/CalendarComponent', module)
...
 				<CalendarComponent
 					name="event_time"
 					label="From 2 days ago to 1 week from now"
-					value={now}
-					datepickerOptions={opts}
+					value={jodaDate}
+					datepickerOptions={{
+						allowInput: true,
+						minDate: jodaDate.minusDays(2),
+						maxDate: jodaDate.plusDays(7),
```
mmcgahan: I don't think this will work - does it? the `datepickerOptions` are values passed straight to flatpickr, which doesn't support js-joda objects

hsbacot: It shouldn't, I hadn't tried the validation on this yet but it would make sense to be a necessary conversion.

hsbacot: This will lead to an interesting interface, all of our dates take `js-joda` objects and all of the functionality extended from `flatpicker` wants a `Date` object. We do the conversion for the value in the component, would we want to wrap date range and all of this functionality as well?

mmcgahan: Hmm tough call. I lean toward _not_ trying to wrap the `options` offered by flatpickr - they're reasonably well-documented [here](https://flatpickr.js.org/options/) and I think treating that as a single source of truth for that config is simplest - e.g. I don't think people would assume that there was a `minDate` option available and then assume that it took a `LocalDate` value without having seen the docs, and once they see the docs, they should know the arg type. I supposed we could change the name of `datepickerOptions` to `flatpickrOptions`, although the flow type definitions are already specified [here](https://github.com/meetup/meetup-web-components/blob/30a537eb04b961f1e3e0b43902a45bcae61bd8b6/src/forms/CalendarComponent.jsx#L12) so once those are exported correctly, consumers should be forced into correct usage... in theory...

src/forms/calendarComponent.story.jsx
```diff
@@ -70,65 +74,56 @@ storiesOf('Forms/CalendarComponent', module)
...
 		return (
 			<div className="span--50">
 				<CalendarComponent
 					name="event_time"
 					label="In the year 3000"
 					value={date}
-					datepickerOptions={opts}
+					datepickerOptions={{
+						allowInput: true,
+						minDate: jodaDate,
```
mmcgahan: this needs update as well

src/forms/calendarComponent.story.jsx
```diff
@@ -70,65 +74,56 @@ storiesOf('Forms/CalendarComponent', module)
...
 			<div className="span--50">
 				<CalendarComponent
 					name="event_time"
 					label="Start at"
 					value={date}
 					error={'Woops, something went wrong.'}
-					datepickerOptions={opts}
+					datepickerOptions={{
+						allowInput: true,
+						minDate: jodaDate,
```
mmcgahan: and here

src/forms/calendarComponent.story.jsx
```diff
@@ -3,10 +3,13 @@ import CalendarComponent from './CalendarComponent';
 import { storiesOf } from '@storybook/react';
 import { decorateWithBasics, decorateWithInfo } from '../utils/decorators';
 import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { LocalDate, convert } from 'js-joda';
 
 import Section from '../layout/Section';
 import Chunk from '../layout/Chunk';
 
+const jodaDate = LocalDate.now();
```
mmcgahan: You probably already know, but just wanted to point out that if you do something like this in a unit test that has a snapshot, you'll want to mock any `.now` call in js-joda so that it returns a fixed value so that the snapshot doesn't change each time it's run - we've done that in a few places in mup-web, e.g. https://github.com/meetup/mup-web/blob/master/src/app/group/event/schedule/selectors/selectInitialValues.test.js#L19-L27

Just for future reference

PR #650, opened 2018-10-08
flowtype textarea component

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-757

#### Description
Converting `<Textarea />` to use flowtype


hsbacot:
@mmcgahan I absolutely agree. I was just discussing this with @jaclynj and was looking for how `<Textarea />` was used in `mup-web` to see if it justified keeping the internal state around.

PR #651, opened 2018-10-09
converting Textarea component tests to enzyme

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-744

#### Description
Updating `<Textarea />` tests to use enzyme.


mmcgahan:
BTW I generally prefer to avoid `beforeEach` behavior because it makes each individual unit test a little harder to understand, but I know there's a lot of it in MWC and other parts of the Meetup codebase

PR #653, opened 2018-10-12
Sds 758 refactor textarea to be stateless

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-758

#### Description
Removes state from `<Textarea />` and provides a backwards compatible component, `<UncontrolledTextarea />`

Do not merge yet.

### DIFF
--- a/src/forms/UncontrolledTextarea.jsx
+++ b/src/forms/UncontrolledTextarea.jsx
@@ -1,14 +1,10 @@
 // @flow
 import * as React from 'react';
 import Textarea from './Textarea';
+import type Props from './Textarea';
 
 type State = {
 	value: string,
-};
-
-type Props = {
-	onChange?: (e: SyntheticInputEvent<*>) => null,
-	value?: string,
 };
 
 /**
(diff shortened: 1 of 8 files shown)


src/forms/UncontrolledTextarea.jsx
```diff
@@ -0,0 +1,67 @@
+// @flow
+import * as React from 'react';
+import Textarea from './Textarea';
+
+type State = {
+	value: string,
+};
+
+type Props = {
```
mmcgahan: should these props match the props for `Textarea`?

hsbacot: That could be useful. I wasn't sure if I should explicitly declare them since this returns `<Textarea />` and they aren't used in this component aside from being passed through. Is it conventional to declare all possible props that might be used?

mmcgahan: all the props that are supported, yeah - consumer components shouldn't have to be aware of the implementation details, they just need to know what they can pass.

I recommend importing the `Props` type from `Textarea` and using it here.

PR #654, opened 2018-10-15
MG-2974: add knobs/flow to layout components

browne0:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MG-2974

#### Description
Added knobs and flow types to our storybook. Hopefully this will make our components more understandable to developers!

image

### DIFF
--- a/src/layout/Bounds.jsx
+++ b/src/layout/Bounds.jsx
@@ -23,15 +23,12 @@ type Props = {
 	isLoading?: boolean,
 };
 
-type State = {};
 /**
  * Design System Component: Provides `bounds` container for components
  * @module BoundsComponent
  */
 
-export class BoundsComponent extends React.PureComponent<Props, State> {
-	state = {};
-
+export class BoundsComponent extends React.PureComponent<Props> {
 	render() {
 		const {
 			children,
--- a/src/layout/Bounds.jsx
+++ b/src/layout/Bounds.jsx
@@ -1,4 +1,5 @@
-import PropTypes from 'prop-types';
+// @flow
+
 import React from 'react';
 import cx from 'classnames';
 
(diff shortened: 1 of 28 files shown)


src/layout/GridList.jsx
```diff
@@ -28,8 +65,8 @@ export class GridListComponent extends React.Component {
 			'gridList',
 			{
 				[`gridList--has${columns.default}`]: !!columns.default,
-				[`atMedium_gridList--has${columns.medium}`]: !!columns.medium,
-				[`atLarge_gridList--has${columns.large}`]: !!columns.large,
+				[`atMedium_gridList--has${columns.medium || 0}`]: !!columns.medium,
```
hsbacot: I'm a little confused here. How come 0 is needed when this should evaluate to `''` when the 0 would be added?

src/layout/Bounds.jsx
```diff
@@ -1,16 +1,37 @@
...
+
+type State = {};
 /**
  * Design System Component: Provides `bounds` container for components
  * @module BoundsComponent
  */
-export class BoundsComponent extends React.Component {
+
+export class BoundsComponent extends React.PureComponent<Props, State> {
+	state = {};
```
hsbacot: How come `state` is being added to the component?

src/layout/GridList.jsx
```diff
@@ -40,19 +77,17 @@ export class GridListComponent extends React.Component {
...
+				[`atLarge_${GRID_AUTOHEIGHT_CLASS}--has${columns.large ||
+					0}`]: !!columns.large,
 			},
 			className
 		);
 
 		const listItemClassNames = cx('gridList-item', {
 			['flex-item']: autoHeight,
-			[itemClassNames]: itemClassNames,
+			[itemClassNames || '']: Boolean(itemClassNames),
```
hsbacot: `|| ''` shouldn't be necessary here because when `itemClassNames` is an empty string or undefined this will return an empty string

hsbacot:
This is awesome! It's a lot of flowtyping and it should be a huge lift to the storybook experience as well. There were a few nitpick comments I left but they're minor compared to what this gets done.

src/layout/GridList.jsx
browne0: I should come back and fix this, it was to appease flow and seemed to work at the time.

src/layout/Bounds.jsx
browne0: OH i forgot to remove this, I was trying to clean up some warnings from storybook

src/layout/GridList.jsx
browne0: this was also to appease flow, i'll fix this

PR #655, opened 2018-10-18
[hackathon] make the event page accessible

burrata:
#### Related issues
Fixes everything

#### Description
screen shot 2018-10-18 at 1 50 21 pm

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -128,6 +128,7 @@ export class Textarea extends React.PureComponent<Props, State> {
 			required,
 			requiredText,
 			disableResize,
+			placeholder,
 			...other
 		} = this.props;
 
@@ -170,6 +171,7 @@ export class Textarea extends React.PureComponent<Props, State> {
 				{helperText && <div className={classNames.helperText}>{helperText}</div>}
 				<textarea
 					type="text"
+					aria-label={placeholder || name}
 					name={name}
 					required={required}
 					className={classNames.textarea}
--- a/src/interactive/InfoToggle.jsx
+++ b/src/interactive/InfoToggle.jsx
@@ -12,7 +12,7 @@ export const InfoTooltipTrigger = props => (
 			props.onClick(e);
 		}}
 	>
-		<span className="infoToggle-trigger align--center" role="img">
+		<span className="infoToggle-trigger align--center" role="img" alt="More info">
 			?
 		</span>
 	</Button>
--- a/src/interactive/InfoToggle.jsx
+++ b/src/interactive/InfoToggle.jsx
@@ -12,7 +12,11 @@ export const InfoTooltipTrigger = props => (
 			props.onClick(e);
 		}}
 	>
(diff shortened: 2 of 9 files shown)


src/interactive/InfoToggle.jsx
```diff
@@ -12,7 +12,11 @@ export const InfoTooltipTrigger = props => (
 			props.onClick(e);
 		}}
 	>
-		<span className="infoToggle-trigger align--center" role="img">
+		<span
```
hsbacot: Hey, awesome that y'all are working on this today. The accessibility functionality on this would ideally be pulled from props to allow more control of users that are changing it.

burrata: ohh, good idea. there are some other issues with this element anyway, so I'll adjust on my next change.

src/forms/Textarea.jsx
```diff
@@ -128,6 +128,7 @@ export class Textarea extends React.PureComponent<Props, State> {
 			required,
 			requiredText,
 			disableResize,
+			placeholder,
```
hsbacot: Make sure to pass `placeholder` through to `textarea` since it is being destructured from other

src/interactive/InfoToggle.jsx
```diff
@@ -23,6 +27,7 @@ const InfoToggle = ({
 	label,
 	tooltipId,
 	tooltipProps,
+	tooltipTriggerProps,
```
hsbacot: Not sure if adding prop objects is the best way to go forward with this because this type of configuration becomes a mess to maintain. It might be best to rethink the api of this component. `tooltipContent` could be children and we could support a `renderTrigger` prop that would take in the icon. My concern with that would be that it introduces a breaking change.

I'm thinking we should add a deprecation warning and refactor how the tooltip works for the next major release. Thoughts? @burrata @mmcgahan

mmcgahan: That sounds really good - I particularly like the idea of leveraging `children` for the tooltip content. If we instead implemented a component that was just a 'standard tooltip with a label", and then used _that_ as a basis for making an 'InfoToggle', we'd be in better shape.

Deprecation + refactor sounds good, unless this is used in a small number of places, in which case a hard breaking change wouldn't be too difficult to make sooner

hsbacot: I'm going to create a story in jira to cleanup this API. In the meantime we can pass the composed `InfoTooltipTrigger` with its props to the `tooltipProps` object without expanding the props passed to info toggle by overriding `trigger`.

ex:
```jsx
tooltipProps={{
  align: select('align', ['left', 'right', 'center'], 'left'),
  withClose: boolean('withClose', false),
  trigger: <InfoTooltipTrigger onClick={() => null} {...otherProps} />
}}
```
cc: @burrata @mmcgahan

burrata: I can add that!

hsbacot:
closing - feel free to re-open as needed

PR #657, opened 2018-10-22
MP-2084: Added new radio-group behavior

DA-Denis:
#### Related issues
Fixes https://meetup.atlassian.net/browse/MP-2084

#### Description
Added new radio-group behavior depending on isActive prop. By passing this attribute you can set radio buttons to disabled state to prevent rendering radio-buttons with default values.

### isActive is set to false
image

### When no isActive specified
image

### DIFF
--- a/src/forms/RadioButtonGroup.jsx
+++ b/src/forms/RadioButtonGroup.jsx
@@ -45,6 +45,7 @@ export default class RadioButtonGroup extends PureComponent {
 			className,
 			wrap,
 			justifyItems,
+			isActive,
 		} = this.props;
 
 		const switchDirectionAttr = switchDirection ? { switchDirection } : '';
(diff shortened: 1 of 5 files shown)


DA-Denis:
@hsbacot This could be an option. I we can manipulate RadioButtons instead of using RadioButtonGroup as well, however this doesn't feel right (just only my personal opinion) to handle each RadioButton separately when they are wrapped in RadioButtonGroup. If we don't want or need this functionality I'll close this PR and use `<RadioButton disabled />` feature instead. What do you think?

DA-Denis:
@hsbacot , @lipenco  Sorry for confusion, but disabled option works only on half. Since 'RadioButtons inidicator' is custom made, it reacts only on 'checked' option. RadioButton Indicator behavior:
```
<span className={classNames.fauxCheckboxClassNames}>
  {inputProps.checked && <span className="radio-indicator" />}
</span>
```

hsbacot:
@DA-Denis I'm not sure what you mean by only works on half. Could you proved a visual example?

Radio buttons can be checked and disabled at the same time.

DA-Denis:
@hsbacot Sure, so by 'works on half' I was meaning that `RadioButton disabled` will disable buttons from interaction but the visual indicator will stay there. So we get this:
image

Instead of:
image

hsbacot:
For future reference, this is being closed because the expected browser behavior allows for radio buttons to be selected and disabled at the same time.

PR #658, opened 2018-10-26
Sds 761 info toggle refactor

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-761
#### Description
Refactor `<InfoToggle />` to use `children` as component props, restructure the tooltip props taken into `<InfoToggle />` and moved the state of tooltips being opened when the trigger is clicked into the tooltip component.

### DIFF
--- a/src/interactive/InfoToggle.jsx
+++ b/src/interactive/InfoToggle.jsx
@@ -1,17 +1,22 @@
+// @flow
 import PropTypes from 'prop-types';
-import React from 'react';
+import * as React from 'react';
 
 import Button from '../forms/Button';
 import Tooltip from './Tooltip';
 
-export const InfoTooltipTrigger = props => (
-	<Button
-		reset
-		onClick={e => {
-			e.preventDefault();
-			props.onClick(e);
-		}}
-	>
+type Props = React.ElementConfig<'div'> & {
+	className?: string,
+	/** The content that's rendered inside the tooltip's content bubble */
+	tooltipContent: React.Node,
+	/** The label rendered next to the Tooltip's trigger */
+	label?: string,
+	/** The unique identifier for the Tooltip */
+	tooltipId: string,
+};
+
+export const InfoTooltipTrigger = () => (
+	<Button reset>
 		<span className="infoToggle-trigger align--center" role="img">
 			?
 		</span>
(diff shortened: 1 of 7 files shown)


src/interactive/Tooltip.jsx
```diff
@@ -60,6 +60,10 @@ class Tooltip extends React.PureComponent {
 		}
 	}
 
+	handleClick = () => {
+		this.setState(prevState => ({ isActive: !prevState.isActive }));
```
mmcgahan: nitpick, but avoid calling the arg to the updater `prevState`, which is a bit misleading (I actually made the update to the React docs in [this PR](https://github.com/reactjs/reactjs.org/pull/1155))

src/interactive/Tooltip.jsx
```diff
@@ -203,6 +209,9 @@ Tooltip.propTypes = {
 	trigger: PropTypes.element.isRequired,
 
 	/** The content that's rendered inside the tooltip */
+	children: PropTypes.element,
+
+	/** The content that's rendered inside the tooltip, to be deprecated in favor of children */
```
mmcgahan: essentially, this _is_ deprecated, not _to be_ deprecated, right?

src/interactive/Tooltip.jsx
```diff
@@ -178,7 +184,7 @@ class Tooltip extends React.PureComponent {
 											reset
 										/>
 									)}
-									{content}
+									{content || children}
```
mmcgahan: nitpick should this be reversed to make it clear that the deprecated prop is the fallback? `children || content`

src/interactive/Tooltip.jsx
```diff
@@ -203,6 +209,9 @@ Tooltip.propTypes = {
 	trigger: PropTypes.element.isRequired,
 
 	/** The content that's rendered inside the tooltip */
+	children: PropTypes.element,
```
mmcgahan: I believe `element` enforces `children` to be a single element - is that the intent? I don't think there's a reason it can't be normal children, in which case I don't think you need to define it as a proptype

hsbacot: No, there's no requirement to have one child.

src/interactive/InfoToggle.jsx
```diff
@@ -20,22 +24,25 @@ export const InfoTooltipTrigger = props => (
...
 		<div className={className} {...other}>
 			<span className="infoToggle-label">{label}</span>
-			<Tooltip
-				id={tooltipId}
-				trigger={<InfoTooltipTrigger onClick={onClick} />}
-				content={tooltipContent}
-				{...tooltipProps}
-			/>
+			<Tooltip trigger={<InfoTooltipTrigger />} {...tooltipOptions}>
+				{tooltipContent || children}
```
mmcgahan: looks like the interface is ambiguous here - do we prefer `children` over `tooltipContent`? Are we deprecating `tooltipContent` as well?

hsbacot: We prefer children here as well

PR #659, opened 2018-11-14
CSS as exported asset

hsbacot:
+ added relative importer to resolve styles from node modules
+ compiled css in new assets directory
+ build:css command added
+ main.scss created to have a 1 to 1 match of files compiled to main.css

#### Related issues
Fixes https://meetup.atlassian.net/browse/sds-773

#### Description
Exposes a compiled css file alongside the sass export in `assets`

### DIFF
--- a/util/sassRelativeImporter.js
+++ b/util/sassRelativeImporter.js
@@ -1,8 +1,12 @@
+// node-sass importer to resolve path names locally of imports from
+// node_modules the same way css-loader does with ~ as a prefix
+// used in build:css to build css from sass
+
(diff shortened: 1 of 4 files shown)


util/sassRelativeImporter.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path');
+
+module.exports = function(url, prev, done) {
```
mmcgahan: _definitely_ need to document this magic - only people familiar with webpack/css-loader will have a clue what the intent is

PR #662, opened 2018-12-11
Shrink largest avatar size from --responsiveMedia-xxl to --space-6

fisxoj:
#### Related issues
https://meetup.atlassian.net/browse/MG-3930

#### Description
This shrinks avatars from 150px at the largest media query to 120px there, and maintains 120px at the smallest.


fisxoj:
I've confirmed with John that that's the desired outcome for now.

PR #666, opened 2019-03-08
Add iconShape to checkbox

juanbiberretta:
#### Related issues
Fixes https://meetup.atlassian.net/browse/OS-260

#### Description
- An an optional `iconShape` prop to Checkbox component. 
- Defaults to `check` shape.

### DIFF
--- a/src/__snapshots__/footer.test.jsx.snap
+++ b/src/__snapshots__/footer.test.jsx.snap
@@ -207,7 +207,7 @@ exports[`Footer exists 1`] = `
           items={
             Array [
               <span>
-                © Meetup 2018
+                © Meetup 2019
               </span>,
               <span>
                 Mock legal copy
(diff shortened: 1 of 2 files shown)


src/__snapshots__/footer.test.jsx.snap
```diff
@@ -207,7 +207,7 @@ exports[`Footer exists 1`] = `
           items={
             Array [
               <span>
-                © Meetup 2018
+                © Meetup 2019
```
juanbiberretta: lol this snapshot was failing.. thought I'd just update it

muptravis: I thiiink we should just change the snapshot.. see here https://meetuphq.slack.com/archives/C099CDFJS/p1546441880004900

muptravis: ^ this is jackie hi

jaclynj: https://github.com/meetup/meetup-web-components/blob/master/src/Footer.jsx#L184 <- this is why it was failing. I think leave it for now and we'll sort it out

PR #667, opened 2019-03-08
Wp 1053 update deploy

mmcgahan:
Revising the deploy routine to make it more reliable and informative


mmcgahan:
looks good to @jaclynj

PR #668, opened 2019-03-12
OG-338: add draft functionality to ProfileDropdown

v-gtsolov:
**Task:** https://meetup.atlassian.net/browse/OG-338

**Description**

This task is specifically for mup-web project, we have separate task for chapstick (https://meetup.atlassian.net/secure/RapidBoard.jspa?rapidView=384&projectKey=OG&view=planning&selectedIssue=OG-333)

Given:
User creates draft on new group

When:
User clicks on the user menu dropdown

Then:
The user should see the draft link of that group into that dropdown user menu

### DIFF
--- a/src/navigation/components/groupDraftItem/GroupDraftItem.jsx
+++ b/src/navigation/components/groupDraftItem/GroupDraftItem.jsx
@@ -8,13 +8,13 @@ export const PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS = 'profileDropdown-draft-group'
 export const GroupDraftItem = ({ groupDraft }) => {
 	return (
 		<li
-			key={groupDraft.urlname}
+			key={groupDraft.editLink}
 			className={cx(PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS, 'list-item')}
 		>
 			<p>{groupDraft.name}</p>
 			<div>
 				<span>{groupDraft.status}</span>
-				<a className="draftprofiledropdown" href={groupDraft.urlname}>
+				<a className="draftprofiledropdown" href={groupDraft.editLink}>
 					<strong>{groupDraft.actionTitle}</strong>{' '}
 					<Icon className="text--blue" shape="arrow-right" size="xs" />
 				</a>
@@ -27,9 +27,9 @@ export default GroupDraftItem;
 
 GroupDraftItem.propTypes = {
 	groupDraft: PropTypes.shape({
-		urlname: PropTypes.string,
+		editLink: PropTypes.string,
 		name: PropTypes.string,
-		status: PropTypes.string, // translated 'In progress' string
+		status: PropTypes.string,
 		actionTitle: PropTypes.string,
 	}),
 };
--- a/src/navigation/components/groupDraftItem/GroupDraftItem.jsx
+++ b/src/navigation/components/groupDraftItem/GroupDraftItem.jsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import Icon from '../../../media/Icon';
+
+export const PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS = 'profileDropdown-draft-group';
+
+export const GroupDraftItem = ({ groupDraft }) => {
+	return (
+		<li
+			key={groupDraft.urlname}
+			className={cx(PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS, 'list-item')}
+		>
+			<p>{groupDraft.name}</p>
+			<div>
+				<span>{groupDraft.status}</span>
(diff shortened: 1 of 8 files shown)


src/navigation/components/profile/ProfileDropdown.jsx
```diff
@@ -23,10 +24,27 @@ export const ProfileDropdownComponent = ({
 	profile,
 	logout,
 	groups,
+	groupDraft,
 	groupHome,
 	allGroupsLabel,
 	allGroupsLink,
 }) => {
+	const groupDraftsContent = groupDraft => (
```
mmcgahan: Instead of redefining this component on every render, prefer making it a pure function that takes `(groupDraft, groupHome)` args.

Even better, make it a function component

```jsx
const GroupDraftItem = ({ groupDraft, groupHome }) => (
  <li>...</li>
)

// ...

<ul className="list">
  {groupDraft && <GroupDraftItem groupDraft={groupDraft} groupHome={groupHome} />
  {groupsContent}
</ul>
```

That will allow you to encapsulate the logic a little better, potentially test the standalone component (if you want), and avoid creating variables that might not be needed on every render.

src/navigation/components/profile/ProfileDropdown.jsx
```diff
@@ -23,10 +24,27 @@ export const ProfileDropdownComponent = ({
...
+			key={groupDraft.urlname}
+			className={cx(PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS, 'list-item')}
+		>
+			<p>{groupDraft.name}</p>
+			<div>
+				<span>{groupDraft.status}</span>
+				<a href={groupHome(groupDraft.urlname)}>
+					<strong>{groupDraft.actionTitle}</strong>{' '}
+				</a>
+				<Icon className="text--blue" shape="arrow-right" size="xs" />
```
hsbacot: Should the arrow be clickable as well here?

v-gtsolov: hmm, that's a good question, I will ask Kara about that

v-gtsolov:
@hsbacot @mmcgahan I made a few changes there, you can check them out again

src/navigation/components/groupDraftItem/GroupDraftItem.jsx
```diff
@@ -0,0 +1,35 @@
...
+export const GroupDraftItem = ({ groupDraft }) => {
+	return (
+		<li
+			key={groupDraft.urlname}
+			className={cx(PROFILE_GROUP_DRAFT_LIST_ITEM_CLASS, 'list-item')}
+		>
+			<p>{groupDraft.name}</p>
+			<div>
+				<span>{groupDraft.status}</span>
+				<a className="draftprofiledropdown" href={groupDraft.urlname}>
```
mmcgahan: If `groupDraft.urlname` is actually the group `urlname`, I think this will end up being something like `href='my-group-name'`, which won't always navigate to the right place. It's sort of tricky because ideally we would generate the link `href` using the user's locale (for something like `/fr-FR/my-group-name`) and we would avoid doing a full server render, but that's only possible from inside the applications with `<Link ...>`

The href should maybe be

```jsx
href={`/${groupDraft.urlname`}
```

as a compromise? The app server will perform the necessary language-aware redirects for that request

src/navigation/components/groupDraftItem/GroupDraftItem.test.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import GroupDraftItem from './GroupDraftItem';
+
+const MOCK_PROPS = {
+	groupHome: jest.fn(),
+	groupDraft: {
+		urlname: 'meetup.com/groups',
```
mmcgahan: I don't _think_ `urlname` will generally have the domain included - probably better to use a more accurate example.

src/navigation/nav.story.jsx
```diff
@@ -142,6 +142,27 @@ storiesOf('Site Chrome/Nav', module)
 			/>
 		);
 	})
+	.add('authenticated but with draft group at the top', () => {
+		const profile = {
+			...navItems.profile,
+			groupDraft: {
+				urlname: '/create',
```
mmcgahan: same deal here - `urlname` doesn't generally have a leading slash

src/navigation/components/groupDraftItem/GroupDraftItem.jsx
v-gtsolov: actually will only redirect to `/create` all the time, probably to change it to draftUrlName?

mmcgahan: ah, ok. yeah `draftUrlname` might be a good name for it, although if it's expected to be a url path that includes a leading slash, `draftUrlpath` might be better

PR #671, opened 2019-03-13
Swarm Migration: Checkbox

sadafie:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-872

#### Description

- Verified that the stories in storybook all work correctly.
- Removed props: `labelClassnames`, `onFocus`, `onBlur`. Note: `onFocus` and `onBlur` are not being used at all, so removal is safe.
- Open Questions:
1. Should `name` and `value` be passed down into the swarm Checkbox? Currently the swarm checkbox does not support / does nothing with `name` and `value` props.
2. Should we add support for `iconShape` to Checkbox? I believe it's in a more current design mock for checkbox but is already being used in mup-web. Some scope creep, but might be easy to just add in.


hsbacot:
@sadafieThis looks great so far.
1. yes, we should add support for name and value to `<Checkbox />`
2. No, we shouldn't support icon shape at the moment. That may change after conversations with design.

src/forms/Checkbox.jsx
```diff
@@ -41,103 +32,51 @@ class Checkbox extends React.PureComponent {
...
+			</SwarmCheckbox>
+		) : (
+			<SwarmCheckbox
+				checked={stateChecked}
+				label={label}
+				id={elId}
+				disabled={disabled}
+				onChange={this.onChange}
+				{...other}
+			/>
```
jaclynj: What about DRYing this up a bit so that if someone needs to change a prop they do it in once place and don't forget to do it in both (like the `checked` prop bug we fixed when we paired)

Something like this instead:

```javascript
		const swarmCheckboxProps = {
				label,
				disabled,
				checked: stateChecked,
				id: elId,
				onChange: this.onChange
		}

		return children ? (
			<SwarmCheckbox
				{...swarmCheckboxProps}
				{...other}
			>
				{children}
			</SwarmCheckbox>
		) : (
			<SwarmCheckbox
				{...swarmCheckboxProps}
				{...other}
			/>
```

src/forms/Checkbox.jsx
```diff
@@ -1,33 +1,24 @@
...
 /**
  * @module Checkbox
  */
 class Checkbox extends React.PureComponent {
 	constructor(props) {
 		super(props);
-		const state = {
-			focused: false,
+
+		this.state = {
```
mmcgahan: can we use class properties in MWC? If so, we can avoid the constructor

```js
class Checkbox extends React.PureComponent {
  state = this.props.checked || false;
  onChange = e => ...; // moved up from line 36
  ///...
```

src/forms/Checkbox.jsx
mmcgahan: this looks odd - can't we just pass the `children` into `SwarmCheckbox` regardless of whether it's falsey? In that case, we don't actually need the value of `children` and can remove it from the destructuring assignment and just let it pass through along with `other`, which would make this a lot simpler

hsbacot: Agreed, this could be simpler. We could omit a few of the props from the destructuring in render and allow them to pass through in other winding up with something like 
```jsx
return (
			<SwarmCheckbox
				checked={this.getChecked()}
				id={elId}
				disabled={disabled}
				onChange={this.onChange}
				{...other}
			/>
		);
```

src/forms/Checkbox.jsx
hsbacot: They should be supported in mwc

src/forms/Checkbox.jsx
```diff
@@ -1,33 +1,24 @@
...
 import React from 'react';
-import cx from 'classnames';
-import Flex from '../layout/Flex';
-import FlexItem from '../layout/FlexItem';
-import Icon from '../media/Icon';
-
-export const FAUX_TOGGLE_CLASS = 'fauxToggle';
-export const FOCUSED_CHECKBOX_CLASS = 'focused';
-export const DISABLED_CHECKBOX_CLASS = 'disabled';
+import { default as SwarmCheckbox } from '@meetup/swarm-components/lib/Checkbox';
```
hsbacot: These are exported from the root level as well so this could also be expressed as 
`import { Checkbox as SwarmCheckbox } from '@meetup/swarm-components'`

sergeychernyshev: That actually didn't work for me (pre-commit hook failed) in Select migration so I copied what Sadaf did here

hsbacot: @sergeychernyshev let's pair to replicate this. I'm curious what the specific error was

sadafie:
Swarm changes depend on : https://github.com/meetup/swarm-ui/pull/21

src/forms/Checkbox.jsx
sadafie: Does passing along `children` as part of `other` actually work? I tried to look this up but couldn't find anything in React docs or otherwise about it.

src/forms/Checkbox.jsx
sadafie: I guess the question is, does it matter? :) They're already classes and our goal is eventually to deprecate meetup-web-components entirely. What would this gain us?

src/forms/Checkbox.jsx
hsbacot: Yes, it's treated the same as other props in react

sadafie:
Correction: https://github.com/meetup/swarm-ui/pull/22

PR #673, opened 2019-03-14
Basic TextInput migrated, still need to replicate a few props.

sergeychernyshev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-875

#### Description
Updated `TextInput` component in `meetup-web-components` to use new `TextInput` component from `swarm-components`.

DoD:
* removed className, labelClassName, selectClassName props
* added flow type and documented optional id attribute (was used before)
* updated required (boolean or string) prop handling to match new requiredText (string) API
* checked Storybook examples to make sure they still function properly
  * [x] type "tel"
  * [x] default
  * [x] with value
  * [x] disabled
  * [x] error state
  * [x] error state as element _(actually works, storybook imported wrong component)_
  * [x] with helper text
  * [x] required
  * [x] search

Outstanding use-cases:
  * [ ] with icon
    * How do we support these elements? Do we have a helper for this?
  * [ ] with char counter if you provide maxLength
    * Do we even need this, need final decision
  * [ ] has a pattern for min length
    * `pattern` prop is passed through as before but it doesn't seem to affect anything (now or before, is this a bug or we are not testing it properly
  * `withErrorList` seems to be styled differently now (just bulleted list instead of colorized), is this the new styling

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -1,17 +1,15 @@
 // @flow
 import PropTypes from 'prop-types';
 import React from 'react';
-import cx from 'classnames';
-import Icon from '../media/Icon';
-import { MEDIA_SIZES } from '../utils/designConstants';
-import CharCounter from './CharCounter';
+
+import { TextInput as SwarmTextInput } from '@meetup/swarm-components/lib/TextInput';
+
 import withErrorList from '../utils/components/withErrorList';
 
 export const FIELD_WITH_ICON_CLASS = 'field--withIcon';
 
 type Props = {
 	children?: React$Node,
-	className?: string,
 	disabled?: boolean,
 	error?: string | React$Node,
 	helperText?: string | React$Node,
@@ -20,7 +18,6 @@ type Props = {
 	id?: string,
 	isSearch?: boolean,
 	label?: string | React$Node,
-	labelClassName?: string,
 	maxLength?: number,
 	name: string,
 	onChange: string => void,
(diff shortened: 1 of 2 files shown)


src/forms/TextInput.jsx
```diff
@@ -123,42 +81,32 @@ export const TextInput = (props: Props): React$Element<*> => {
...
 					{label}
+					{required ? <span> {requiredText}</span> : ''}
 				</label>
 			)}
-			{helperText && <div className={classNames.helperText}>{helperText}</div>}
-			<div style={{ position: 'relative' }}>
-				<input
-					type={isSearch ? 'search' : 'text'}
+
+			{helperText && <p data-swarm-select-helper-text="1">{helperText}</p>}
```
hsbacot: What's the purpose of the data attribute here?

src/forms/TextInput.jsx
```diff
@@ -123,42 +81,32 @@ export const TextInput = (props: Props): React$Element<*> => {
...
+
+			<div data-swarm-select-wrapper="1">
```
hsbacot: & here

hsbacot: Are these to be replaced with `<FieldLabel />` and `<FieldHelper />`?

sergeychernyshev: Yep, the purpose is to have something generic for styling labels and helper text.

src/forms/TextInput.jsx
```diff
@@ -123,42 +81,32 @@ export const TextInput = (props: Props): React$Element<*> => {
...
+		<div>
 			{label && (
-				<label
-					className={classNames.label}
-					htmlFor={id}
-					data-requiredtext={required && requiredText}
-				>
+				<label htmlFor={id || name}>
 					{label}
+					{required ? <span> {requiredText}</span> : ''}
```
hsbacot: Could be shortened to `{required && (<span>{requiredText}</span>)}

sergeychernyshev: It needs that extra space, but yes, I agree `&&` probably better.

PR #674, opened 2019-03-14
updating mwc to use the button component from the swarm-components

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-871

#### Description
Integrates Button from swarm components into meetup-web-components

### DIFF
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -1,13 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import cx from 'classnames';
-import Flex from '../layout/Flex';
-import FlexItem from '../layout/FlexItem';
 
-export const BUTTON_CLASS = 'button';
-export const BUTTON_ICON_WRAPPER_CLASS = 'button-icon-wrapper';
-export const BUTTON_LABEL_CLASS = 'button-label';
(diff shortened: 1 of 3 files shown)


src/forms/Button.jsx
```diff
@@ -1,105 +1,42 @@
...
-		const Component = component;
+		// checking for the icon component signature if passed as icon={<Icon shape="search"/>} and grabbing the shape prop
+		// since this functionality is built into button in order to enforce strict sizes on button icons
+		const props = { icon: icon && icon.props && icon.props.shape };
 
-		if (disabled) {
-			delete other.onClick;
+		if (component !== 'button') {
+			console.warn(
+				'All components are button elements. Future iterations will support links as independent components and the logic will be handled here'
```
jaclynj:
```suggestion
				'All Swarm UI v2 components are button elements. Future iterations will support links as independent components and the logic will be handled here'
```

sadafie: To continue off of Jackie's comment -- how do we want to handle `LinkButton`? There's a PR in swarm UI. I assume we want to continue supporting anchor tag buttons, so this would would need to be updated to return `LinkButton` or `Button`.

hsbacot: We should circle back to this and update it use the `LinkButton`

PR #675, opened 2019-03-15
OG-338: fixing groupDraft position in object

v-gtsolov:
Task: https://meetup.atlassian.net/browse/OG-338

This is bug fixing for some of the task OG-338 and the previous PR for this task was: https://github.com/meetup/meetup-web-components/pull/668

Now we have to change the statement for showing the groups listing, because we have to catch the case when we have no groups listed, but we have a draft group.

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -140,6 +140,7 @@ export class Nav extends React.Component {
 			messages,
 			notifications,
 			groups,
+			groupDraft,
 			profile,
 			updatesLabel,
 			logo,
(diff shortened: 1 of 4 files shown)


src/navigation/Nav.jsx
```diff
@@ -325,7 +327,7 @@ export class Nav extends React.Component {
 				),
 				onClickAction:
 					media.isAtMediumUp && !isGroupsLoaded
-						? profile.getSelfGroupsQuery
+						? (profile.getSelfGroupsQuery, profile.getSelfGroupDraftQuery)
```
mmcgahan: this syntax doesn't look right - should only be one function, not two inside parens?

PR #680, opened 2019-03-19
SDS-877 migrate TogglePill to swarm ui

sergeychernyshev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-877

#### Description
Migrated `TogglePill` over to Swarm UI version
- Removed various class name props
- Removed deprecated `useRadio`, `small` and `topic` props
- Removed deprecated internal state and reliance on special `ToggleWrapper` component
- Updated story file to remove deprecated use-cases


src/forms/TogglePill.jsx
```diff
@@ -104,15 +33,6 @@ TogglePill.propTypes = {
 
 	/** Whether the pill is toggled on/selected */
 	isActive: PropTypes.bool,
-
-	/** Whether to use input[type="radio"] to limit the selection to one toggle pill in a group */
```
sadafie: Are these props all removed because they are not used by any of the consumers?

sergeychernyshev: Yes, we are just dropping support for these and these props and internal state management.

PR #681, opened 2019-03-19
a few tweaks, one bugfix ('test' children)

mmcgahan:
a couple small tweaks to clean up the mgiration branch

### DIFF
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -6,35 +6,30 @@ import { Button as SwarmButton } from '@meetup/swarm-components';
 /**
  * @module Button
  */
-class Button extends React.PureComponent {
-	render() {
-		const {
-			// removing reset & fullWidth props from other to prevent an invalid boolean being passed to <button>
-			reset, // eslint-disable-line no-unused-vars
-			fullWidth, // eslint-disable-line no-unused-vars
-			icon,
-			hasHoverShadow, // eslint-disable-line no-unused-vars
-			component,
-			...other
-		} = this.props;
+const Button = props => {
+	const {
+		// removing reset & fullWidth props from other to prevent an invalid boolean being passed to <button>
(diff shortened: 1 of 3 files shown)


src/forms/Button.jsx
```diff
@@ -6,35 +6,30 @@ import { Button as SwarmButton } from '@meetup/swarm-components';
...
-
-		if (component !== 'button') {
-			console.warn(
-				'All Swarm UI v2 Button components are button elements. Future iterations will support links as independent components and the logic will be handled here'
-			);
-		}
-
-		return (
-			<SwarmButton {...props} {...other}>
-				Test
```
mmcgahan: this is a bug

sadafie:
Aside from the bug fix, why these changes? I believe in another PR, we agreed to keep the class structure and a few other things to minimize the set of changes in MWC. 

I'm fine with the changes in general, but I want to understand why you felt the need to for these changes so that if it is important we are all agreeing to it and applying the same approaches everywhere, including future components.

src/forms/Select.jsx
```diff
@@ -45,7 +42,7 @@ export const SelectInput = (props: Props) => {
 			name={name}
 			error={error}
 			helperText={helperText}
-			{...requiredProps}
+			requiredText={requiredText}
```
sergeychernyshev: Looks like you're removing the pattern I was using (maybe due to my misunderstanding of how props work).

What happens if require is `false`? will `requiredText` be set to false? how will it affect the fact that `requiredText` is only [supposed to be a string](https://github.com/meetup/swarm-ui/blob/master/packages/swarm-components/src/Select.jsx#L33)?

sadafie:
@mmcgahan This PR is outdated - should it be updated or closed?

mmcgahan:
closing

PR #682, opened 2019-03-19
SDS-878 Migrated ToggleSwitch to use Swarm UI Toggle

sergeychernyshev:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-878

#### Description
Updated `ToggleSwitch` to use Swarm UI's `Toggle` component.

- Removed class name props
- Removed deprecated `onCallback` and `offCallback` props
- Removed deprecated internal state and reliance on special `ToggleWrapper` component
- Updated story file to remove deprecated use-cases

Outstanding Question:
- We'll need `<FieldLabel>` here as well.


mmcgahan:
Does this need a 'do not merge' or 'prelim' label because of outstanding question? Otherwise, LGTM

sergeychernyshev:
Don't know, we merged the TextInput in (they don't go into master, but a migration branch)

@hsbacot, how are we planning to address the labels? separate PR for all elements or should we keep PRs open?

hsbacot:
@sergeychernyshev merging into mwc without `FieldLabel` is ok. The main concern is having a label prop in what is consumed versus having the label composed with the fields.

src/forms/ToggleSwitch.jsx
```diff
@@ -1,105 +1,28 @@
...
-		} = this.props;
-
-		const labelId = labelledBy || `label-${name}`;
+		const { isActive, label, labelledBy, id, name, ...other } = this.props;
 
-		const getActiveIconColor = isActive =>
-			isActive ? C_TEXT_PRIMARY : C_TEXT_SECONDARY;
-		const getColor = (disbled, isActive) =>
-			disabled ? C_TEXT_HINT : getActiveIconColor(isActive);
+		const labelProps = labelledBy ? { ariaLabelledby: labelledBy } : {};
```
hsbacot: The `araiLabelledby` prop should be `'aria-labelledby'` since all `aria-*` attributes are fully supported in jsx.

hsbacot: This also drops support for the `aria-labelledby` to be set when `labelledBy` is not passed as a prop

sergeychernyshev: Support for native `aria-labeledby` prop should still be there though - it will get applier through `{...other}` spread.

sergeychernyshev:
Actual style bug was fixed by upgrading `swarm-styles` to 0.1.7 (latest version) which is not on main migration branch yet.

sadafie:
Test updates looked good. Dont worry about the *other* failing tests. Merge away!

PR #683, opened 2019-03-21
Adding new class to Modal so we can do style overrides for mobile

eilinora:
#### Description
We need to be able to do some style changes to the inner modal content and there isn't a way to do this currently. I've added a way to pass a class to that inner content.

### DIFF
--- a/src/interactive/Modal.jsx
+++ b/src/interactive/Modal.jsx
@@ -130,6 +130,7 @@ export class ModalComponent extends React.Component {
 	render() {
 		const {
 			className,
+			modalClassName,
 			children,
 			fullscreen,
 			fixed,
@@ -151,7 +152,7 @@ export class ModalComponent extends React.Component {
 
 		const classNames = cx(className, 'modal');
 
-		const modalClasses = cx('view view--modal', {
+		const modalClasses = cx('view view--modal', modalClassName, {
 			'view--modalFull': fullscreen,
 			'view--modalFixed': fixed && !this.state.isMobileSize,
 			'view--modalSnap': !fullscreen,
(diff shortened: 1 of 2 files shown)


src/interactive/Modal.jsx
```diff
@@ -130,6 +130,7 @@ export class ModalComponent extends React.Component {
 	render() {
 		const {
 			className,
+			modalClassName,
```
hsbacot: Were you able to try passing in a class to `className` and then target the modal? It should be supported by creating a class at the top level.

```.css
.customClassFromConsumer .view--modal {
  ...
}
```

and then passing the `.customClassFromConsumer` to the `className` prop. ex:
```jsx
<Modal className='customClassFromConsumer' {...props} />
```

or if you're using css modules as 

```jsx
import styleModule from './styles'

<Modal className={styleModule.customClassFromConsumer} {...props} />
```

eilinora: I tried that but it doesn't work, because the `view--modal` class also gets hashed

eilinora: The css winds up looking like this, so it wont target view--modal
Screen Shot 2019-03-21 at 12 13 08 PM

hsbacot: I see, it looks like we're hashing the nested classes too. In that case it should work by indicating that view module is global. ex:
```scss
.customClassFromConsumer {
  :global {
    .view--modal {
      ...
    }
  }
}
```

hsbacot: It's better explained in the css loader scope docs here -> https://github.com/webpack-contrib/css-loader#scope

eilinora:
Apparently this isn't needed 🙌 for `:global`... closing

PR #684, opened 2019-03-22
fix checkbox unit test

sadafie:
#### Related issues
Fixes https://meetup.atlassian.net/browse/<XXX-870>

#### Description

- remove class based unit tests
- snapshots DID NOT need updating (weird)
- fix a bug in the Checkbox component that was causing failing tests

### DIFF
--- a/src/forms/Checkbox.jsx
+++ b/src/forms/Checkbox.jsx
@@ -47,6 +47,7 @@ class Checkbox extends React.PureComponent {
 		const {
 			checked, // eslint-disable-line no-unused-vars
 			controlled, // eslint-disable-line no-unused-vars
+			onChange, // eslint-disable-line no-unused-vars
 			id,
 			label,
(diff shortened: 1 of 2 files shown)


src/forms/Checkbox.jsx
```diff
@@ -47,6 +47,7 @@ class Checkbox extends React.PureComponent {
 		const {
 			checked, // eslint-disable-line no-unused-vars
 			controlled, // eslint-disable-line no-unused-vars
+			onChange, // eslint-disable-line no-unused-vars
```
sadafie: this was a bug in the component. We explicitly pass the `onChange` prop to the input here - https://github.com/meetup/meetup-web-components/pull/684/files#diff-4d4c311525b22a86cb63d1ba89a786b5R68. But it was also getting passed along in `other` and so overriding the onChange callback.

PR #686, opened 2019-03-22
Sds 912 radio button group swarm

sadafie:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-912

#### Description

Usage:

```
    <RadioButtonGroup ...>
         <RadioButton .... />
         <RadioButton .... />
         ...
      </RadioButtonGroup>
```

Alternatively, you could use any element as a child, i.e. this is valid:

```
    <RadioButtonGroup ...>
         <TogglePill ...>
         <TogglePill ...>
          ...
     </RadioButtonGroup>
```

and this is also valid:

```
    <RadioButtonGroup ...>
          <Button ....>
          <Button ....>
          ....
     </RadioButtonGroup>
```

### ToDO:

- Need to debug the case where `<TogglePill>` is used as child options as opposed to `RadioButton`

### DIFF
--- a/src/forms/RadioButtonGroup.jsx
+++ b/src/forms/RadioButtonGroup.jsx
@@ -8,46 +8,32 @@ export default class RadioButtonGroup extends PureComponent {
 	constructor(props) {
 		super(props);
 
-		const initialState = {
-			value: props.selectedValue,
+		this.state = {
+			selectedValue: props.selectedValue,
 		};
 
-		React.Children.forEach(props.children, radio => {
-			initialState[radio.props.value] = radio.props.value === props.selectedValue;
-		});
-
-		this.state = initialState;
-		this.onChange = this.onChange.bind(this);
+		this.handleChange = this.handleChange.bind(this);
 		this.isSelected = this.isSelected.bind(this);
 	}
 
 	componentWillReceiveProps(nextProps) {
 		if ('selectedValue' in nextProps) {
-			const newState = {};
-
-			Object.keys(this.state).forEach(stateKey => {
-				if (stateKey === 'value') {
-					newState.value = nextProps.selectedValue;
-				} else {
-					newState[stateKey] = stateKey === nextProps.selectedValue;
-				}
+			this.setState({
+				selectedValue: nextProps.selectedValue,
 			});
-
-			this.setState(newState);
 		}
 	}
 
-	onChange(e) {
+	handleChange(e) {
 		this.props.onChange && this.props.onChange(e);
 
-		// TODO: also update state of individual buttons
 		this.setState({
-			value: e.target.value,
+			selectedValue: e.target.value,
 		});
 	}
 
 	isSelected(radio) {
-		return radio.props.value === this.state.value;
+		return radio.props.value === this.state.selectedValue;
 	}
 
 	render() {
@@ -57,9 +43,9 @@ export default class RadioButtonGroup extends PureComponent {
 			<div className={className}>
 				{React.Children.map(children, radio =>
 					React.cloneElement(radio, {
-						onChange: this.onChange,
-						checked: this.isSelected(radio),
+						onChange: this.handleChange,
 						...radio.props,
+						checked: this.isSelected(radio),
 					})
 				)}
 			</div>
--- a/src/forms/RadioButtonGroup.jsx
+++ b/src/forms/RadioButtonGroup.jsx
@@ -1,5 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import Flex from '../layout/Flex';
+import FlexItem from '../layout/FlexItem';
 
 /**
  * @class RadioButtonGroup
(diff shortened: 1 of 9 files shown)


src/forms/RadioButtonGroup.jsx
```diff
@@ -1,73 +1,68 @@
...
+
+			Object.keys(this.state).forEach(stateKey => {
+				if (stateKey === 'value') {
+					newState.value = nextProps.selectedValue;
+				} else {
+					newState[stateKey] = stateKey === nextProps.selectedValue;
+				}
 			});
+
+			this.setState(newState);
```
sadafie: I don't particularly like this approach. Would have liked to use hooks. Also, I think that I should move all this logic to `onChange` and that having a `componentWillReceiveProps` method is completely unnecessary, but this is how the old implementation was doing it.

src/forms/RadioButtonGroup.jsx
```diff
@@ -1,73 +1,68 @@
...
  * @class RadioButtonGroup
  */
 export default class RadioButtonGroup extends PureComponent {
 	constructor(props) {
 		super(props);
-		this.state = {
-			selectedValue: props.selectedValue,
+
+		const initialState = {
+			value: props.selectedValue,
```
mmcgahan: there's some problematic stuff in how this component is written, but _some_ of the is problems were there in the initial implementation so I'm not sure how much we can 'fix'.

In general, it's [not a good idea to copy props into state](https://overreacted.io/writing-resilient-components/#principle-1-dont-stop-the-data-flow) - in this case, every time the parent re-renders, the current `state.value` value will be overwritten with `props.selectedValue`, which would cause problems.

The updates to the code also overwrite the `this.state[value]` map on every render, which probably isn't what is intended.

src/forms/RadioButtonGroup.jsx
```diff
@@ -1,73 +1,68 @@
...
 		};
-		this.handleChange = this.handleChange.bind(this);
+
+		React.Children.forEach(props.children, radio => {
+			initialState[radio.props.value] = radio.props.value === props.selectedValue;
```
mmcgahan: I might be missing it, but it doesn't look like this 'radio value' state is being used at all in rendering? I see it being written here and in `componentWillReceiveProps`, but never being read. Can you go back to just having `this.state.selectedValue`?

src/forms/RadioButtonGroup.jsx
mmcgahan: yeah this component should probably not be used at all given how re-rendering will reset its internal state. We probably need a backlog ticket to rewrite it correctly

src/forms/RadioButtonGroup.jsx
```diff
@@ -84,13 +60,6 @@ RadioButtonGroup.propTypes = {
 	/** Additional class name/s to add to the wrapper element */
 	className: PropTypes.string,
 
-	/** The axis the radio buttons will be layed out on */
-	direction: PropTypes.oneOf(['row', 'column']),
```
sadafie: @hsbacot I removed the `direction` prop because it relies on `Flex`. Is this correct? If not, what should be the correct behavior here?

src/forms/RadioButtonGroup.jsx
```diff
@@ -24,50 +23,27 @@ export default class RadioButtonGroup extends PureComponent {
...
-			switchDirection,
-			children,
-			name,
-			className,
-			wrap,
-			justifyItems,
-		} = this.props;
-
-		const switchDirectionAttr = switchDirection ? { switchDirection } : '';
+		const { children, className } = this.props;
```
sergeychernyshev: Do we need to explicitly use `className` here or `...other` would be better and would carry other props as well?
We don't seem to do anything special with `className` but just apply it to the outmost component, right?

mmcgahan: I guess technically that would change the original interface, which doesn't support arbitrary props, but I like the idea

src/forms/RadioButtonGroup.jsx
hsbacot: We need to support this since it used to determine the layout of the radio buttons in some places (and is actively used in mup-web). We could add support via css that is conditionally applied here or return the flex structure while in mwc since the flex shouldn't impact the styles.

sadafie: My only thought on this is that if at some point in time we want product teams to move from using MWC -> Swarm-Components, they're going to need to implement the CSS to create a column layout because we don't intend to provide this to them. So either 1) we could make this change now and write the css in the mup-web integration branch (not in this branch), or 2) we keep Flex but defer it to later for product teams. 

I can add Flex back in, but I just think this component needs a bit more planning after the first release. I'm also leaning toward creating a `RadioButtonGroup` component in `mup-web` that does state management + css/layout with data selectors because ultimately that's what's going to HAVE to happen when we remove MWC and are solely using Swarm-Components. We shouldn't end up in a state where every product team creates their own, new RadioButtonGroup component each time, which is what will happen unless we plan / influence it somehow.

src/forms/RadioButtonGroup.jsx
sadafie: I also think that using `..other` is better, but I'm going to leave it as is to limit the changes to the interface / hopefully fewer bugs introduced.

sadafie:
@hsbacot The only remaining task on this branch is to fix the use case where `RadioButtonGroup` has `TogglePill` as children. In the migration branch, I see we got rid of the `useRadio` attribute on `TogglePill`, and this breaks the `RadioButtonGroup` functionality for that case.

src/forms/RadioButtonGroup.jsx
```diff
@@ -24,46 +25,26 @@ export default class RadioButtonGroup extends PureComponent {
...
 		return (
-			<Flex
-				direction={direction}
-				{...switchDirectionAttr}
-				className={className}
-				wrap={wrap}
-				justifyItems={justifyItems}
-			>
-				{React.Children.map(children, option => (
+			<Flex direction={direction} className={className}>
```
sadafie: I intentionally removed the `wrap` and `justifyItems` props because:

1) They're not documented anywhere
2) Mup-web, pro-web, and MWC do not use those props.

src/forms/RadioButtonGroup.jsx
hsbacot: We should keep support for flex here, it will be around in some form for an indefinite amount of time. It is still being decided how spacing should work as swarm evolves.

hsbacot: Functionally we should document how to handle state for a radio group but we should not handle it because that would result in duplicate state being stored.

sadafie:
Paired with hunter on the last of these changes. Going to merge

PR #687, opened 2019-03-25
updating button tests

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-904

#### Description
Updating failing tests related to `<Button />`

### DIFF
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -23,7 +23,6 @@ class Button extends React.PureComponent {
 		const props = { icon: icon && icon.props && icon.props.shape };
 
 		if (component !== undefined && component !== 'button' && component !== 'a') {
-			console.log('****', component);
 			console.warn(
 				'All Swarm UI v2 Button components are button elements. Future iterations will support links as independent components and the logic will be handled here'
 			);
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -22,7 +22,8 @@ class Button extends React.PureComponent {
 		// since this functionality is built into button in order to enforce strict sizes on button icons
 		const props = { icon: icon && icon.props && icon.props.shape };
 
-		if (component !== 'button' || 'a') {
+		if (component !== undefined && component !== 'button' && component !== 'a') {
+			console.log('****', component);
 			console.warn(
 				'All Swarm UI v2 Button components are button elements. Future iterations will support links as independent components and the logic will be handled here'
 			);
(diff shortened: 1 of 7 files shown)


src/forms/button.test.jsx
```diff
@@ -18,83 +12,62 @@ describe('Button', () => {
 		});
 	});
 
-	it('applies variant classes for each variant prop', () => {
-		const variants = ['primary', 'fullWidth', 'small', 'bordered'];
+	it('has a default state', () => {
+		const button = mount(<Button>Click Me</Button>);
```
sadafie: Slightly more DRY approach would be to define a `renderComponent` function:

```
const renderComponent = (props = {}) => mount(<Button {...props}>Click Me</Button>);

describe('Button', () => {
...

   const button = renderComponent({primary:true});
})
```

src/forms/Button.jsx
```diff
@@ -22,17 +22,17 @@ class Button extends React.PureComponent {
 		// since this functionality is built into button in order to enforce strict sizes on button icons
 		const props = { icon: icon && icon.props && icon.props.shape };
 
-		if (component !== 'button') {
+		if (component !== 'button' || 'a') {
```
jaclynj: This logic isn't right. the second half of the statement is just 'a' which will always evaluate to true. I'm not sure what you actually want here.. maybe this `component !== 'button' && component !== 'a'`

jaclynj: I think you should also add a test to make sure this warning is logged only when you want it to.

src/forms/button.test.jsx
```diff
@@ -18,83 +12,62 @@ describe('Button', () => {
...
-		button.simulate('click');
-		expect(onClick).toHaveBeenCalled();
+	it('has a bordered state', () => {
+		const button = mount(<Button bordered>Click Me</Button>);
+		expect(
+			button.getDOMNode().attributes.getNamedItem('data-swarm-button').value
+		).toBe('bordered');
 	});
 
-	it('does not execute onClick when disabled', () => {
```
jaclynj: Should we also have this test? make sure not to call onclick when the button is disabled?

hsbacot: This is a test that predates the support being default in react 🦖

src/forms/button.test.jsx
hsbacot: I prefer a little repetition here to keep the react api in the test definition. This prevents looking around for the definition of the `renderComponent` when you want to change.

PR #690, opened 2019-03-27
Package.json updates

sadafie:
It's been awhile since we've upgraded packages and/or updates the package.json for MWC. I did a quick passthrough. Storybook runs successful. I'm going to use the beta version to test in mup-web and pro-web.


sadafie:
D'oh. Somehow the migration branch commits got into this. Will close and create a new one

PR #691, opened 2019-03-27
Updates to package.json

sadafie:
I noticed that the package.json for MWC hasn't gotten attention in some time. I upgraded package versions, removed unused dependencies, and so forth. Certain peerDependencies introduce unnecessary dependencies for pro-web. Overall, a bit of cleanup. Storybook runs correct and I just need to use the beta version of this PR to test in mup-web and pro-web.


sadafie:
Contains all the commits from the migration branch when it's not supposed to! Closing.. not sure why this happens.

PR #698, opened 2019-04-15
adding labels within mwc

hsbacot:
Adding label and helper text to select inputs in mwc using swarm label and helper components

### DIFF
--- a/src/forms/Select.jsx
+++ b/src/forms/Select.jsx
@@ -2,6 +2,7 @@
 import * as React from 'react';
 
 import { default as SwarmSelect } from '@meetup/swarm-components/lib/Select';
+import { FieldHelper, FieldLabel } from '@meetup/swarm-components';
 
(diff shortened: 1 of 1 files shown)


src/forms/Select.jsx
```diff
@@ -56,15 +57,18 @@ export class SelectInput extends React.PureComponent<Props> {
...
-				helperText={helperText}
-				{...requiredProps}
-				{...other}
-			/>
+			<React.Fragment>
+				{label && <FieldLabel>{label}</FieldLabel>}
+				{helperText && <FieldHelper>{label}</FieldHelper>}
+				<SwarmSelect
+					id={id || name}
+					label={label}
```
sadafie: We probably shouldn't be passing the label here too. It's a bit misleading / confusing. I can't remember if the `swarm-ui` select element expects a label prop, but if it does, we should remove that and clean this up.

PR #700, opened 2019-04-15
Bugfix select labels

hsbacot:
`htmlFor` specified on label
label removed from select component props
helper text text corrected

### DIFF
--- a/src/forms/Select.jsx
+++ b/src/forms/Select.jsx
@@ -56,13 +56,14 @@ export class SelectInput extends React.PureComponent<Props> {
 			? { requiredText: typeof required === 'string' ? required : '*' }
 			: {};
 
+		const fieldId = id || name;
+
 		return (
 			<React.Fragment>
-				{label && <FieldLabel>{label}</FieldLabel>}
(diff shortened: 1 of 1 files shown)


src/forms/Select.jsx
```diff
@@ -56,13 +56,14 @@ export class SelectInput extends React.PureComponent<Props> {
 			? { requiredText: typeof required === 'string' ? required : '*' }
 			: {};
 
+		const fieldId = id || name;
```
sadafie: This is not specific to this ticket, but I'm wondering if we should file a ticket to do this...

The `name` prop is not guaranteed to be unique, correct? In general, whenever I've seen the `id || name` construct in our code (it's all over), I've wondered if we should append a unique id to the name. Maybe something like:

```
const fieldId = id || `${name}-${getUniqueId()}`;

getUniqueId = () => Math.random().toString(36).substr(2,9);
```

Or something similar

hsbacot: Let's follow up with this because it likely happens in other places we link labels to inputs

PR #701, opened 2019-04-15
textarea migration

hsbacot:
#### Related issues
Fixes https://meetup.atlassian.net/browse/SDS-934

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -1,9 +1,13 @@
 // @flow
 import * as React from 'react';
-import cx from 'classnames';
 import autosize from 'autosize';
-import CharCounter from './CharCounter';
 import withErrorList from '../utils/components/withErrorList';
+
+import {
+	Textarea as SwarmTextarea,
+	FieldLabel,
+	FieldHelper,
+} from '@meetup/swarm-components';
 
 type Props = React.Element<HTMLTextAreaElement> & {
 	/** Adds an `id` attribute to the input, and associates it with the `<label />` */
@@ -113,9 +117,7 @@ export class Textarea extends React.PureComponent<Props, State> {
 			name,
 			value, // eslint-disable-line no-unused-vars
 			label,
-			labelClassName,
-			className,
-			error,
+			labelClassName, // eslint-disable-line no-unused-vars
 			rows,
 			style = {},
 			maxHeight,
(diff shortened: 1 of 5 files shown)


src/forms/Textarea.jsx
```diff
@@ -113,82 +117,47 @@ export class Textarea extends React.PureComponent<Props, State> {
 			name,
 			value, // eslint-disable-line no-unused-vars
 			label,
-			labelClassName,
-			className,
-			error,
+			labelClassName, // eslint-disable-line no-unused-vars
```
sadafie: I'd suggest removing labelClassName and any other unused props from the props type. Thinking back on this, during QA and during the intermediate phase of having MWC + Swarm UI, keeping the prop definitions up to date will help eliminate a lot of confusion.

jaclynj: I disagree.. If we remove this from being destructured but unused here, then it would get passed all the way through (here: https://github.com/meetup/swarm-ui/blob/master/packages/swarm-components/src/Textarea.jsx#L32) and we'd get errors about passing down labelClassName to the dom element. So if we do remove this, we will have to remove the prop everywhere in mup-web where the prop (or any other now-unused prop) is passed.

I think we should leave these formerly used props, and perhaps add a console warning about it.

src/forms/Textarea.jsx
```diff
@@ -113,82 +117,47 @@ export class Textarea extends React.PureComponent<Props, State> {
...
 		// Avoid passing maxLength as an HTML attribute
 		if (maxLength) delete other.maxLength;
 
 		return (
 			<div className="inputContainer">
-				{label && (
-					<label
-						className={classNames.label}
-						htmlFor={id}
-						data-requiredtext={required && requiredText}
```
jaclynj: We're no longer supporting this? What if the field is required?

src/forms/Textarea.jsx
```diff
@@ -113,82 +111,47 @@ export class Textarea extends React.PureComponent<Props, State> {
...
 					id={id}
 					value={this.state.value}
 					{...other}
 				/>
-				{maxLength && (
-					<CharCounter
-						maxLength={parseInt(maxLength, 10)}
-						valueLength={parseInt(this.state.value.length, 10)}
-					/>
-				)}
```
jaclynj: I remember you mentioned this, we don't have to support this for the first release either?

src/forms/Textarea.jsx
hsbacot: @jaclynj I had the same thoughts. @sadafie means to clean up the proptype definitions on line 18, but she wasn't able to comment on that line.

src/forms/Textarea.jsx
hsbacot: We don't have specs but it'll come up in vqa and we can address it then.

cc: @sadafie

src/forms/Textarea.jsx
jaclynj: ah got it

PR #703, opened 2019-04-15
destructuring required text

hsbacot:
Destructures required text to keep it from passing it to dom
For beta build

### DIFF
--- a/src/forms/Textarea.jsx
+++ b/src/forms/Textarea.jsx
@@ -122,6 +122,7 @@ export class Textarea extends React.PureComponent<Props, State> {
 			helperText,
 			required,
 			requiredText, // eslint-disable-line no-unused-vars
+			autosize, // eslint-disable-line no-unused-vars
 			disableResize,
 			...other
 		} = this.props;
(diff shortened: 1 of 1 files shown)


src/forms/Textarea.jsx
```diff
@@ -139,9 +120,6 @@ export class Textarea extends React.PureComponent<Props, State> {
 					required={required}
 					onChange={this.onChange}
 					rows={rows}
-					ref={textarea => {
```
sadafie: Hmmm, does removing the ref break expected behavior in mup-web? Not quite sure how we would know this, but I remember in some components ref's were manually talked in like this intentionally.

src/forms/Textarea.jsx
```diff
@@ -66,26 +65,6 @@ export class Textarea extends React.PureComponent<Props, State> {
 		return overrideValue(nextProps);
 	}
 
-	/**
-	 * Turns on autosize if requested
-	 * @return {undefined} side effect only
-	 */
-	componentDidMount() {
-		if (this.props.autosize) {
```
sadafie: Does removing autosize break mup-web behavior? I'm wondering if we should just leave it in but no pass through to swarm ui?

src/forms/Textarea.jsx
hsbacot: This ref is only used to as a way to pass a dom node to autosize, which calculates size of the texarea. It'll break the functionality but this should be added to the swarm-components version. Right now the refs wouldn't be forwarded since this is a swarm component and not the `textarea` node.

PR #706, opened 2019-05-08
Button support for react-router-dom <Link> component

sadafie:
#### Related issues
Fixes https://meetup.atlassian.net/browse/WP-1198

#### Description

- I'm applying swarm v2 data selectors directly to the `<Link>` component. This is somewhat messy, and is a duplication of code in `swarm-ui`, but I'm importing utilities from `swarm-ui` as much as possible.
- Visually, it looks "ok" in storybook. Could use another set of eyes.
- I added a new `<Link>` + icon state in storybook.
- I added a new unit test case for `<Link>` + icon as well.
- Tested locally in `mup-web-aws` and the functionality appears correct.

The more correct solution is to implement this in `swarm-ui`, but that introduces a `react-router-dom` dependency which we want to avoid, so implementing it here for now. This should be later replaced with a valid Swarm v2 solution.


src/forms/Button.jsx
```diff
@@ -1,8 +1,19 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import Link from 'react-router-dom/Link';
```
mmcgahan: New warning from `react-router-dom` v5.0.0 (upgrade coming soon) - this import will need to be

```js
import { Link } from 'react-router-dom';
```

I think that's a valid way to do it with v4 as well, so might as well set it up that way now

src/forms/Button.jsx
```diff
@@ -13,29 +24,69 @@ class Button extends React.PureComponent {
...
 			...other
 		} = this.props;
 
 		// checking for the icon component signature if passed as icon={<Icon shape="search"/>} and grabbing the shape prop
-		// since this functionality is built into button in order to enforce strict sizes on button icons
-		const props = { icon: icon && icon.props && icon.props.shape };
+		const iconShape = icon && icon.props && icon.props.shape;
+
+		// this.props.component is either a string or a Link component with a `component.name` prop
+		const componentName = component
```
mmcgahan: nested ternaries are hard to reason about, consider this option to make the logic explicit

```js
const componentName = component === Link ? 'Link' : component;
```

src/forms/Button.jsx
```diff
@@ -13,29 +24,69 @@ class Button extends React.PureComponent {
...
+									{children}
+									<Icon shape={iconShape} size="xs" color={color} />
+								</React.Fragment>
+							) : (
+								<React.Fragment>
+									<Icon shape={iconShape} size="xs" color={color} />
+									{children}
+								</React.Fragment>
+							)}
+						</span>
```
mmcgahan: consider moving this deeply-nested ternary into a pure function component that can be called like

```jsx
icon
  ? <ButtonIcon right={right} shape={iconShape} color={color}>{children}</ButtonIcon>
  : children
```

Alternatively, there are some cool ways of [using flexbox to reverse the layout order of elements](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction) without changing markup, and that would be a pretty slick solution that would making the layout rule affect the JSX

src/forms/Button.jsx
sadafie: `component === Link` unfortunately doesn't work. It'll always return false. Leaving as if for now - this should be a temporary solution anyhow. We need to add this support into swarm-ui and remove from here in v2.

src/forms/Button.jsx
sadafie: This is pretty much 100% copy/paste code from swarm-ui:

https://github.com/meetup/swarm-ui/blob/master/packages/swarm-components/src/Button.jsx#L107

But! I like this idea as an improvement for swarm-ui. 

As mentioned in this previous comment, this code should hopefully only be temporary and get replace by a `<SwarmButton component={Link}>` or something similar in a future iteration.

PR #714, opened 2019-06-03
Add link to Meetup Experiences to Desktop nav

carawarner:
🍐 with @brookeweil 

#### Related issues
https://www.pivotaltracker.com/story/show/166165107
https://github.com/meetup/mup-web/pull/5942

#### Description
See description on [mup-web PR](https://github.com/meetup/mup-web/pull/5942). 

#### Screenshot 
image

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -4911,7 +4911,7 @@ input.field--reset:active {
       font-weight: 600; } }
 
 .pill {
-  background: #5c5cff;
+  background: var(--color-viridian);
   border-radius: 8px;
   color: white;
   float: right;
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -4910,6 +4910,11 @@ input.field--reset:active {
       color: #00a2c7 !important;
       font-weight: 600; } }
 
+@media only screen and (min-width: 640px) {
+  .navItemLink--experiences .navItem-label {
+    color: #00a2c7 !important;
+    font-weight: 600; } }
+
 .counterBadge {
   background-color: #00d480;
   border: 3px solid white;
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -4910,10 +4910,16 @@ input.field--reset:active {
       color: #00a2c7 !important;
       font-weight: 600; } }
 
-@media only screen and (min-width: 640px) {
-  .navItemLink--experiences .navItem-label {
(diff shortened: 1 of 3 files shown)


assets/css/main.css
```diff
@@ -4910,6 +4910,11 @@ input.field--reset:active {
       color: #00a2c7 !important;
       font-weight: 600; } }
 
+@media only screen and (min-width: 640px) {
+  .navItemLink--experiences .navItem-label {
+    color: #5c5cff !important;
```
hsbacot: Do we specifically want this to be a color not in swarm?

carawarner: Hi @hsbacot, we got the 'okay' from design to proceed with using this distinctive purple color. Right now this is the only place it's used outside the experiences repos. If we need to use it in a second place, I'll work on getting it added to the official color constants in Swarm. That okay?

src/navigation/Nav.jsx
```diff
@@ -252,6 +253,15 @@ export class Nav extends React.Component {
 				),
 			},
 			media.isAtMediumUp && !self.is_pro_admin && createMeetupLink,
+			// If you want to use pill elsewhere, consider making a Pill
+			// component that takes a color prop.
+			media.isAtMediumUp && {
+				shrink: true,
+				linkTo: experiences.link,
```
sadafie: What happens if `experiences` is undefined? Do you need to update each usage of `Nav` in pro-web and mup-web to guarantee the existence of this?

assets/css/main.css
sadafie: Out of curiosity, does swarm already have a purple color? And if so, why are we designing outside of swarm colors?

src/navigation/Nav.jsx
carawarner: We had added a null check but hadn't committed it. Committed and pushed.

assets/css/main.css
carawarner: @sadafie do I have access to swarm-ui colors in `meetup-web-components`? I just tried to use `$C_veridian` and got an undefined variable error.

hsbacot: @carawarner It would be $color_viridian if importing the scss variables from `@meetup/swarm-constants` or `--color-viridian` if using the css variables (recommended)

PR #715, opened 2019-06-03
add click tracking to sign up modal

browne0:
#### Related issues
Fixes https://www.pivotaltracker.com/story/show/166418085
#### Description

**Why**
There are numerous ways to navigate to the Member Registration page (secure.meetup.com/register) and these ways have different variations that product, marketing & analysts need to be able to understand.

**As a** GA user
**I want** to see when a member's registration flow starts from the "Sign Up" button in the header or footer on the new guest homepage experience (meetup.com)
**So I can** visualize the flow a user experienced when registering or where they may have fallen off

**What is the desired behavior?**
Acceptance Criteria
Scenario:
Given a guest on (meetup.com [new experience])
When the guest clicks sign up in header or footer)
Then a JS event needs to push data into the dataLayer (see criteria in "More info" below and Steve A. & Malik have code examples) and when a user clicks on one of these buttons (Continue with FB/Google or Sign Up with Email) an event with the appropriate data should be visible in GTM via preview mode

### DIFF
--- a/src/SignupModal.jsx
+++ b/src/SignupModal.jsx
@@ -21,37 +21,21 @@ export const SIGNUP_MODAL_FACEBOOK_CLASS = `${SIGNUP_MODAL_CLASS}-facebook`;
 export const SIGNUP_MODAL_GOOGLE_CLASS = `${SIGNUP_MODAL_CLASS}-google`;
 export const SIGNUP_MODAL_EMAIL_CLASS = `${SIGNUP_MODAL_CLASS}-email`;
 
+
 /**
  * @param {Object} props component properties
  * @returns {React.element} SignupModal
  */
 export const SignupModal = ({ onDismiss, signupOptions, ...other }) => {
-	const facebookOnClick = () => {
+
+	const gaTracking = type => () => {
 		window.dataLayer = window.dataLayer || [];
 		window.dataLayer.push({
 			event: 'registrationStart',
 			form: {
-				type: 'facebookAuthButton',
+				type
 			},
 		});
-	};
-	const emailOnClick = () => {
-			window.dataLayer = window.dataLayer || [];
-			window.dataLayer.push({
-				event: 'registrationStart',
-				form: {
-					type: 'email',
-				},
-			});
-	};
-	const googleOnClick = () => {
-			window.dataLayer = window.dataLayer || [];
-			window.dataLayer.push({
-				event: 'registrationStart',
-				form: {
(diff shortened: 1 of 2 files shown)


src/SignupModal.jsx
```diff
@@ -26,6 +26,34 @@ export const SIGNUP_MODAL_EMAIL_CLASS = `${SIGNUP_MODAL_CLASS}-email`;
  * @returns {React.element} SignupModal
  */
 export const SignupModal = ({ onDismiss, signupOptions, ...other }) => {
+	const facebookOnClick = () => {
```
hsbacot: I'm not sure tracking is something that we'd want to address in each component. Is there more context to tracking GA in shared components or a way we've historically done this?

hsbacot: The functions could be consolidated into one as well, ex:
```jsx
const gaTracking = type => {
  window.dataLayer = window.dataLayer || [];
  window.dataLayer.push({
    event: 'registrationStart',
    form: {
      type
    },
  });
}
```

sergeychernyshev: It would need to return a handler though, something like
```js
const gaTracking = type => () => {
  window.dataLayer = window.dataLayer || [];
  window.dataLayer.push({
    event: 'registrationStart',
    form: {
      type
    },
  });
}
```
and used like:
```html
<a onClick={gaTracking('facebookAuthButton'}>...</a>
<a onClick={gaTracking('email'}>...</a>
<a onClick={gaTracking('googleAuthButton'}>...</a>
```

browne0: ahhh good idea! i'll do that instead

sergeychernyshev: Yeah, not sure if adding tracking like this to a shared component makes sense.

It's probably better to have additional click handler props in this component and manage tracking in the application where it's used.

hsbacot: cc: @mmcgahan

sergeychernyshev:
LGTM

src/SignupModal.jsx
browne0: as a follow up i factored it out to be prop functions that consumers can pass in to do tracking in their own apps.

hsbacot: Not sure we want to have specific props here for these functions. This specific type of configuration is hard to maintain and not an intuitive API for users. Changes like this should really wait to be merged in.

PR #716, opened 2019-06-06
166006697 Tooltip Styling

ILysenkoDA:
#### Related issues
Fixes https://www.pivotaltracker.com/story/show/166006697

#### Description
Icon: 24x24
Padding: 5px 

image

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -4775,7 +4775,10 @@ input.field--reset:active {
 
 .tooltip-closeBtn {
   float: right;
-  padding: 4px; }
+  padding: 5px !important; }
+  .tooltip-closeBtn .svg-icon {
+    height: 24px;
+    width: 24px; }
 
 .typeahead {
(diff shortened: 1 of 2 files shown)


assets/css/main.css
```diff
@@ -4775,7 +4775,10 @@ input.field--reset:active {
 
 .tooltip-closeBtn {
   float: right;
-  padding: 4px; }
+  padding: 5px !important; }
```
v-gtsolov: is there a way to escape the `!important` part?

ILysenkoDA: @v-gtsolov The button icon prop overwriting padding of the icon

v-gtsolov: can we make stronger selector for it than, and remove the `!important` part

hsbacot: This will get revisited by us with design and may be reverted back to 20px, however this is good for now

PR #717, opened 2019-06-06
Display mobile "events" link by default

daniel-hayes:
#### Related issues
Fixes https://www.pivotaltracker.com/story/show/166520016

#### Description
We have launched the `events` tab and no longer need to hide this value by default. It is currently not appearing on pro-web. 

Before:
Screen Shot 2019-06-06 at 12 58 47 PM
After:
Screen Shot 2019-06-06 at 12 58 02 PM


daniel-hayes:
@lipenco "events" should not be hidden. This original change was put in place because it was a beta release and we had put this link behind a feature flag toggling it off/on using this 'show' prop

jaclynj:
The code itself looks fine to me but it’s weird that MWC has this very pro-specific component that needs to have the props correspond to pro links. Since we’d like to phase out MWC it might make sense to move this out and into pro-web. Do you know what the history of this is and why this is in Meetup Web Components?

daniel-hayes:
@jaclynj I'm afraid I do not know it's history and I would be happy to discuss the alternate solutions available. 

I believe this exists here because we want Pro users who are not only on pro-web pages, but also mup-web to see this dropdown menu, and since this shared component exists in both places, it made sense to add here. However, I completely agree that it feels strange I need to update this in 3 different places in order to make the change work across Meetup. I don't know of any alternative solutions at the moment other than directly implementing this in both mup-web and pro-web? what do you think?

sadafie:
@daniel-hayes @jaclynj I have some familiarity with the history here - we wanted a single shared header across mup-web and pro - and this was the simplest place to put it. At first, it was pretty generic, but over time many people have contributed to it and there's been component scope creep. However, the path forward is that `nav` (and footer) should not exist in a components library. Because the web-platform team currently own meetup-web-components, we should/will make a plan to hopefully move nav out of meetup-web-components in a way that works for both mup-web and pro-web. This hasn't been prioritized yet, so for now we can just move forward with this. 

cc @hsbacot @mmcgahan

jaclynj:
ok, fine with me.

daniel-hayes:
Thanks for clearing that up @sadafie

PR #725, opened 2019-07-24
166979381 Tabs left selector

ignatevp:

### DIFF
--- a/assets/scss/components/_tabs.scss
+++ b/assets/scss/components/_tabs.scss
@@ -70,3 +70,19 @@ $tabBorderWidth: 3px;
 		border-right: $tabBorderWidth solid;
 	}
 }
+
+.tabs-tab--selectedLeft {
+	border-bottom: $tabBorderWidth solid $C_accent;
+	box-shadow: none;
+	color: $C_accent;
+	font-weight: $W_medium;
+
+	.tabs--isVertical & {
+		border-bottom: 0;
+		border-left: $tabBorderWidth solid;
+	}
+}
+
+.tabs-tab--notSelectedLeft {
+	padding-left: $tabBorderWidth !important;
+}
--- a/src/interactive/Tabs.jsx
+++ b/src/interactive/Tabs.jsx
@@ -7,21 +7,12 @@ import cx from 'classnames';
  */
 export class TabsTab extends React.Component {
 	render() {
-		const {
-			children,
-			isSelected,
-			className,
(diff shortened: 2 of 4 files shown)


src/interactive/Tabs.jsx
```diff
@@ -7,13 +7,21 @@ import cx from 'classnames';
  */
 export class TabsTab extends React.Component {
 	render() {
-		const { children, isSelected, className, ...other } = this.props;
+		const {
+			children,
+			isSelected,
+			className,
+			isLeftSelectorLayout,
```
eilinora: can you simplify this name to `isLeft` :)

src/interactive/Tabs.jsx
```diff
@@ -7,13 +7,21 @@ import cx from 'classnames';
...
+			...other
+		} = this.props;
+		const selectedClass = isLeftSelectorLayout
```
eilinora: no need to do this if else like this its the gift of what cx profiles. Instead i would do...
```javascript
{
					'tabs-tab--selectedLeft': isSelected && isLeftSelectorLayout,
					'tabs-tab--notSelectedLeft': !isSelected && isLeftSelectorLayout,
					'tabs-tab--selected': isSelected && !isLeftSelectorLayout
}
```

eilinora: And actually you can move that object into the `cx` init below :)

src/interactive/Tabs.jsx
ignatevp: ok I just was afraid it won't be obvious what is left

assets/scss/components/_tabs.scss
```diff
@@ -70,3 +70,19 @@ $tabBorderWidth: 3px;
 		border-right: $tabBorderWidth solid;
 	}
 }
+
+.tabs-tab--selectedLeft {
```
mmcgahan: A useful pattern for thinking about classnames is how they can be composed together, e.g. having a 'base class' that gets combined with other classes to 'tune' the style. In the future, a better approach would be to _always_ expect a `tabs-tab--selected` class to be applied to _any_ selected tab, with an _additional_ class of `tabs-tab--selectedLeft` in order to provide 'left' overrides. In that case, you would be able to avoid duplicating the styles in `.tabs-tab--selected`, and just provide the overrides.

cc @eilinora in case you're doing similar updates in the future

PR #726, opened 2019-07-25
Deprecation warnings #167078853

hsbacot:
#### Related issues
https://www.pivotaltracker.com/story/show/167078853

Adds deprecation warning to console when using components that are deprecated and will be removed from this repository on October 3rd, 2019

### DIFF
--- a/src/forms/Checkbox.jsx
+++ b/src/forms/Checkbox.jsx
@@ -1,11 +1,12 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Checkbox as SwarmCheckbox } from '@meetup/swarm-components';
+import DeprecationWarning from '../utils/components/DeprecationWarning';
 
 /**
  * @module Checkbox
  */
-class Checkbox extends React.PureComponent {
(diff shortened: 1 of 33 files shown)


src/forms/Checkbox.jsx
```diff
@@ -1,11 +1,12 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Checkbox as SwarmCheckbox } from '@meetup/swarm-components';
+import DeprecationWarning from '../utils/components/DeprecationWarning';
 
 /**
  * @module Checkbox
  */
-class Checkbox extends React.PureComponent {
+export class Checkbox extends React.PureComponent {
```
sergeychernyshev: Does this mean that if somebody imports Checkbox explicitly, they will not get the warning?
If we don't have any code that does that, why do we need to export it now?

hsbacot: It's for tests. The higher order component now wraps the default export causing the tests to not have the proper instance without it

PR #730, opened 2019-08-23
[PRELIM] Node 10 upgrade

sadafie:
#### Related issues


.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-  - "8.11"
+  - 10.16.3
```
mmcgahan: `'10.16'` to allow patch updates?

assets/css/main.css
```diff
@@ -16,51 +16,51 @@
  */
 /* Default values */
 :root {
-  --c-white: white;
-  --c-lightGray: #f6f7f8;
-  --c-mediumGray: #757575;
-  --c-darkGray: #353e48;
-  --c-coolGrayLight: #e4e9ed;
+  --c-white: rgb(255, 255, 255);
```
mmcgahan: any idea where these changes came from?

sadafie: node-sass upgrade. the rgb values are correct. not sure the exact change in node-sass, but perhaps another one of their bug fixes that has a side-effect.

PR #732, opened 2019-09-11
disable toggle switch button

andrew-bykovets:
#### Related issues
https://www.pivotaltracker.com/story/show/168262009

#### Description
add an option to disabled toggle but show hidden text

Screen Shot 2019-09-11 at 15 12 05

### DIFF
--- a/src/interactive/AccordionPanel.jsx
+++ b/src/interactive/AccordionPanel.jsx
@@ -56,8 +56,22 @@ class AccordionPanel extends React.Component {
 	_handleToggle(e) {
 		e.preventDefault();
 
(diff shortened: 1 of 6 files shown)


src/interactive/AccordionPanel.jsx
```diff
@@ -183,8 +199,8 @@ class AccordionPanel extends React.Component {
 				<div
 					role="tab"
 					aria-controls={`panel-${panelId}`}
-					aria-expanded={isOpen}
-					aria-selected={isOpen}
+					aria-expanded={isOpen && !isDisabledPanelOpen}
```
komlev: no sure if this is correct
Panel should be considered opened if it has disabled content

PR #782, opened 2021-02-12
[176853966] Add search bar to header

vsleibenko:
#### Related issues
This PR resolves: https://www.pivotaltracker.com/story/show/176853966

#### Description
Main goal was to create search bar in header to be similar to build-meetup

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/src/navigation/Nav.jsx
+++ b/src/navigation/Nav.jsx
@@ -262,6 +262,7 @@ export class Nav extends React.Component {
 			isSearchEnabled &&
 				!media.isAtMediumUp && {
 					shrink: true,
+					icon: search.icon,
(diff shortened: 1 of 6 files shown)


src/navigation/Nav.jsx
```diff
@@ -262,7 +262,6 @@ export class Nav extends React.Component {
 			isSearchEnabled &&
 				!media.isAtMediumUp && {
 					shrink: true,
-					icon: search.icon,
```
BomberNastya, line 265: I would keep this icon for backward compatibility. Maybe do something like this instead: 
```icon: !isNewNavActive ? search.icon : undefined```

BomberNastya (approved):
LGTM with one suggestion

PR #835, opened 2023-01-25
add _blank to LinkItem

oneilsw:
[Shortcut](https://app.shortcut.com/meetup/story/66738/fe-desktop-add-pro-button-in-core-nav-bar)

To fix- 
"Expected : Click on the Pro button should open a page with Pro Landing Page in a new tab (act the same as pro banner)
Actual: Click on the Pro button opens the page with Pro Landing Page in the same tab."


NikSubbo (approved):
I would call it not isTryProButton, but something like "isTargetBlank", because it is not ProButton specific and can be used in other cases.

PR #863, opened 2023-04-13
Web Alt text Event Create page [sc-72006]

meetup-oleksandr-dudiuk:
#### Related issues
Fixes PIVOTAL-xxx

#### Description
Set the presentation role as the default for our image component.

In many cases, we use an icon with the `img` role without an `aria-label`, which can cause a lot of errors. In my opinion, it would be better to set an `aria-label` for icons that are used without supporting text, instead default role `img`.

### DIFF
--- a/src/media/Icon.jsx
+++ b/src/media/Icon.jsx
@@ -19,11 +19,16 @@ class Icon extends React.PureComponent {
 	}
 
 	render() {
-		const { className, circled, role, ...other } = this.props;
+		const { className, circled, role, ariaLabel, ...other } = this.props;
 
 		return (
 			<span className={className}>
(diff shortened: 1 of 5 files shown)


komlev (approved):

meetup-anton-hulak (approved):

src/media/Icon.jsx
```diff
@@ -45,6 +50,12 @@ Icon.propTypes = {
 
 	/** What color the icon should be filled with */
 	color: PropTypes.string,
+
+	/** Gives icon role */
+	role: PropTypes.bool,
```
mkleymenov, line 55: Shouldn't it be `PropTypes.string`?

PR #873, opened 2023-04-20
Fixed role issues for `AccordionPanel`

ekaterinashenets:
#### Description
Fixed role issues for AccordionPanel

### DIFF
--- a/src/interactive/AccordionPanelGroup.jsx
+++ b/src/interactive/AccordionPanelGroup.jsx
@@ -78,13 +78,13 @@ class AccordionPanelGroup extends React.Component {
 
 		return (
 			<ul
-				role="tablist"
+				// role="tablist"
 				aria-multiselectable={multiSelectable}
 				className={classNames}
 				{...other}
 			>
 				{accordionPanels.map((panel, i) => (
-					<li key={i} className="list-item flush--top" role="tab">
+					<li key={i} className="list-item flush--top" /* role="tab"*/>
 						{React.cloneElement(
 							panel,
 							{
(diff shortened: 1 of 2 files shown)


src/interactive/AccordionPanelGroup.jsx
```diff
@@ -78,13 +78,13 @@ class AccordionPanelGroup extends React.Component {
...
 			<ul
-				role="tablist"
+				// role="tablist"
 				aria-multiselectable={multiSelectable}
 				className={classNames}
 				{...other}
 			>
 				{accordionPanels.map((panel, i) => (
-					<li key={i} className="list-item flush--top" role="tab">
+					<li key={i} className="list-item flush--top">
```
meetup-oleksandr-dudiuk, line 87: `role="list"` and `role="listitem"` can't be used here?  could this be a solution?

ekaterinashenets: `role="list"` and `role="listitem"` causes issues of another type for all the `li` elements:

**Interactive controls must not be nested**
_Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies_

ekaterinashenets: @meetup-oleksandr-dudiuk  found a solution with `tree` and `treeitem`.
Looks like this suits good for the accordion components - see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role

meetup-oleksandr-dudiuk (approved):

mkleymenov (approved):

PR #876, opened 2023-04-21
Added aria-label attribute to fields without visible label

ekaterinashenets:
#### Description
Added aria-label attribute to fields without visible label

### DIFF
--- a/src/forms/TextInput.jsx
+++ b/src/forms/TextInput.jsx
@@ -115,6 +115,7 @@ export class TextInput extends React.Component<Props> {
 						pattern={pattern}
 						disabled={disabled}
 						id={id}
+						aria-label={label ? id || name : undefined}
 						{...optionalInputProps}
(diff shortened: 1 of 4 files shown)


src/forms/TextInput.jsx
```diff
@@ -115,6 +115,7 @@ export class TextInput extends React.Component<Props> {
 						pattern={pattern}
 						disabled={disabled}
 						id={id}
+						aria-label={label ? undefined : id || name}
```
mkleymenov, line 118: It's a bit difficult to understand whether `|| name` is a fallback for the `id` or the entire ternary operator. Could you add round brackes to explicitly specify precedence of operations?

ekaterinashenets: @mkleymenov it was added, but prettier removed it :)

mkleymenov (approved):

PR #882, opened 2023-05-01
Event Create Nested According to Spec [sc-72013]

meetup-oleksandr-dudiuk:
#### Related issues
Fixes PIVOTAL-xxx

### DIFF
--- a/src/navigation/components/NavItem.jsx
+++ b/src/navigation/components/NavItem.jsx
@@ -100,6 +100,7 @@ export const NavItem = props => {
 
 	const trigger = (
 		<div
+			role="button"
 			aria-haspopup
 			className={NAV_ITEM_CLASS}
 			onClick={onClickAction}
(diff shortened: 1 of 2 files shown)


src/navigation/components/NavItem.jsx
```diff
@@ -99,15 +99,14 @@ export const NavItem = props => {
 	);
 
 	const trigger = (
-		<Button
```
meetup-oleksandr-dudiuk, line 102: The button was removed because the parent element already has the role of a button. If we were to use the button element, it would result in a nested role, which is not desirable.

meetup-anton-hulak: Changing button to div may cause other accessibility issues since div by default is not clickable element and should have `role="button"` attribute to act like button

meetup-oleksandr-dudiuk: let me test it, maybe you right 🤔

meetup-oleksandr-dudiuk: yeah its works fine 👍

meetup-anton-hulak (approved):

PR #883, opened 2023-05-12
Implemented workaround to set selected attribute to the date in calendar

ekaterinashenets:
#### Description

This pr implements workaround to set selected attribute to the day element in calendar

### DIFF
--- a/src/forms/CalendarComponent.jsx
+++ b/src/forms/CalendarComponent.jsx
@@ -161,7 +161,7 @@ export class CalendarComponent extends React.Component<Props> {
 						if (!dayElem.ariaCurrent) {
 							dayElem.ariaPressed = isSelected;
 						} else if (isSelected) {
-							dayElem.ariaPressed = true; // apply aria-pressed attr to current date only in case it's true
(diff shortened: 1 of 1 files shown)


src/forms/CalendarComponent.jsx
```diff
@@ -154,6 +155,15 @@ export class CalendarComponent extends React.Component<Props> {
...
 					value={this.getPickrValue(value)}
+					onDayCreate={(_, __, ___, dayElem) => {
+						const isSelected = dayElem.classList.contains('selected');
+						dayElem.role = 'button';
+						if (!dayElem.ariaCurrent) {
+							dayElem.ariaPressed = isSelected;
+						} else if (isSelected) {
+							dayElem.ariaPressed = true; // apply aria-pressed attr to current date only in case it's true
+						}
+					}}
```
ekaterinashenets, line 166: There are no options/props in this library which might help to set corresponding attribute to the day element in Calendar. 
I was able to find a solution to set `aria-pressed` attribute value directly

PR #890, opened 2023-10-06
Pro Dashboard Form elements must have labels [sc-74561]

meetup-oleksandr-dudiuk:
#### Related issues
Fixes PIVOTAL-xxx

### DIFF
--- a/src/forms/Checkbox.jsx
+++ b/src/forms/Checkbox.jsx
@@ -68,6 +68,7 @@ export class Checkbox extends React.PureComponent {
 			disabled,
 			name,
 			value = '',
+			ariaLabel,
 			className: customStyle,
 			...other
 		} = this.props;
@@ -86,7 +87,6 @@ export class Checkbox extends React.PureComponent {
 					className="checkbox"
 					role="checkbox"
 					aria-checked={stateChecked}
-					aria-label="checkbox"
 					tabIndex={0}
 					onKeyPress={this.handleKeyPress}
 				>
@@ -102,6 +102,7 @@ export class Checkbox extends React.PureComponent {
 					onChange={this.onChange}
 					readOnly={!this.onChange || disabled}
 					name={name}
+					aria-label={ariaLabel || 'checkbox'}
 					value={`${value}`}
 				/>
 				<span>{label}</span>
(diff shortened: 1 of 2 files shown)


src/forms/Checkbox.jsx
```diff
@@ -102,6 +102,7 @@ export class Checkbox extends React.PureComponent {
 					onChange={this.onChange}
 					readOnly={!this.onChange || disabled}
 					name={name}
+					aria-label={ariaLabel || 'checkbox'}
```
NikSubbo, line 105: Do we need additional prop for it? It will require to add it to all existing checkboxes.
I think we can just make it `aria-label={label || 'checkbox'}`

meetup-oleksandr-dudiuk: 🤔  make sense hah

meetup-oleksandr-dudiuk: I will do it like this {`${label} - checkbox ` || 'checkbox'}

src/forms/Checkbox.jsx
```diff
@@ -102,6 +101,7 @@ export class Checkbox extends React.PureComponent {
 					onChange={this.onChange}
 					readOnly={!this.onChange || disabled}
 					name={name}
+					aria-label={`${label} - checkbox ` || 'checkbox'}
```
NikSubbo, line 104: Left side will be always true, because it is not empty string.
If label is undefined, it will be "undefined - checkbox".
Screenshot from 2023-10-06 14-20-06

src/forms/Checkbox.jsx
```diff
@@ -102,6 +101,9 @@ export class Checkbox extends React.PureComponent {
 					onChange={this.onChange}
 					readOnly={!this.onChange || disabled}
 					name={name}
+					aria-label={
+						typeof label === 'string' ? `${label} - checkbox ` : 'checkbox'
```
NikSubbo, line 105: There is extra space after `${label} - checkbox`, but overall approve.

NikSubbo (approved):

src/forms/Checkbox.jsx
meetup-oleksandr-dudiuk: yep label also can be a element so I switch it to
 aria-label={ typeof label === 'string' ? \`${label} - checkbox \` : 'checkbox' }

PR #899, opened 2024-01-26
Feature/sc 109316/fe pro dashboard accessibility fixes

meetup-oleksandr-dudiuk:
#### Related issues
Fixes PIVOTAL-xxx

### DIFF
--- a/src/__snapshots__/footer.test.jsx.snap
+++ b/src/__snapshots__/footer.test.jsx.snap
@@ -207,7 +207,7 @@ exports[`Footer exists 1`] = `
           items={
             Array [
               <span>
-                © 2023 Meetup LLC
+                © 2024 Meetup LLC
               </span>,
             ]
           }
--- a/src/forms/Button.jsx
+++ b/src/forms/Button.jsx
@@ -106,15 +106,15 @@ export class Button extends React.PureComponent {
 				{icon ? (
 					<span>
 						{right ? (
-							<span>
+							<React.Fragment>
 								{children}
 								<Icon shape={iconShape} size="xs" {...iconProps} />
-							</span>
+							</React.Fragment>
 						) : (
-							<span>
+							<React.Fragment>
 								<Icon shape={iconShape} size="xs" {...iconProps} />
 								{children}
-							</span>
+							</React.Fragment>
 						)}
 					</span>
 				) : iconOnly ? (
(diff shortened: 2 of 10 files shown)


src/__snapshots__/footer.test.jsx.snap
```diff
@@ -207,7 +207,7 @@ exports[`Footer exists 1`] = `
           items={
             Array [
               <span>
-                © 2023 Meetup LLC
+                © 2024 Meetup LLC
```
meetup-anton-hulak, line 210: Maybe it does makes sense to use `new Date().getFullYear()` here?

src/navigation/Nav.jsx
```diff
@@ -282,7 +283,11 @@ export class Nav extends React.Component {
 		};
 
 		const getConnectionsIcon = () => (
-			<img className="proIcon" alt={connections.label} src={CONNECTIONS_ICON} />
+			<img
+				className="proIcon"
+				alt={connections && connections.label}
```
meetup-anton-hulak, line 288: Is it possible that connections can be `null` or `undefined`?  If so let's have a fallback string value here

src/forms/Button.jsx
```diff
@@ -85,7 +90,40 @@ export class Button extends React.PureComponent {
...
+			<button
+				data-swarm-button={buttonType}
+				data-swarm-size={getSwarmSize(this.props)}
+				data-icon={getIconPosition(this.props)}
+				data-swarm-width={width}
+				type="button"
+				ref={forwardedRef}
+				{...other}
+			>
+				{icon ? (
```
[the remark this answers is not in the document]
meetup-oleksandr-dudiuk: it's just a copypast from swarm repo

src/navigation/Nav.jsx
meetup-oleksandr-dudiuk: yep it's possible cause on pro-web connections not implemented and I just try to fix issue for update a lib version, it's not a part of my work

src/forms/Button.jsx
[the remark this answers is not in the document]
meetup-anton-hulak: I think it's a good chance to improve the code ;)

src/navigation/Nav.jsx
meetup-oleksandr-dudiuk: I think it should be removed when connections will be implemented on pro-web, but idk who deals with this

meetup-anton-hulak: I see, but it will fix the type error, but still will leave alt attribute blank. Why not to use ternary here? `connectios ? connections.label : 'pro-icon` (or some other fallback)

meetup-anton-hulak (approved):

